PR #14, opened 2015-05-16
Fix tests and deprecation warnings

nalimilan:
Using 'call' as a variable name currently triggers a bug in
Julia 0.4: https://github.com/JuliaLang/julia/issues/11295
Work around it for now by using a different name.

Also get rid of deprecation warnings printed during the tests.

This PR currently breaks compatibility with 0.3. Should I use Compat.jl, or do you prefer creating a new branch for 0.4?

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,1 +1,2 @@
 julia 0.3-
+Compat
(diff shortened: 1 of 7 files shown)


tshort:
Thanks for the updates, @nalimilan! As far as compatibility, I prefer the Compat.jl option.

nalimilan:
OK, I've added `@compat` calls. Tests now pass with both 0.3.8 and 0.4 master.

tshort:
Thanks again, @nalimilan.

nalimilan:
You're welcome. Should a new release be tagged?

tshort:
Yes. Good point.

PR #26, opened 2015-08-19
Added @ix! and @where!

bramtayl:
I've added these mutate-in-place functions. I'm not sure I fully understand the distinction between the two. The moreargs version of @ix is not duplicated (I'm not sure what it does), and neither are the function argument methods of @where for the same reason. There are tests.


tshort:
I'm unsure about this one. `@where!` might make sense. Is `deleterows!` noticeably faster or more memory efficient than creating a new DataFrame?

bramtayl:
Actually, I don't know. I assumed it would be because you don't have to duplicate all that data.

bramtayl:
If it is the case that deleterows! isn't actually that much more efficient, it seems like there must exist a better alternative. Does Julia have the equivalent of a getindex! function?

bramtayl:
Going by the test below, deleterows! actually seems to be slowing things down fourfold.

```{julia}
scale = 10000000

@time begin
test = DataFrame(a = fill(1, scale), b = fill(2, scale) )
test2 = deleterows!(test, 2:2:scale)
end

@time begin
test = DataFrame(a = fill(1, scale), b = fill(2, scale) )
test2 = test[2:2:scale, :]
end
```

bramtayl:
Would subdataframes make sense for something like this?

tshort:
SubDataFrame versions make a lot of sense to me. I'm not sure what the best convention should be.

bramtayl:
@subset maybe?

tshort:
I like the name `@subset`. I'm not sure how well it will fit in when chaining things together. Hopefully, `@by`, `@orderby`, and `@select` all work. I'm not sure about `@transform`, though. And, I don't know if `@where` will give you another SubDataFrame.

bramtayl:
It doesn't seem that `@transform`, `@select`, `@byrow!`, or `@subset` will work with subdataframes.

```{julia}
df = DataFrame(a = [1, 2, 3], b = [1, 2, 3])

sub = SubDataFrame(df, [1, 2])

sub2 = SubDataFrame(sub, [1])
@transform(sub, c = :a)
@select(sub, :a)
@byrow(sub, :a = :b)
```

tshort:
Most of those macros don't work for SubDataFrames because basic setindex or getindex operations needed to implement them are not supported in the DataFrames package. That means it'll be tricky to use subsets like this. Maybe as base Julia moves more towards use of views, we can tag along and improve DataFrames support around this.

PR #28, opened 2015-10-12
removed 0.4 deprecation wanrings

merl-dev:


tshort:
Does this still work with Julia v0.3?

alyst:
Only when wrapped in `@compat()`

ViralBShah:
How about we do 0.4 changes, and tag new releases for Julia 0.4?

tshort:
Good plan, Viral.

tkelman:
Adding `@compat` would be pretty harmless, this doesn't seem like a very good reason to drop 0.3 support just yet.

Should really turn on travis here.

ViralBShah:
Yes, we can add `@compat` here, but I figured there was no problem making this work on master at least and merging the PR.

alyst:
BTW, are there some core-devs-approved guidelines/recommendations for dropping 0.3 support in the packages?

tkelman:
See the readme on METADATA.jl. If you drop 0.3 support, be sure to change your REQUIRE file to say `julia 0.4` and increment at least the minor package version via `Pkg.tag("Foo", :minor)`

alyst:
@tkelman thanks!

PR #36, opened 2015-11-13
Add benchmarks from R's data.table benchmarks

tshort:
These benchmarks are from my favorite R package:

https://github.com/Rdatatable/data.table/wiki/Benchmarks-:-Grouping

We are way behind. All of the DataFrame grouping operations are at least 10X slower than data.table. On the plus side, the syntax compares favorably with either data.table or dplyr.

### DIFF
--- a/test/data.table.timings.jl
+++ b/test/data.table.timings.jl
@@ -1,0 +1,86 @@
+# Julia benchmarks from R's data.table
+# https://github.com/Rdatatable/data.table/wiki/Benchmarks-:-Grouping
+
+using DataFrames, DataFramesMeta
+using NullableArrays
+
+N=10_000_000; K=100
+srand(1)
+
+# Array version
+
+DA = DataFrame(
+  id1 = P(rand([symbol(string("id", i)) for i=1:K], N)),          # large groups (char)
+  id2 = P(rand([symbol(string("id", i)) for i=1:K], N)),          # large groups (char)
+  id3 = P(rand([symbol(string("id", i)) for i=1:Int(N/K)], N)),   # small groups (char)
+  id4 = P(rand(1:K, N)),                          # large groups (int)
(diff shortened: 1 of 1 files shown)


alyst:
There's my PR JuliaStats/DataFrames.jl#850 to address joining and grouping stability. Would try to post the results for this benchmark, when I will have little more time. Also it would be nice if someone reviews that PR at some point; it' I'm using it on daily basis for 10^6 rows datasets, because the multicolumn join in master have some troubles.

As for syntax, IMO for joins dplyr is a little bit better: writing `inner_join()` is more readable than `join(.... kind=:inner)`; also the requirement that a) the joined column names must match and b) have to be specified explicitly is somewhat redundant and at times annoying.

tshort:
@alyst, hi! I already tried your branch on these benchmarks. I was hopeful that by using Arrays, your PR would be faster, but it wasn't any faster. We're still quite slow. We probably have type stability issues. I think there's a lot of good stuff in your PR, though. Over the weekend, I will try to post to that issue to discuss more.

alyst:
@tshort Thanks! It would be nice to pinpoint the lines, where these inefficiencies occur. I mentioned in that PR one case, where with the help of memory profiling I have identified the line that caused serious regression, and it was actually not even in `DataFrames`, but `DataArrays`.

I haven't run the new benchmark yet, but we have to take into account that it's synthetic, so it tests a few situations, but not all. For example, there are cases where the master would be faster than my PR, just because you don't need to build the hash for single-column join.

For a more comprehensive outlook, it would be nice to have the benchmark based on some moderately sized DB (i.e.several tables, multicolumn keys, ~10^8 rows).

nalimilan:
@cjprybol These benchmarks could be interesting to have in DataTables to measure the speed of the new implementation from https://github.com/JuliaData/DataTables.jl/pull/17 compared with data.table.

Issue #42, opened 2016-03-26
Ergonomic ways of recoding data


tshort:
This issue parallels https://github.com/JuliaStats/DataFrames.jl/issues/369. 

`@byrow!` may be close to what @nalimilan is after in the issue above. Here is a recoding example:

```julia
n = nrow(df)
df[:x] = Array(Int, n)
df[:y] = NullableArray(Float64, n)
@byrow! df begin
    :x = :a > 1 ? 4 : 9
    :y = :a * :b
end
```
The main irritation with the code above is allocating new columns. That could be handled by extending `@transform` or making a new macro that does something similar:

```julia
@transform(df, x = Array{Int}, y = NullableArray{Float64})
```
Another option is to add a feature to `@byrow!` that does this:

```julia
@byrow! df let x = Array{Int},
               y = NullableArray{Float64}
    :x = :a > 1 ? 4 : 9
    :y = :a * :b
end
```
Or:
```julia
@byrow! df begin
    @newcol x = Array{Int}
    @newcol y = NullableArray{Float64}
    :x = :a > 1 ? 4 : 9
    :y = :a * :b
end
```
Any opinions on these or other approaches? Any other features that `byrow!` needs to better recode data?

nalimilan:
That's really cool. I would say the ideal solution would be to automatically create new columns, choosing the most appropriate type like Julia does for variables. Though I guess it would be tricky to decide whether to use a `NullableArray` or a standard array.

Given it may be hard to get this to work, an explicit syntax is a good solution (be it temporary or final). I think I'd prefer the `@newcol` form, but writing `@newcol x::Array{Int}` to match the standard Julia syntax. Actually, maybe `@byrow!` could detect type assertions and treat them as declarations of a new column, without even `@newcol`.

Then there's also the question of how to allow propagating null values. I guess we need to wait for the `@lift` macro to work fine in NullableArrays.jl before implementing this.

Finally, there's the minor annoyance of writing `:x` instead of `x`, which we discussed elsewhere already. But that would really be the icing on the cake: I think we can close this issue even if we don't have it. :-)

PS: Your example above doesn't work when copy/pasting: looks like you forgot one line creating `df`. See also https://github.com/JuliaStats/DataFrames.jl/issues/933

PR #44, opened 2016-03-27
Add @newcol feature to @byrow!

tshort:
See #42 for background. Some notes:

* New columns don't get added to the original DataFrame, only the result.

* The syntax is based on the suggestion by @nalimilan:

```julia
df = DataFrame(A = 1:3, B = [2, 1, 2])
df2 = @byrow! df begin
    @newcol colX::Array{Float64}
    @newcol colY::DataArray{Int}
    :colX = :B == 2 ? pi * :A : :B
    if :A > 1 
        :colY = :A * :B
    end
end
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,6 +123,27 @@ let x = 0.0
 end
 ```
 
+Note that the let block is required here to create a scope to allow assignment
+of `x` within `@byrow!`.
+
+`byrow!` also supports special syntax for allocating new columns to make
+`byrow!` more useful for data transformations. The syntax `@newcol
+x::Array{Int}` allocates a new column `:x` with an `Array` container with eltype
+`Int`. Note that the returned AbstractDataFrame includes these new columns, but
+the original `df` is not affected. Here is an example where two new columns are
+added:
+
+```julia
+df = DataFrame(A = 1:3, B = [2, 1, 2])
+df2 = @byrow! df begin
+    @newcol colX::Array{Float64}
+    @newcol colY::DataArray{Int}
+    :colX = :B == 2 ? pi * :A : :B
+    if :A > 1 
+        :colY = :A * :B
+    end
+end
+```
 
 ## LINQ-Style Queries and Transforms
 
--- a/src/byrow.jl
+++ b/src/byrow.jl
@@ -26,12 +26,37 @@ end
 # other than Expr (generally a Symbol or a literal).
 byrow_replace(x) = x
 
+function byrow_find_newcols(e::Expr, newcol_decl)
+    if e.head == :macrocall && e.args[1] == symbol("@newcol")
+        ea = e.args[2]
+        # expression to assign a new column to df
+        return (nothing, Any[Expr(:kw, ea.args[1], Expr(:call, ea.args[2].args[1], ea.args[2].args[2], :_N))])
+    else
+        if isempty(e.args)
+            return (e.args, Any[])
+        end
+        newargs = Any[]
+        for ea in e.args
+            (nea, newcol) = byrow_find_newcols(ea, newcol_decl)
+            nea != nothing && push!(newargs, nea) 
+            nea == nothing && length(newcol) > 0 && append!(newcol_decl, newcol) 
+        end
+        return (Expr(e.head, newargs...), newcol_decl)
+    end
+end
+
+byrow_find_newcols(x, newcol_decl) = (x, Any[])
+
 function byrow_helper(df, body)
-    e_delimiters = Expr(:(=), :row, :( 1:length($df[1]) ))
-    e_forloop = Expr(:for, e_delimiters, byrow_replace(body))
-    return Expr(:block,
-                Expr(:macrocall, symbol("@with"), df, e_forloop),
-                df)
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -123,6 +123,26 @@ let x = 0.0
 end
 ```
 
+Note that the let block is used here to create a scope to allow assignment of
```
nalimilan: Not really related, but I don't get why `let` is presented here. AFAICT it isn't specific to `@byrow!` and isn't particularly useful with that macro (at least not more than in other situations).

README.md
```diff
@@ -123,6 +123,26 @@ let x = 0.0
 end
 ```
 
+Note that the let block is used here to create a scope to allow assignment of 
+`x` within `@byrow!`.
+
+`byrow!` also supports special syntax for allocating new columns to make
+`byrow!` more useful for data transformations. The syntax `@newcol
+x::Array{Int}` allocates a new column `:x` with an `Array` container with eltype
+`Int`. Note that the returned AbstractDataFrame includes these new columns, but
```
nalimilan: AbstractDataFrame needs backquotes (or just say "data frame").

I find it a bit problematic that the new columns are not added to the original data frame, given that the macro is called `@byrow!`. Is there a technical reason for this?

README.md
```diff
@@ -123,6 +123,26 @@ let x = 0.0
...
+the original `d` is not affected. Here is an example:
```
nalimilan: d -> df

src/byrow.jl
```diff
@@ -6,7 +6,7 @@ export @byrow!
 ## @byrow!
 ##
 ##############################################################################
-
+
```
nalimilan: Spacing bug here.

src/byrow.jl
```diff
@@ -46,11 +71,21 @@ Includes support for control flow and `begin end` blocks. Since the
...
 
+`byrow!` also supports special syntax for allocating new columns. The syntax
+`@newcol x::Array{Int}` allocates a new column `:x` with an `Array` container
+with eltype `Int`. Note that the returned AbstractDataFrame includes these new
+columns, but the original `d` is not affected. This feature makes it easier to
+use `byrow!` for data transformations.
+
 ### Arguments
 
 * `d` : an AbstractDataFrame
```
nalimilan: AbstractDataFrame needs backquotes here and below too.

README.md
tshort: That code won't work without the `let`block or another block that creates a local scope. At the REPL, you need the `let` block or a `global` keyword for `x`. Inside a function, it's okay without the `let` block.

nalimilan: Ah, OK. Maybe say "is required" instead of "is used" then.

nalimilan:
Great! Have you considered whether simply writing `x::Array{Int}` would work? It seems that looking for `ex.head == symbol("::")` could do the trick.

README.md
tshort: Thx.

I'm still pondering whether to modify the original. It's easy to change (I've implemented it both ways). The bang is there mainly because the user can overwrite columns. My main reason is that I suspect that future type-stable AbstractDataFrames won't allow `setindex!`; the columns will be fixed. I think this is true of CompositeDataFrames and TypedTables. So, I thought that it would be good to avoid `setindex!` and get used to using the returned value.

tshort:
It should work to leave out the `@newcol` and just use `x::Array{Int}`. I decided against it because it seems too magical. The `@newcol` indicator should help readability.

nalimilan:
> It should work to leave out the @newcol and just use x::Array{Int}. I decided against it because it seems too magical. The @newcol indicator should help readability.

Well, I like it because it's really the same as declaring the type of a variable. And if in the future we are able to create columns automatically (guessing its type), it will be exactly parallel to how variables work in Julia.

README.md
nalimilan: > My main reason is that I suspect that future type-stable AbstractDataFrames won't allow setindex!; the columns will be fixed. I think this is true of CompositeDataFrames and TypedTables. So, I thought that it would be good to avoid setindex! and get used to using the returned value.

Hmm, though issue indeed. We'll only be able to judge once we have more experience with typed data frames.

tshort: Even though `setindex!` may not be allowed in type-stable AbstractDataFrames, macros have enough magic to still allow re-assignment. We can have `@assign d[:a] = newCol` do the following:

```julia
df = @transform(df, a = newCol)
```
We can take the same approach with `@newcol`/`byrow!`. However, all of the other DataFramesMeta functions require the use of the returned value, so maybe the current approach is more consistent. I'm still undecided here.

nalimilan: Yeah, I had thought about doing that. But that would introduce type instability for typed data frames, so it should only be allowed for the current `DataFrame` to avoid trapping users. Maybe better keep it as you wrote it, and possibly revise this later.

nalimilan: @andyferris As you can see above, the DataFramesMeta API should work fine for TypedTables too. If you find cases where you'd need a different API, filing an issue to discuss this could be a good idea.

PR #50, opened 2016-04-05
Improvements to CompositeDataFrames, including type-stable row access

tshort:
Includes:

*  `row(d,i)` and `eachrow(d, i)` for type-stable row access
* Named composite type creation
* More documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -256,8 +256,10 @@ All of the other macros are based on `@with`.
 
 # CompositeDataFrame
 
-A CompositeDataFrame is an AbstractDataFrame built using Composite
-types. The advantages of this are:
+A `CompositeDataFrame` is a type-stable `AbstractDataFrame` built using composite
+types. Each column is a field in a composite type. `CompositeDataFrame` is an
+abstract type; each concrete composite type inherits from this. The advantages
+of this approach are:
 
 * You can access single columns directly using `df.colA`. This is type stable,
   so code should be faster. (There is still the function boundary to worry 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -281,6 +283,58 @@ Some downsides include:
...
+    a::Vector{Int}
+    b::Vector{Float64}
+    c::DataVector{Float64}
+end
+
+MyDF(n::Integer) = MyDF(zeros(Int, n), zeros(n), DataArray(zeros(n)))
+d = MyDF(10)
+```
+
+Note that a CompositeDataFrame is type stable with field access like `df.colA`
```
nalimilan: Missing backquotes here.

nalimilan: Missing backquotes here.

src/compositedataframe.jl
```diff
@@ -1,23 +1,93 @@
 using Compat
 
-export AbstractCompositeDataFrame, CompositeDataFrame
+export AbstractCompositeDataFrame, AbstractCompositeDataFrameRow, 
+       CompositeDataFrame, row
 
+"""
+    AbstractCompositeDataFrame
+    
+An abstract type that is an AbstractDataFrame. Each type that inherits from
```
nalimilan: Here too.

src/compositedataframe.jl
```diff
@@ -1,23 +1,93 @@
...
+"""
 abstract AbstractCompositeDataFrame <: AbstractDataFrame
 
+abstract AbstractCompositeDataFrameRow
+
+
+"""
+    row(cdf::AbstractCompositeDataFrame, i)
+    
+Return the `CompositeDataFrameRow` object for row `i` of `cdf`. This object has
```
nalimilan: I would have said "Return row `i` of `cdf` as a `CompositeDataFrameRow` object." Maybe more natural.

src/compositedataframe.jl
```diff
@@ -1,23 +1,93 @@
...
+
+"""
+```julia
+CompositeDataFrame(columns::Vector{Any}, cnames::Vector{Symbol})
+CompositeDataFrame(columns::Vector{Any}, cnames::Vector{Symbol}, typename::Symbol)
+CompositeDataFrame(; kwargs...)
+CompositeDataFrame(typename::Symbol; kwargs...)
+```
+
+A constructor of AbstractCompositeDataFrames that mimics the `DataFrame`
```
nalimilan: Here too.

src/compositedataframe.jl
```diff
@@ -1,23 +1,93 @@
...
+```julia
+df = CompositeDataFrame(x = 1:3, y = [2, 1, 6])
+dfr = row(df, 3)
+dfr.y   # 6
+```
+"""
+row() = nothing
+
+"""
+```julia
```
nalimilan: The convention is to use a four-space indent.

src/compositedataframe.jl
```diff
@@ -1,23 +1,93 @@
...
+This uses `eval` to create a new type within the current module. 
+
+### Arguments
+
+* `columns` : contains the contents of the columns
+* `cnames` : the names of the columns
+* `typename` : the optional name of the type created
+* `kwargs` : the key gives the column names, and the value is the column contents
+
+### Returns
```
nalimilan: In the recent guidelines, we decided that the return value would generally better be documented directly in the function description

nalimilan: In the recent guidelines, we decided that the return value would generally better be documented directly in the function description

nalimilan:
Cool. How do you envision the difference between `CompositeDataFrame` and `TypedTable`? Looks like the general idea is similar, right?

tshort:
Thanks for the reviews, @nalimilan. The `CompositeDataFrame` and the `TypedTable` are aiming for similar goals. The advantages of a `CompositeDataFrame` are:

* You can use `df.colA` syntax, and this is type-stable and fast.
* It's a normal Julia type and doesn't need generated functions.

The advantages of a `TypedTable` are:

* Because it uses generated functions, this approach might be less stressful on Julia's type system if you make or modify lots of data frames.
* If the `@pure` syntax works out, `df[:colA]` could be quite fast. I'm really hoping this works out. Right now, the API for TypedTables is fairly cumbersome.

In both cases, you can't use `setindex!`.

nalimilan:
Thanks for the explanation. Ideally, all data frame variants would more or less support the same high-level API, so that one can easily switch between them to compare their respective merits. Maybe you could move `CompositeDataFrame` to its own package, to make it more visible?

tshort:
It makes sense for `CompositeDataFrame` to be in its own package. I'll probably hold off though. As @johnmyleswhite noted in https://github.com/JuliaStats/Roadmap.jl/issues/19, we have too many packages in JuliaStats that don't get support and updates. I'm afraid this might be the same. The best way to implement typed data frames is still up in the air, and changes to Julia might make another change more attractive (using `@pure` for example).

I made these recent updates because I finally used `CompositeDataFrame` in a project. Most of the time, `@with` and `@byrow!` are sufficiently clean and fast enough for my stuff.

PR #51, opened 2016-04-05
Fix depwarn() usage

alyst:
In Julia 0.4.5 `depwarn()` is not exported, the Base module should be explicitly specified.


tshort:
Thanks, @alyst!

PR #55, opened 2016-06-05
Fixed `using DataFramesMeta: @where`

cstjean:
Before, this yielded an error:

```julia
using DataFramesMeta: @where
using RDatasets: dataset

df = dataset("datasets", "iris")

@where df :Species .== "setosa"
```


tshort:
Thanks!

PR #63, opened 2016-09-27
Duplicate definitions of DataFrames.nrow

gustafsson:
Two functions implemented the exact same signature `DataFrames.nrow(cdf::AbstractCompositeDataFrame)`. The last one (the effective one) will break if the `cdf` has 0 columns.


ararslan:
The nightly failure appears unrelated--something about DataArrays.

PR #61, opened 2016-07-29
Some code cleaning

bramtayl:

PR #71, opened 2017-07-03
Some code cleaning

nalimilan:
Continuation of https://github.com/JuliaStats/DataFramesMeta.jl/pull/61, whose parent repository has been removed.

### DIFF
--- a/src/DataFramesMeta.jl
+++ b/src/DataFramesMeta.jl
@@ -17,42 +17,51 @@ include("byrow.jl")
 ##
 ##############################################################################
 
-replace_syms(x, membernames) = x
-function replace_syms(e::Expr, membernames)
-    if e.head == :call && length(e.args) == 2 && e.args[1] == :^
-        return e.args[2]
-    elseif e.head == :.     # special case for :a.b
-        return Expr(e.head, replace_syms(e.args[1], membernames),
-                            typeof(e.args[2]) == Expr && e.args[2].head == :quote ? e.args[2] : replace_syms(e.args[2], membernames))
-    elseif e.head == :call && length(e.args) == 2 && e.args[1] == :_I_
-        nam = :($(e.args[2]))
-        if haskey(membernames, nam)
-            return membernames[nam]
-        else
-            a = gensym()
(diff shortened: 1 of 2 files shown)


bramtayl:
Hmm, looking back on this I've learned a few new things. For example, the multi-line quote blocks will introduce extraneous line number information. Also, the bug that caused issues with macrotools is gone (I think it was the kw vs = distinction), so I could add back in more macrotools.

bramtayl:
However, I'm hard at work on LazyQuery, which when complete should have the same functionality with substantially simpler syntax and more flexibility with dispatch. Any help would be appreciated.

nalimilan:
I just wanted to apply the minimal fixes to get the package working on Julia 0.6. I'd rather leave further improvements for later, especially given that I'm not very good at macro tricks.

ararslan:
I'd also prefer we not add a dependency on MacroTools here

nalimilan:
(Though I must say I'm having quite a hard time fixing the failures, so without help I probably won't succeed...)

nalimilan:
Merging since tests still pass on 0.4 and 0.5.

bramtayl:
It looks like the 0.6 issues are being caused by the current module deprecation. Here is some code from LazyContext that might help:

```
if VERSION < v"0.7.0-DEV.484" # Julia PR #22064
    my_macroexpand(m::Module, ex) = eval(m, :(macroexpand($(Expr(:quote, ex)))))
    __module__ = :( current_module() )
    export @__MODULE__
    """
        @__MODULE__

    Return the current module, for compatibility with older versions of
    Julia.
    """
    macro __MODULE__()
        esc(__module__)
    end
else
    my_macroexpand = macroexpand
end
```
If you want I can put in a PR.

nalimilan:
It would be great if you could make a PR. Fixing things that are probably trivial to somebody familiar with macros like you would probably take me hours if not days to fix.

bramtayl:
Aw shucks!

bramtayl:
Ok I'll take a look this evening

PR #72, opened 2017-07-04
fix some deprecation warnings and errors for 0.6

joshbode:
Replace `&` with `.&` and use updated syntax for `abstract types`


joshbode:
jinx - looks #73 covers most (if not all) of this

PR #73, opened 2017-07-04
updates for 0.6

bramtayl:
Changes for 0.6, plus some very limited cleanups.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: julia
 julia:
-  - 0.4
   - 0.5
   - 0.6
   - nightly
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: julia
 julia:
   - 0.4
   - 0.5
+  - 0.6
   - nightly
 os:
   - linux
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,8 +2,8 @@ environment:
   matrix:
   - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x86/0.5/julia-0.5-latest-win32.exe"
   - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x64/0.5/julia-0.5-latest-win64.exe"
-  - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x86/0.6/julia-0.5-latest-win32.exe"
-  - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x64/0.6/julia-0.5-latest-win64.exe"
+  - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x86/0.6/julia-0.6-latest-win32.exe"
+  - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x64/0.6/julia-0.6-latest-win64.exe"
   - JULIA_URL: "https://julialangnightlies-s3.julialang.org/bin/winnt/x86/julia-latest-win32.exe"
   - JULIA_URL: "https://julialangnightlies-s3.julialang.org/bin/winnt/x64/julia-latest-win64.exe"
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,9 +1,9 @@
 environment:
   matrix:
-  - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x86/0.4/julia-0.4-latest-win32.exe"
-  - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x64/0.4/julia-0.4-latest-win64.exe"
   - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x86/0.5/julia-0.5-latest-win32.exe"
   - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x64/0.5/julia-0.5-latest-win64.exe"
+  - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x86/0.6/julia-0.5-latest-win32.exe"
+  - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x64/0.6/julia-0.5-latest-win64.exe"
   - JULIA_URL: "https://julialangnightlies-s3.julialang.org/bin/winnt/x86/julia-latest-win32.exe"
   - JULIA_URL: "https://julialangnightlies-s3.julialang.org/bin/winnt/x64/julia-latest-win64.exe"
 
--- a/src/DataFramesMeta.jl
+++ b/src/DataFramesMeta.jl
@@ -1,7 +1,6 @@
 module DataFramesMeta
 
 using DataFrames
-import DataFrames
 
 # Basics:
 export @with, @ix, @where, @orderby, @transform, @by, @based_on, @select
@@ -30,7 +29,7 @@ mapexpr(f, e) = Expr(e.head, map(f, e.args)...)
 
 replace_syms!(x, membernames) = x
 replace_syms!(q::QuoteNode, membernames) =
-    replace_syms!( Meta.quot(q.value), membernames)
+    replace_syms!(Meta.quot(q.value), membernames)
 replace_syms!(e::Expr, membernames) =
     if onearg(e, :^)
         e.args[2]
@@ -45,12 +44,13 @@ replace_syms!(e::Expr, membernames) =
     end
 
 protect_replace_syms!(e, membernames) = e
-protect_replace_syms!(e::Expr, membernames) =
+function protect_replace_syms!(e::Expr, membernames)
     if e.head == :quote
         e
     else
         replace_syms!(e, membernames)
     end
+end
 
 function replace_dotted!(e, membernames)
     x_new = replace_syms!(e.args[1], membernames)
@@ -76,7 +76,7 @@ end
 
 function with_anonymous(body)
     d = gensym()
-    anon( d, with_helper(d, body) )
+    anon(d, with_helper(d, body))
 end
 
 """
@@ -180,7 +180,7 @@ and(x, y) =
     end
 
 function where_helper(d, args...)
-    :( $where($d, $(with_anonymous( reduce(and, args)))))
+    :($where($d, $(with_anonymous(reduce(and, args)))))
 end
 
 """
@@ -356,7 +356,7 @@ end
 ##############################################################################
 
 function based_on_helper(x, args...)
-    with_args = :(DataFramesMeta.@with(_DF, DataFrames.DataFrame($(args...))))
+    with_args = :(DataFramesMeta.@with(_DF, DataFrames.DataFrame($(map(replace_equals_with_kw, args)...))))
     :( DataFrames.combine(map(_DF -> $with_args, $x)) )
 end
 
@@ -444,12 +444,13 @@ function Base.select(d::Union{AbstractDataFrame, Associative}; kwargs...)
     return result
 end
 
-replace_equals_with_kw(e) =
+function replace_equals_with_kw(e)
     if e.head == :(=)
         Expr(:kw, e.args[1], e.args[2])
     else
         e
     end
+end
 
 expandargs(x) = x
 expandargs(q::QuoteNode) = Expr(:kw, q.value, q)
--- a/src/DataFramesMeta.jl
+++ b/src/DataFramesMeta.jl
@@ -421,7 +421,7 @@ end
 ##############################################################################
 
 function by_helper(x, what, args...)
-    with_args = with_anonymous(:($(DataFrames.DataFrame)($(replace_equals_with_kw.(args)...))))
+    with_args = with_anonymous(:($(DataFrames.DataFrame)($(map(replace_equals_with_kw, args)...))))
     :( $(DataFrames.by)($x, $what, $with_args) )
 end
 
@@ -489,7 +489,7 @@ expandargs(e::Expr) =
 
 function select_helper(x, args...)
     DF = gensym()
-    select_args = with_helper(DF, :($select($DF, $(expandargs.(args)...))))
+    select_args = with_helper(DF, :($select($DF, $(map(expandargs, args)...))))
     Expr(:let, select_args, :($DF = $x) )
 end
 
--- a/src/DataFramesMeta.jl
+++ b/src/DataFramesMeta.jl
@@ -247,11 +247,7 @@ orderbyconstructor(d) = x -> x
 function orderby_helper(d, args...)
     construct_args = :(DataFramesMeta.orderbyconstructor(_D)($(args...)))
     with_args = :(DataFramesMeta.@with(_DF, $construct_args))
-    quote
-        let _D = $d
-            DataFramesMeta.orderby(_D, _DF -> $with_args)
-        end
-    end
+    Expr(:let, :(DataFramesMeta.orderby(_D, _DF -> $with_args)), :(_D = $d) )
 end
 
 """
--- a/src/DataFramesMeta.jl
+++ b/src/DataFramesMeta.jl
@@ -551,7 +551,7 @@ Base.getindex(gd::GroupedDataFrame, I::AbstractArray{Int}) = GroupedDataFrame(gd
 
 export P, PassThrough
 
-@compat mutable struct PassThrough{T} <: AbstractVector{T}
+type PassThrough{T} <: AbstractVector{T}
     x::AbstractVector{T}
 end
 const P = PassThrough
--- a/src/DataFramesMeta.jl
+++ b/src/DataFramesMeta.jl
@@ -58,31 +58,31 @@ function replace_dotted!(e, membernames)
     Expr(:., x_new, y_new)
 end
 
-anon(x, y) = Expr(:->, Expr(:tuple, x...), y)
-anon(x::Symbol, y) = Expr(:->, x, y)
-
 function with_helper(d, body)
     membernames = Dict{Any, Symbol}()
+    funname = gensym()
     body = replace_syms!(body, membernames)
     if isempty(membernames)
         body
     else
-        funargs = map(keys(membernames)) do key
-            :( $d[$key] )
+        quote
+            function $funname($(values(membernames)...))
+                $body
+            end
+            $funname($(map(keys(membernames)) do key
+                :($d[$key])
+            end...))
         end
-        Expr(:call, anon(values(membernames), body), funargs...)
     end
 end
 
 function with_anonymous(body)
     d = gensym()
-    anon(d, with_helper(d, body))
+    :($d -> $(with_helper(d, body)))
 end
 
 """
-```julia
-@with(d, expr)
-```
+    @with(d, expr)
 
 `@with` allows DataFrame columns or Associative keys to be referenced as symbols.
 
@@ -110,7 +110,7 @@ The following
 becomes
 
 ```julia
-tempfun(a,b) = a + b + 1
+tempfun(a, b) = a + b + 1
 tempfun(d[:a], d[:b])
 ```
 
@@ -122,31 +122,53 @@ variable `expr` rather than a symbol.
 
 ### Examples
 
-```julia
-y = 3
-d = Dict(:s => 3, :y => 44, :d => 5)
+```jldoctest
+julia> using DataFramesMeta
 
-@with(d, :s + :y + y)
+julia> y = 3;
 
-df = DataFrame(x = 1:3, y = [2, 1, 2])
-x = [2, 1, 0]
+julia> d = Dict(:s => 3, :y => 44, :d => 5);
 
-@with(df, :y + 1)
-@with(df, :x + x)  # the two x's are different
+julia> @with(d, :s + :y + y)
+50
 
-x = @with df begin
-    res = 0.0
-    for i in 1:length(:x)
-        res += :x[i] * :y[i]
-    end
-    res
-end
+julia> df = DataFrame(x = 1:3, y = [2, 1, 2]);
 
-@with(df, df[:x .> 1, ^(:y)]) # The ^ means leave the :y alone
+julia> x = [2, 1, 0];
 
-colref = :x
-@with(df, :y + _I_(colref)) # Equivalent to df[:y] + df[colref]
+julia> @with(df, :y + 1)
+3-element DataArrays.DataArray{Int64,1}:
+ 3
+ 2
+ 3
 
+julia> @with(df, :x + x)  # the two x's are different
+3-element DataArrays.DataArray{Int64,1}:
+ 3
+ 3
+ 3
+
+julia> @with df begin
+            res = 0.0
+            for i in 1:length(:x)
+                res += :x[i] * :y[i]
+            end
+            res
+        end
+10.0
+
+julia> @with(df, df[:x .> 1, ^(:y)]) # The ^ means leave the :y alone
+2-element DataArrays.DataArray{Int64,1}:
+ 1
+ 2
+
+julia> colref = :x;
+
+julia> @with(df, :y + _I_(colref)) # Equivalent to df[:y] + df[colref]
+3-element DataArrays.DataArray{Int64,1}:
+ 3
+ 3
+ 5
 ```
 
 `@with` creates a function, so scope within `@with` is a *hard scope*,
@@ -184,9 +206,7 @@ function where_helper(d, args...)
 end
 
 """
-```julia
-@where(d, i...)
-```
+    @where(d, i...)
 
 Select row subsets in AbstractDataFrames or groups in GroupedDataFrames.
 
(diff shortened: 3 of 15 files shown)


bramtayl:
Well, I got 0.5 and 0.6 working; maybe I'll take another stab tomorrow on 0.7

.travis.yml
```diff
@@ -2,6 +2,7 @@ language: julia
 julia:
   - 0.4
   - 0.5
+  - 0.6
```
nalimilan: Can you also add the badge to README.md?

src/DataFramesMeta.jl
```diff
@@ -182,13 +172,15 @@ where(d::AbstractDataFrame, arg) = d[arg, :]
 where(d::AbstractDataFrame, f::Function) = d[f(d), :]
 where(g::GroupedDataFrame, f::Function) = g[Bool[f(x) for x in g]]
 
-collect_ands(x::Expr) = x
-collect_ands(x::Expr, y::Expr) = :($x & $y)
-collect_ands(x::Expr, y...) = :($x & $(collect_ands(y...)))
+and(x, y) =
```
nalimilan: I'd say just drop support for versions before 0.6, anyway we already have a working version for these releases. @ararslan Does that sound OK to you?

src/DataFramesMeta.jl
```diff
@@ -43,27 +46,39 @@ replace_syms!(e::Expr, membernames) =
 
 protect_replace_syms!(e, membernames) = e
 protect_replace_syms!(e::Expr, membernames) =
-    e.head == :quote ? e : replace_syms!(e, membernames)
+    if e.head == :quote
```
nalimilan: The old version as a one-liner was more readable IMHO. If you change that you should use the long `function` syntax, but that doesn't sound like a good idea.

src/DataFramesMeta.jl
```diff
@@ -28,6 +29,8 @@ onearg(e, f) = e.head == :call && length(e.args) == 2 && e.args[1] == f
 mapexpr(f, e) = Expr(e.head, map(f, e.args)...)
 
 replace_syms!(x, membernames) = x
+replace_syms!(q::QuoteNode, membernames) =
+    replace_syms!( Meta.quot(q.value), membernames)
```
nalimilan: Remove space.

src/DataFramesMeta.jl
```diff
@@ -1,6 +1,7 @@
 module DataFramesMeta
 
 using DataFrames
+import DataFrames
```
nalimilan: Why is this needed?

src/DataFramesMeta.jl
```diff
@@ -43,27 +46,39 @@ replace_syms!(e::Expr, membernames) =
...
+            :( $d[$key] )
         end
-        $funname($(funargs...))
+        Expr(:call, anon(values(membernames), body), funargs...)
     end
 end
 
+function with_anonymous(body)
+    d = gensym()
+    anon( d, with_helper(d, body) )
```
nalimilan: Remove spaces too, that's not the style used in the file AFAICT.

src/DataFramesMeta.jl
```diff
@@ -182,13 +172,15 @@ where(d::AbstractDataFrame, arg) = d[arg, :]
...
+    if VERSION < v"0.6.0-"
+        :($x & $y)
+    else
+        :($x .& $y)
+    end
 
 function where_helper(d, args...)
-    with_args = :($DataFramesMeta.@with(_DF, $(collect_ands(args...))))
-    :( $DataFramesMeta.where($d, _DF -> $with_args) )
+    :( $where($d, $(with_anonymous( reduce(and, args)))))
```
nalimilan: Same here (spaces are only used inside `:(`, not after any parenthesis).

src/DataFramesMeta.jl
```diff
@@ -475,17 +448,27 @@ function Base.select(d::Union{AbstractDataFrame, Associative}; kwargs...)
...
 end
 
-
-function select_helper(x, args...)
-    select_args = :(select(_DF; $(expandargs(args)...)))
-    quote
-        let _DF = $x
-            DataFramesMeta.@with(_DF, $select_args)
-        end
+replace_equals_with_kw(e) =
```
nalimilan: Same style remark here and below: either use ternary operator, or `function`.

src/byrow.jl
```diff
@@ -51,31 +51,27 @@ function byrow_helper(df, body)
     (e_body, e_newcols) = byrow_find_newcols(body, Any[])
     e_delimiters = Expr(:(=), :row, :( 1:_N ))
     e_forloop = Expr(:for, e_delimiters, byrow_replace(e_body))
-    return quote
-        _N = length($df[1])
-        _DF = @transform($df, $(e_newcols...))
-        @with _DF $e_forloop
-        _DF
-    end
+    :(_N = length($df[1]); _DF = @transform($df, $(e_newcols...)); @with _DF $e_forloop; _DF)
```
nalimilan: Is this really more readable than the previous form?

src/byrow.jl
```diff
@@ -51,31 +51,27 @@ function byrow_helper(df, body)
...
 Act on a DataFrame row-by-row.
 
 Includes support for control flow and `begin end` blocks. Since the
-"environment" induced by `@byrow! df` is implicitly a single row of `df`, 
+"environment" induced by `@byrow! df` is implicitly a single row of `df`,
 use regular operators and comparisons instead of their elementwise counterparts
 as in `@with`. Note that the scope within `@byrow!` is a hard scope.
 
 `byrow!` also supports special syntax for allocating new columns. The syntax
-`@newcol x::Array{Int}` allocates a new column `:x` with an `Array` container
```
nalimilan: Are you sure this is right? IIRC the `::` syntax was intended.

src/compositedataframe.jl
```diff
@@ -43,14 +43,18 @@ A constructor of an `AbstractCompositeDataFrame` that mimics the `DataFrame`
 constructor.  This returns a composite type (not immutable) that is an
 `AbstractCompositeDataFrame`.
 
-This uses `eval` to create a new type within the current module.
+This uses `eval` to create a new type within the module specified by the
+inmodule keyword.
```
nalimilan: "keyword *argument"*. Use backticks around `inmodule`. Should also mention `inmodule = DataFramesMeta` before `kwargs...` in signatures.

src/compositedataframe.jl
```diff
@@ -43,14 +43,18 @@ A constructor of an `AbstractCompositeDataFrame` that mimics the `DataFrame`
...
 
 ### Arguments
 
 * `columns` : contains the contents of the columns
 * `cnames` : the names of the columns
 * `typename` : the optional name of the type created
 * `kwargs` : the key gives the column names, and the value is the column contents
+* `inmodule = DataFramesMeta` : a keyword argument to specify what module you
+want to define the type in. Consider passing `current_module()` or
```
nalimilan: Align this line and the next to `` `inmodule` `` AFAIK. If you drop support for Julia < 0.6, no need to mention old Julia versions.

test/compositedataframes.jl
```diff
@@ -55,4 +57,26 @@ for x in eachrow(df)
...
+
+@test names(hcat(df, df2)) == [:A, :B, :C, :D]
+@test names(hcat(df2, df)) == [:C, :D, :A, :B]
+@test names(hcat(df, df3)) == [:A, :B, :C, :D]
+
+@test CompositeDataFrame(DataFrame(df)) == df
+@test names(CompositeDataFrame(DataFrame(df), [:C, :D])) == [:C, :D]
+@test row() == nothing
+@test df[1, :A] == 1
+@test names( df[[1, 2]] ) == [:A, :B]
```
nalimilan: No spaces (here and below). `[1, 2]` could be `1:2`.

test/dataframes.jl
```diff
@@ -36,8 +37,8 @@ idx2 = :B
 @test  @where(df, :A .> x)          == df[df[:A] .> x,:]
 @test  @where(df, :B .> x)          == df[df[:B] .> x,:]
 @test  @where(df, :A .> :B)         == df[df[:A] .> df[:B],:]
-@test  @where(df, :A .> 1, :B .> 1) == df[(df[:A] .> 1) & (df[:B] .> 1),:]
-@test  @where(df, :A .> 1, :A .< 4, :B .> 1) == df[(df[:A] .> 1) & (df[:B] .> 1),:]
+@test  @where(df, :A .> 1, :B .> 1) == df[map(&, df[:A] .> 1, df[:B] .> 1),:]
```
nalimilan: If you drop support for Julia < 0.6, use `.&`.

test/dataframes.jl
```diff
@@ -53,13 +54,15 @@ y = 0
...
-    @newcol colY::DataArray{Int}
+    @newcol colX = Array{Float64}(_N)
+    @newcol colY = DataArray(Int, _N)
     :colX = :B == 2 ? pi * :A : :B
     if :A > 1
         :colY = :A * :B
     end
 end
+
+import MacroTools
```
nalimilan: ?

test/runtests.jl
```diff
@@ -1,4 +1,4 @@
-fatalerrors = false
+fatalerrors = true
```
nalimilan: Probably better keep `false` here, as it allows running all tests to see all errors, and only fail at the end.

.travis.yml
nalimilan: Actually the badge is already there.

src/DataFramesMeta.jl
bramtayl: I like if else better than ternary because ternary gives artificially inflated coverage.

src/DataFramesMeta.jl
bramtayl: Oh no! I spent all that time getting it working on 0.4 :(

src/byrow.jl
bramtayl: In fact, :: (or any two arg expr) should work as well. However, I did have to change it so that the user needs to manually use _N. This was because the deprecation of the Array(T, size) syntax.

test/compositedataframes.jl
bramtayl: I was trying hit coverage of the `Base.getindex{T <: DataFrames.ColumnIndex}(cdf::AbstractCompositeDataFrame, col_inds::AbstractVector{T})` constructor

src/byrow.jl
bramtayl: This is on purpose. The multi-line quote blocks will introduce extraneous line number information into user code.

src/compositedataframe.jl
bramtayl: `current_module()` is only deprecated as of 0.7.

src/byrow.jl
nalimilan: I don't understand. What is `_N`? Wouldn't `::Vector{Int}` work since we're dealing with columns?

src/compositedataframe.jl
nalimilan: OK. Would make sense to mention the versions then, since in a few months the old version won't be relevant at all.

src/byrow.jl
nalimilan: Just curious, what kind of information?

src/DataFramesMeta.jl
nalimilan: I wouldn't say that's a sufficient reason. If think it's been discussed in Base already.

Anyway, when you use `if` blocks, please also use the long `function` form (there are several cases like this), and keep the short syntax for one-liners.

src/compositedataframe.jl
```diff
@@ -33,24 +33,28 @@ row() = nothing
 
 """
 ```julia
-CompositeDataFrame(columns::Vector{Any}, cnames::Vector{Symbol})
-CompositeDataFrame(columns::Vector{Any}, cnames::Vector{Symbol}, typename::Symbol)
-CompositeDataFrame(; kwargs...)
-CompositeDataFrame(typename::Symbol; kwargs...)
+CompositeDataFrame(columns::Vector{Any}, cnames::Vector{Symbol}; inmodule = DataFramesMeta)
+CompositeDataFrame(columns::Vector{Any}, cnames::Vector{Symbol}, typename::Symbol; inmodule = DataFramesMeta)
```
nalimilan: This line needs to be wrapped at 92 characters.

test/grouping.jl
```diff
@@ -14,9 +14,11 @@ g = groupby(d, :x)
 @test  DataFrame(@where(g, length(:x) > 5))[:n][1:3] == @data [5, 6, 7]
 
 @test  DataFrame(orderby(g, x -> mean(x[:n]))) == DataFrame(@orderby(g, mean(:n)))
-#@test  DataFrames.based_on(@orderby(g, mean(:n)), x -> x[1,:x])[:x1] == [3,1,2]
```
nalimilan: Why remove this? Is it tested elsewhere? Else, better keep it as it might reflect a feature which is supposed to work but doesn't.

bramtayl: I added in a `@based_on` test which should cover this.

src/byrow.jl
nalimilan: Scratch that, `_N` is not the dimension of the array as I first thought. Though it still don't understand the change.

bramtayl: Can you initialize an empty column in a dataframe (or subdataframe) of a certain type without using the length of the dataframe?

bramtayl: The `@newcol` feature might be more trouble than it's worth; it could simply be accomplished by a transform. Giving transform (or actually, any with function) access to _N would be useful, similar to how n() is used in dplyr.

nalimilan: IIRC the point was to support a syntax similar to SAS's `data` step, which is quite practical when doing many operations. Having to specify `_N` manually goes against the goal of having a concise syntax, and that doesn't make a lot of sense since "new column" necessarily means "a vector with a length equal to the number of rows of the data frame". Do you have a use case in mind where `_N` would be useful in `@byrow!`? Note that this command is not as flexible as the other ones.

bramtayl: Well the reason I changed it in the first place is the deprecation of `Array(T, size)`. The macro in the previous version was taking `Container{Eltype}` and turning it into `Container(Eltype, _N)`.  This used to work for both Arrays and DataArrays. Now I could have changed it to make instead `Container{Eltype}(_N)`, which would have been the correct syntax for Arrays but not for  DataArrays. Tldr. the macro was making unwarranted assumptions about constructors, which was why it was able to leave out _N.

src/byrow.jl
bramtayl: Well, user code after a macroexpand would contain line numbers from the byrow.jl file. Actually, I guess that's probably a good thing for backtraces?, but maybe due to a bug, these lines would show up as uncovered. Right now coverage for this pull request is up to like 97% in part because of changes I made like this.

src/byrow.jl
nalimilan: OK, but that's a bug in DataArrays, so let's use `Container{Eltype}(_N)`, avoid testing this with `DataArray`, and this the problem in that package.

src/byrow.jl
nalimilan: Yes, I'd rather get good backtraces than nominally high coverage. It's not a big deal if these lines appear as uncovered while they are covered. That's a bug which should be fixed elsehwere, let's not make the codebase worse because of this.

bramtayl: P.S. this macro (as with all macros here) are completely escaped such that both _N and _DF will end up in user code. I was originally thinking this was a bug, because a user could get unexpected results when referencing these variables, but in fact, an aspiring user could use this strategically. Maybe it deserves documentation?

nalimilan: Yeah, ideally they should be documented (unless we escape them, but keeping them doesn't sound too dangerous and could be useful).

bramtayl: The best of both worlds would be to replace all references to line numbers in generated code with the line numbers in user code. (Hopefully) if there's an error to find it should be in user code and not in package code. I have a NumberedLines package which can do this. Dunno if we want to add another dependency.

nalimilan: That would clearly be ideal, but currently the future of this package is not completely clear AFAIK, so I'm not sure it's worth spending too much time on new features before knowing where we're heading to. Though if the dependency is not too large it could be worth it.

bramtayl: This would also require __source__ which is only available in 0.7

bramtayl:
Got to pass on 0.7. Looks like `@>` from Lazy is broken.

src/DataFramesMeta.jl
```diff
@@ -207,18 +219,72 @@ Multiple `i` expressions are "and-ed" together.
...
+│ Row │ x │ y │
+├─────┼───┼───┤
+│ 1   │ 2 │ 1 │
+│ 2   │ 3 │ 2 │
+
+julia> @where(df, :x .> x, :y .== 3)
+0×2 DataFrames.DataFrame
+
+julia> d = DataFrame(
+            n = 1:20,
```
nalimilan: That's an unusual way of wrapping lines. I'd rather keep `n` on the first line, but split `x` over multiple lines. Same remark in the next docstring.

src/DataFramesMeta.jl
nalimilan: Please switching this to the long `function` form.

src/DataFramesMeta.jl
```diff
@@ -253,19 +319,14 @@ orderbyconstructor(d::AbstractDataFrame) = (x...) -> DataFrame(Any[x...])
...
 function orderby_helper(d, args...)
-    construct_args = :(DataFramesMeta.orderbyconstructor(_D)($(args...)))
-    with_args = :(DataFramesMeta.@with(_DF, $construct_args))
-    quote
-        let _D = $d
-            DataFramesMeta.orderby(_D, _DF -> $with_args)
-        end
-    end
+    _D = gensym()
+    :(let $_D = $d
```
nalimilan: Why not keep `quote... end`? Looks cleaner.

src/DataFramesMeta.jl
```diff
@@ -274,14 +335,39 @@ Sort by criteria. Normally used to sort groups in GroupedDataFrames.
...
+```jldoctest
+julia> using DataFrames, DataFramesMeta
+
+julia> d = DataFrame(
+            n = 1:20,
+            x = [3, 3, 3, 3, 1, 1, 1, 2, 1, 1, 2, 1, 1, 2, 2, 2, 3, 1, 1, 2]);
+
+julia> g = groupby(d, :x);
+
+julia> @macroexpand @orderby(g, mean(:n))
```
nalimilan: What does `@macroexpand` do here?

src/DataFramesMeta.jl
```diff
@@ -320,22 +406,14 @@ function transform(g::GroupedDataFrame; kwargs...)
...
-    kw.args[2] = :( _DF -> DataFramesMeta.@with(_DF, $(kw.args[2]) ) )
-    kw
-end
-
 function transform_helper(x, args...)
-    # convert each kw arg value to: _DF -> @with(_DF, arg)
-    newargs = [args...]
-    map!(convert_kw!, newargs)
-    :( transform($x, $(newargs...)) )
+    :( $transform($x, $(map(args) do kw
```
nalimilan: `quote... end` could make sense here too.

src/DataFramesMeta.jl
```diff
@@ -411,14 +536,12 @@ end
 ##############################################################################
 
 function by_helper(x, what, args...)
-    with_args = :(DataFramesMeta.@with(_DF, DataFrames.DataFrame($(args...))))
-    :( DataFrames.by($x, $what, _DF -> $with_args) )
+    :($by($x, $what,
+        $(with_anonymous(:($DataFrame($(map(replace_equals_with_kw, args)...)))))))
```
nalimilan: Align with `$x`.

src/DataFramesMeta.jl
```diff
@@ -475,22 +629,33 @@ function Base.select(d::Union{AbstractDataFrame, Associative}; kwargs...)
...
+    if e.head == :(=)
+        Expr(:kw, e.args[1], e.args[2])
+    else
+        e
     end
 end
 
+expandargs(x) = x
+expandargs(q::QuoteNode) = Expr(:kw, q.value, q)
+expandargs(e::Expr) =
```
nalimilan: Use long `function` form here too.

src/DataFramesMeta.jl
```diff
@@ -475,22 +629,33 @@ function Base.select(d::Union{AbstractDataFrame, Associative}; kwargs...)
...
+    if e.head == :quote
+        Expr(:kw, e.args[1], e)
+    else
+        replace_equals_with_kw(e)
+    end
 
+function select_helper(x, args...)
+    DF = gensym()
+    select_args = with_helper(DF, :($select($DF, $(map(expandargs, args)...))))
+    :(let $DF = $x
```
nalimilan: `quote... end`.

src/byrow.jl
```diff
@@ -28,18 +28,23 @@ byrow_replace(x) = x
 
 function byrow_find_newcols(e::Expr, newcol_decl)
     if e.head == :macrocall && e.args[1] == Symbol("@newcol")
-        ea = e.args[2]
+        ea =
```
nalimilan: It's more common to write this as:
```julia
if ...
    ea = ...
else
   ea ...
end
```

src/byrow.jl
```diff
@@ -48,34 +53,34 @@ end
...
 use regular operators and comparisons instead of their elementwise counterparts
 as in `@with`. Note that the scope within `@byrow!` is a hard scope.
 
 `byrow!` also supports special syntax for allocating new columns. The syntax
 `@newcol x::Array{Int}` allocates a new column `:x` with an `Array` container
 with eltype `Int`. Note that the returned `AbstractDataFrame` includes these new
 columns, but the original `d` is not affected. This feature makes it easier to
-use `byrow!` for data transformations.
+use `byrow!` for data transformations. `_N` is introduced to represent the
+length of the dataframe, `_D` represents the dataframe including added columns,
```
nalimilan: "data frame". "current row" means index of the current row?

src/byrow.jl
```diff
@@ -89,16 +94,33 @@ The modified `AbstractDataFrame`.
...
-df2 = @byrow! df begin
-    @newcol colX::Array{Float64}
-    :colX = :B == 2 ? pi * :A : :B
-end
+julia> using DataFrames, DataFramesMeta
+
+julia> df = DataFrame(A = 1:3, B = [2, 1, 2]);
+
+julia> let x = 0
+            @byrow!(df, if :A + :B == 3; x += 1 end)  #  This doesn't work without the let
```
nalimilan: Would be clearer with `begin... end` block on multiple lines. Same below.

src/byrow.jl
```diff
@@ -89,16 +94,33 @@ The modified `AbstractDataFrame`.
...
+julia> @byrow! df if :A > :B; :A = 0 end
+3×2 DataFrames.DataFrame
+│ Row │ A │ B │
+├─────┼───┼───┤
+│ 1   │ 1 │ 2 │
+│ 2   │ 0 │ 1 │
+│ 3   │ 0 │ 2 │
+
+julia> df2 = @byrow! df begin
+            @newcol colX::Array{Float64}
```
nalimilan: Four-space indentation.

src/compositedataframe.jl
```diff
@@ -23,76 +23,89 @@ of the field in `cdf`.
...
 row() = nothing
 
 """
-```julia
-CompositeDataFrame(columns::Vector{Any}, cnames::Vector{Symbol})
-CompositeDataFrame(columns::Vector{Any}, cnames::Vector{Symbol}, typename::Symbol)
-CompositeDataFrame(; kwargs...)
-CompositeDataFrame(typename::Symbol; kwargs...)
-```
+    CompositeDataFrame(columns::Vector{Any}, cnames::Vector{Symbol}; inmodule = DataFramesMeta)
```
nalimilan: Wrap lines at 92 chars.

src/linqmacro.jl
```diff
@@ -27,22 +26,40 @@ The following embedded function calls are equivalent to their macro version:
...
-    by(:b, meanX = mean(:x), meanY = mean(:y)) |>
-    orderby(:meanX) |>
-    select(:meanX, :meanY, var = :b)
+```jldoctest
+julia> using DataFrames, DataFramesMeta
+
+julia> n = 100;
+
+julia> df = DataFrame(a = rand(1:3, n),
+            b = ["a","b","c","d"][rand(1:4, n)],
```
nalimilan: Align with `a`.

src/linqmacro.jl
```diff
@@ -27,22 +26,40 @@ The following embedded function calls are equivalent to their macro version:
...
+julia> @linq select(orderby(x1, :b, -:meanX), var = :b, :meanX, :meanY)
+3×3 DataFrames.DataFrame
+│ Row │ var │ meanX    │ meanY   │
+├─────┼─────┼──────────┼─────────┤
+│ 1   │ "a" │ 0.665682 │ 6.65682 │
+│ 2   │ "b" │ 0.617848 │ 6.17848 │
+│ 3   │ "d" │ 0.568289 │ 5.68289 │
+
+julia> @linq df |>
+            transform(y = 10 * :x) |>
```
nalimilan: Four-space indent.

.travis.yml
```diff
@@ -1,7 +1,7 @@
 language: julia
 julia:
-  - 0.4
```
nalimilan: Also remove the README.md badge.

src/compositedataframe.jl
```diff
@@ -38,7 +38,8 @@ row() = nothing
 
 """
     CompositeDataFrame(columns::Vector{Any}, cnames::Vector{Symbol}; inmodule = DataFramesMeta)
-    CompositeDataFrame(columns::Vector{Any}, cnames::Vector{Symbol}, typename::Symbol; inmodule = DataFramesMeta)
+    CompositeDataFrame(columns::Vector{Any}, cnames::Vector{Symbol}, typename::Symbol;
+        inmodule = DataFramesMeta)
```
nalimilan: Should be aligned with `column`.

nalimilan:
Hmm, and you should remove Julia 0.4. from appveyor.yml, which will make the CI turn green.

src/DataFramesMeta.jl
```diff
@@ -182,19 +194,20 @@ where(d::AbstractDataFrame, arg) = d[arg, :]
...
-collect_ands(x::Expr) = x
-collect_ands(x::Expr, y::Expr) = :($x & $y)
-collect_ands(x::Expr, y...) = :($x & $(collect_ands(y...)))
+function and(x, y)
+    if VERSION < v"0.6.0-"
+        :($x & $y)
+    else
+        :($x .& $y)
+    end
+end
```
ararslan: It would be better to move the conditional outside of the function definition, i.e.
```julia
if VERSION < v"0.6.0-"
    and(x, y) = :($x & $y)
else
    and(x, y) = :($x .& $y)
end
```
Otherwise the version check is performed every time the function is called.

src/byrow.jl
```diff
@@ -28,18 +28,22 @@ byrow_replace(x) = x
 
 function byrow_find_newcols(e::Expr, newcol_decl)
     if e.head == :macrocall && e.args[1] == Symbol("@newcol")
-        ea = e.args[2]
+        if VERSION < v"0.7-"
```
ararslan: It would be good to be more exact with the version cutoff used. What changed in Base such that the number of arguments in the expression increased? If you can find the PR that introduced that change, you can use contrib/commit-name.sh in the Julia repo to get the exact version that corresponds to that change.

ararslan: Also since this is in a function (but in this case the conditional is less easily separable, the condition should have an `@static` to evaluate the branch once at parse time.

nalimilan: Probably https://github.com/JuliaLang/julia/commit/3c3ced4153cdc2c75319b41155c38ad1f2545fc2? That's `0.7.0-DEV.481`.

bramtayl: Actually I think it was this: https://github.com/JuliaLang/julia/pull/21746

bramtayl: Try `:(@newcol Array{Int}).args` in 0.7

nalimilan: That would be 0.7.0-DEV.357 then.

nalimilan:
Can you also remove Julia 0.4 from appveyor.yml?

appveyor.yml
```diff
@@ -1,9 +1,9 @@
 environment:
   matrix:
-  - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x86/0.4/julia-0.4-latest-win32.exe"
-  - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x64/0.4/julia-0.4-latest-win64.exe"
   - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x86/0.5/julia-0.5-latest-win32.exe"
   - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x64/0.5/julia-0.5-latest-win64.exe"
+  - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x86/0.6/julia-0.5-latest-win32.exe"
```
nalimilan: Typo `0.5` in the filename... :-/

nalimilan:
Thanks! It's really cool to have DataFramesMeta working on Julia 0.6.

bramtayl:
Yeah for sure. I think there's a number of improvements possible going forward. For example, it seems easy enough to widen the scope to include datatables as well. Also, I wonder if anything could be done about type stability/other optimizations for the grouped operations. I've been consider adding DataFramesMeta as a (second) backend to LazyQuery. For some things, like grouped operations, DataFramesMeta seems much more intuitive than Query (which currently doesn't have a good alternative to dplyr summarize).

src/compositedataframe.jl
```diff
@@ -9,9 +9,9 @@ export AbstractCompositeDataFrame, AbstractCompositeDataFrameRow,
 An abstract type that is an `AbstractDataFrame`. Each type that inherits from
 this is expected to be a type-stable data frame.
 """
-abstract AbstractCompositeDataFrame <: AbstractDataFrame
+@compat abstract type AbstractCompositeDataFrame <: AbstractDataFrame end
```
tkelman: this requires Compat 0.17

PR #75, opened 2017-07-12
Remove type piracy on Base.select(::Associative)

tkelman:
is this needed to make things work here? If so, can that be refactored?

### DIFF
--- a/src/DataFramesMeta.jl
+++ b/src/DataFramesMeta.jl
@@ -296,7 +296,7 @@ end
 ##
 ##############################################################################
 
-Base.select(d::AbstractDataFrame, arg) = d[arg]
+select(d::AbstractDataFrame, arg) = d[arg]
 
 
 ##############################################################################
@@ -622,7 +622,7 @@ end
 ##
 ##############################################################################
 
-function Base.select(d::Union{AbstractDataFrame, Associative}; kwargs...)
+function select(d::Union{AbstractDataFrame, Associative}; kwargs...)
     result = typeof(d)()
     for (k, v) in kwargs
         result[k] = v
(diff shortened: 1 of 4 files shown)


tkelman:
Anyone know what this is used for here? Would it work to use a package-local DataFramesMeta.select without importing it? Pretty fishy to be relying on behavior of Associative that Base doesn't define.

nalimilan:
I think the type piracy is due to a more fundamental function piracy: `select` is used in the SQL sense in this package, which has nothing to do with `Base.select`. That's not a big deal for data frames, but it's more annoying for `Dict`, and yet the package supports operations on both types.

A solution would be to unexport all functions like `select`, since IIRC they are mostly intended to be used via macros like `@select` or inside a `@linq` block (which could automatically add the required prefix).

nalimilan:
Great, I wouldn't have expected this to be so easy! But maybe we should apply this change more systematically? It would feel weird to be unable to access `select` directly, but that it would work for other functions.

tkelman:
There actually doesn't seem to be a whole lot of method extension here - other than these `select`s, https://github.com/JuliaStats/DataFramesMeta.jl/search?utf8=%E2%9C%93&q=+Base&type= shows `names`, `hcat`, `values` in a test (which is maybe a bit dubious too). Since the results shown per file are limited, we can see there are also normal things like `getindex`, `size`, `start`, `next`, `done`, `length`, `map`, all of which look fine. And I don't see any `import` or `importall` in this package.

nalimilan:
I just meant that `where`, `orderby`, and `transform` should be unexported too.

tkelman:
Like that?

tkelman:
May be worth opening an issue in Base and think about moving select into the Sort module?

tshort:
I had intended on exposing the functions in the API, but I'm good with removing that and limiting the API to the macro use. I'm also fine with removing support for Dicts.

nalimilan:
Thanks for commenting @tshort. Let's unexport these functions for now, and hopefully we can rename `Base.select` so that it's available for packages in 1.0.

PR #76, opened 2017-07-19
Modified collect_ands in @where to work for julia 0.6

sylvaticus:
In julia 0.6 is no longer possible to use Array & Array, you need to use instead Array .& Array..
As a consequence the @where macro was no longer working where more of one filter argument was given. This patch solves the problem introducing `.&` in the collect_ands statements.

(PS: I modified only the collect_ands.. I don't know why the patch also include some other whitespace changes)

### DIFF
--- a/src/DataFramesMeta.jl
+++ b/src/DataFramesMeta.jl
@@ -64,7 +64,7 @@ end
 
 ### Arguments
 
-* `d` : an AbstractDataFrame or Associative type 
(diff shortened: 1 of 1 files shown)


nalimilan:
Thanks, but I suspect this is already fixed on master, which explains why there are conflicts. Can you confirm that's the case?

sylvaticus:
yes, thank you.. I should have pulled master before.. great to know it has been already fixed..

PR #79, opened 2017-11-28
DataFrames 0.11 support

nalimilan:
Also drop support for Julia 0.5 and NullableArrays.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: julia
 julia:
-  - 0.5
   - 0.6
   - nightly
 os:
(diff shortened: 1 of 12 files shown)


nalimilan:
I don't understand why tests fail, as they pass locally on 0.6. Ideas?

ararslan:
Wow, excellent failure message: `ERROR: LoadError: "Tests failed"`. A+

tshort:
I like! I like! I love the way DataFrames now supports more vector types, including `Vector`.

The crappy error message for failed tests is my fault. I can't imagine why I did it that way.

I'll see if I can reproduce test failures locally (but no guarantees as all I've got is a windows laptop on travel).

tshort:
I think `test/REQUIRE` needs `DataArrays` added.

nalimilan:
Cool, thanks! Tests pass on 0.6 now.

The best solution will be to move to testsets eventually.

Issue #82, opened 2018-01-16
How to mitigate deprecation warning when using orderby?


JockLawrie:
Hi there,

I'm using Julia 0.6 and DataFramesMeta 0.3.0.
The following code produces a deprecation warning.
I'm not sure what I need to change to avoid the warning. Any ideas?

```julia
using DataFrames
using DataFramesMeta

tbl = DataFrame(a = [22,11,33])
@linq tbl |> orderby(:a)
```

ararslan:
What's the warning text and backtrace?

JockLawrie:
Thanks for the quick response.

```julia
julia> @linq tbl |> orderby(:a)
WARNING: calling vector of vectors constructor without passing column names is deprecated
Stacktrace:
 [1] depwarn(::String, ::Symbol) at ./deprecated.jl:70
 [2] Type at /home/jock/.julia/v0.6/DataFrames/src/deprecated.jl:4 [inlined]
 [3] #35 at /home/jock/.julia/v0.6/DataFramesMeta/src/DataFramesMeta.jl:316 [inlined]
 [4] ##739 at /home/jock/.julia/v0.6/DataFramesMeta/src/DataFramesMeta.jl:70 [inlined]
 [5] (::##6#7{DataFrames.DataFrame})(::DataFrames.DataFrame) at /home/jock/.julia/v0.6/DataFramesMeta/src/DataFramesMeta.jl:72
 [6] orderby(::DataFrames.DataFrame, ::##6#7{DataFrames.DataFrame}) at /home/jock/.julia/v0.6/DataFramesMeta/src/DataFramesMeta.jl:313
 [7] eval(::Module, ::Any) at ./boot.jl:235
 [8] eval_user_input(::Any, ::Base.REPL.REPLBackend) at ./REPL.jl:66
 [9] macro expansion at ./REPL.jl:97 [inlined]
 [10] (::Base.REPL.##1#2{Base.REPL.REPLBackend})() at ./event.jl:73
while loading no file, in expression starting on line 323
```

ararslan:
Looks like the warning is coming from DataFrames, and something in DataFramesMeta is calling a constructor that was deprecated in https://github.com/JuliaData/DataFrames.jl/pull/1325.

JockLawrie:
FYI - I'm using DataFrames 0.11.4.
Anything I need to do?
Thanks again.

JockLawrie:
OK great, thanks!

[...]

PR #83, opened 2018-01-16
fix depwarn in orderby

bkamins:
A quickfix to reseolve #82. It calls an internal function of `DataFrames` for consistency but I assume we should allowed it here as it is a closely coupled package.
In general the reason for the problem was that we actually do not care about column names in this case.


bkamins:
CI failures seem unrelated

nalimilan:
Thanks. If we really don't care about names, why not generate a sequence of names? I agree it's not a big deal to call an internal DataFrames method, but that's even better if we can avoid it.

bkamins:
OK - changing to `Symbol.(1:length(x))` 😄.

PR #90, opened 2018-03-07
various fixes for 0.7

ExpandingMan:
These are a bunch of simple fixes for 0.7.  There are still some outstanding problems.  In particular, wrong variables seem to get substituted in some cases.  It's almost as if DataFramesMeta is depending on ordering of `Expr` args somewhere where it shouldn't be, but it wasn't easy to find the problem (and I didn't).  This is a good first step at least.

### DIFF
--- a/src/compositedataframe.jl
+++ b/src/compositedataframe.jl
@@ -170,7 +170,7 @@ This iterator is created by calling `eachrow(df)` where `df` is an
 
 See also `row(cdf, i)`.
 """
-immutable CDFRowIterator{T <: AbstractCompositeDataFrame}
+struct CDFRowIterator{T <: AbstractCompositeDataFrame}
     df::T
     len::Int
 end
--- a/src/compositedataframe.jl
+++ b/src/compositedataframe.jl
@@ -78,10 +78,10 @@ function CompositeDataFrame(columns::Vector{Any},
                             inmodule = DataFramesMeta)
     rowtypename = Symbol(typename, "Row")
     # TODO: length checks
-    type_definition = :(type $typename <: AbstractCompositeDataFrame end)
+    type_definition = :(mutable struct $typename <: AbstractCompositeDataFrame end)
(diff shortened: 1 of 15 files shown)


src/compositedataframe.jl
```diff
@@ -138,7 +138,10 @@ DataFrames.index(cdf::AbstractCompositeDataFrame) = DataFrames.Index(names(cdf))
 #########################################
 
 Base.getindex(cdf::AbstractCompositeDataFrame, col_inds::DataFrames.ColumnIndex) = getfield(cdf, col_inds)
-Base.getindex{T <: DataFrames.ColumnIndex}(cdf::AbstractCompositeDataFrame, col_inds::AbstractVector{T}) = CompositeDataFrame(Any[ getfield(cdf, col_inds[i]) for i = 1:length(col_inds) ], names(cdf)[col_inds])
+function Base.getindex(cdf::AbstractCompositeDataFrame, col_inds::AbstractVector{T}
+                      ) where {T<:DataFrames.ColumnIndex}
```
nalimilan: I'd keep this on the previous line, I don't think it's common to start a line with `)`.

test/dataframes.jl
```diff
@@ -1,17 +1,17 @@
...
 using DataFramesMeta
 
 df = DataFrame(A = 1:3, B = [2, 1, 2])
 
 x = [2, 1, 0]
 
-@test  @with(df, :A + 1)   ==  df[:A] + 1
-@test  @with(df, :A + :B)  ==  df[:A] + df[:B]
-@test  @with(df, :A + x)   ==  df[:A] + x
+@test  @with(df, :A .+ 1)   ==  df[:A] + 1
```
nalimilan: Shouldn't `.+` be used on the RHS too?

tshort:
I'm sorry for letting this slip. Is this okay to merge now? I see it still passes on Travis on v0.6.

nalimilan:
Should still be OK, if conflicts can be fixed.

Issue #94, opened 2018-05-27
Use of qualified function names in @linq


bkamins:
Following https://stackoverflow.com/questions/50547688/julia-using-groupby-with-dataframesmeta-and-lazy.

Would it be possible to add support for passing qualified function names to `@linq` macro?

tshort:
If I'm reading it right, `@linq` works. It's use of Lazy that has the conflict. The main options I see are:

1. Remove all uses of Lazy from the docs/examples, and encourage `@linq`.

2. Pull in `@>` from Lazy.

bkamins:
Yes - I do not say that `@linq` is broken. I am making a feature request 😄. Incidentally this request is needed with Lazy.jl because of `groupby` name clash.

Let me explain the problem more explicitly. Assume you do not want to pollute a namespace for instance (which can be expected in production code) and use `import` instead of `using`. How would you use `@linq` then?

```
julia> import DataFrames, DataFramesMeta

julia> x = DataFrames.DataFrame(rand(3,4))
3×4 DataFrames.DataFrame
│ Row │ x1       │ x2       │ x3       │ x4       │
├─────┼──────────┼──────────┼──────────┼──────────┤
│ 1   │ 0.789172 │ 0.2509   │ 0.107305 │ 0.504798 │
│ 2   │ 0.173546 │ 0.847316 │ 0.292337 │ 0.880935 │
│ 3   │ 0.304413 │ 0.460651 │ 0.368497 │ 0.581835 │

julia> DataFramesMeta.@linq x |> DataFrames.groupby(:x1)
ERROR: MethodError: Cannot `convert` an object of type Expr to an object of type DataFramesMeta.SymbolParameter
This may have arisen from a call to the constructor DataFramesMeta.SymbolParameter(...),
since type constructors fall back to convert methods.
```

whereas with `using` you have:

```
julia> using DataFrames, DataFramesMeta

julia> @linq x |> groupby(:x1)
DataFrames.GroupedDataFrame  3 groups with keys: Symbol[:x1]
First Group:
1×4 DataFrames.SubDataFrame{Array{Int64,1}}
│ Row │ x1       │ x2     │ x3       │ x4       │
├─────┼──────────┼────────┼──────────┼──────────┤
│ 1   │ 0.789172 │ 0.2509 │ 0.107305 │ 0.504798 │
⋮
Last Group:
1×4 DataFrames.SubDataFrame{Array{Int64,1}}
[...]

PR #95, opened 2018-05-27
make linq accept qualified function names

bkamins:
A proposal of a solution to #94.


bkamins:
@tshort 
Tests on Julia 0.6.2 go through and on Julia 0.7 fail, but I guess this is a general problem (and not an issue introduced by this PR - right?).

tshort:
Nice! Looks good to me. Yes, tests are failing on nightly Julia.

Issue #98, opened 2018-07-30
Grouped by issue with type conversion


pdeffebach:
```
using DataFrames, DataFramesMeta
df = DataFrame(
    id = [1,1,1,2,2,2,3,3,3],
    a = ["foo", "bar", "baz","foo", "bar", "baz","foo", "bar", "baz"],
    b = ["x", missing, "z","y", missing, missing,"x", missing, "z"],
    c = [missing, "w", missing, missing, "w", missing, missing,"w", missing],
    d = [missing, missing, 10, missing, missing, 20, missing, missing, 15],
    c_number = [missing, 1, missing, missing, 1, missing, missing, 1, missing],
    )

julia> df2 = @linq df |>
       groupby(:id) |>
       @transform(newc = maximum(skipmissing(:c)))
ERROR: MethodError: Cannot `convert` an object of type String to an object of type Char
This may have arisen from a call to the constructor Char(...),
since type constructors fall back to convert methods.
Stacktrace:
 [1] setindex!(::Array{Char,1}, ::String, ::UnitRange{Int64}) at .\array.jl:591
 [2] insert_multiple_entries!(::DataFrames.DataFrame, ::String, ::UnitRange{Int64}, ::Symbol) at C:\Users\Pdeffebach\.julia\v0.6\DataFrames\src\dataframe\dataframe.jl:373
 [3] setindex!(::DataFrames.DataFrame, ::String, ::UnitRange{Int64}, ::Symbol) at C:\Users\Pdeffebach\.julia\v0.6\DataFrames\src\dataframe\dataframe.jl:511
 [4] #transform#40(::Array{Any,1}, ::Function, ::DataFrames.GroupedDataFrame) at C:\Users\Pdeffebach\.julia\v0.6\DataFramesMeta\src\DataFramesMeta.jl:400
 [5] (::DataFramesMeta.#kw##transform)(::Array{Any,1}, ::DataFramesMeta.#transform, ::DataFrames.GroupedDataFrame) at .\<missing>:0

julia> df2 = @linq df |>
       groupby(:id) |>
       @transform(newc = maximum(skipmissing(:c_number)))
9×7 DataFrames.DataFrame
│ Row │ id │ a   │ b       │ c       │ d       │ c_number │ newc │
├─────┼────┼─────┼─────────┼─────────┼─────────┼──────────┼──────┤
│ 1   │ 1  │ foo │ x       │ missing │ missing │ missing  │ 1    │
[...]

PR #99, opened 2018-08-02
Allow for scalars in a grouped @transform

pdeffebach:
Resolves issue #98 with a one-liner. 

This way we can spread grouped vector-to-scalar operations across groups easily. This behavior is already feasible, but it threw an error with strings because `eltype(String)` is `Char`. Now it will allocate a vector of type `String`. 

Next step is  to automatically expand the type of the vector if a group is entirely missing. Note that this would have thrown an error before this PR as well.

### DIFF
--- a/src/DataFramesMeta.jl
+++ b/src/DataFramesMeta.jl
@@ -393,7 +393,8 @@ function transform(g::GroupedDataFrame; kwargs...)
     idx1 = [1; 1 + idx2[1:end-1]]
     for (k, v) in kwargs
         first = v(g[1])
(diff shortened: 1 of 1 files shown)


bkamins:
Actually given https://discourse.julialang.org/t/best-way-to-append-vectors-of-various-types/12837 maybe we should fix also this in one shot?

This would require precomputing all results and finding their common type. DataFrames.jl uses internal `promote_col_type` to achieve it.

pdeffebach:
Closed in favor of #101 This has that implementation.

PR #101, opened 2018-08-02
Change the way grouped transforms work

pdeffebach:
With this PR, now we allocate a vector using `reduce(v ...generator for all operations on groups...)` 

however I still have an error with type promotions. Consider the following example:

```
 df = DataFrame(a = [1,2,3, missing, missing], b = [1,1,1,2,2])
g = groupby(df, :b)
 @transform(g, x = mean(:a))
```

It seems like `reduce(append!, t(ig) for ig in g` should automatically promote types, but it doesn't.

### DIFF
--- a/src/DataFramesMeta.jl
+++ b/src/DataFramesMeta.jl
@@ -398,10 +398,10 @@ function transform(g::GroupedDataFrame; kwargs...)
             end
             t = Tables.allocatecolumn(eltype(first), size(result, 1))
             t[idx1[1]:idx2[1]] = first
-            for i in 2:length(g)
+            @inbounds for i in 2:length(g)
                 out = v(g[i])
                 if !(out isa AbstractVector)
-                    throw("Return value must be an `AbstractVector` for all groups or for none of them")
+                    throw(ArgumentError("Return value must be an `AbstractVector` for all groups or for none of them"))
                 elseif length(out) != size(g[i], 1)
                     throw("If a function returns a vector, the result " * 
                           "must have the same length as the groups it operates on")
@@ -417,10 +417,10 @@ function transform(g::GroupedDataFrame; kwargs...)
         else
             t = Tables.allocatecolumn(typeof(first), size(result, 1))
             @views fill!(t[idx1[1]:idx2[1]], first)
-            for i in 2:length(g)
+            @inbounds for i in 2:length(g)
                 out = v(g[i])
                 if out isa AbstractVector
-                    throw("Return value must be an `AbstractVector` for all groups or for none of them")
+                    throw(ArgumentError("Return value must be an `AbstractVector` for all groups or for none of them"))
                 end
                 S = typeof(out)
                 T = eltype(t)
--- a/src/DataFramesMeta.jl
+++ b/src/DataFramesMeta.jl
@@ -393,8 +393,8 @@ function transform(g::GroupedDataFrame; kwargs...)
         first = v(g[1])
         if first isa AbstractVector 
             if length(first) != size(g[1], 1)
-                throw("If a function returns a vector, the result " * 
-                      "must have the same length as the groups it operates on")
+                throw(ArgumentError("If a function returns a vector, the result " * 
+                                    "must have the same length as the groups it operates on"))
             end
             t = Tables.allocatecolumn(eltype(first), size(result, 1))
             t[idx1[1]:idx2[1]] = first
@@ -403,8 +403,8 @@ function transform(g::GroupedDataFrame; kwargs...)
                 if !(out isa AbstractVector)
                     throw(ArgumentError("Return value must be an `AbstractVector` for all groups or for none of them"))
                 elseif length(out) != size(g[i], 1)
-                    throw("If a function returns a vector, the result " * 
-                          "must have the same length as the groups it operates on")
+                    throw(ArgumentError("If a function returns a vector, the result " * 
+                                        "must have the same length as the groups it operates on"))
                 end
                 S = eltype(out)
                 T = eltype(t)
--- a/src/DataFramesMeta.jl
+++ b/src/DataFramesMeta.jl
@@ -385,31 +385,44 @@ function transform(d::Union{AbstractDataFrame, AbstractDict}; kwargs...)
     return result
 end
 
-
 function transform(g::GroupedDataFrame; kwargs...)
     result = DataFrame(g)
     idx2 = cumsum(Int[size(g[i],1) for i in 1:length(g)])
     idx1 = [1; 1 .+ idx2[1:end-1]]
-
     for (k, v) in kwargs
-        output_iterator = (v(ig) for ig in g)
-        if first(output_iterator) isa AbstractVector 
-            first(output_iterator)
-            T = mapreduce(eltype, promote_type, output_iterator)
-            if length(first(output_iterator)) != size(g[1], 1)
+        first = v(g[1])
+        if first isa AbstractVector 
+            if length(first) != size(g[1], 1)
                 throw("If a function returns a vector, the result " * 
                       "must have the same length as the groups it " *
                       "operates on")
             end
+            result[k] = Array{eltype(first)}(undef, size(result, 1))
+            result[idx1[1]:idx2[1], k] = first
+            for i in 2:length(g)
+                out = v(g[i])
+                S = eltype(out)
+                T = eltype(result[k])
+                if !(S <: T)
+                    result[k] = copyto!(similar(result[k]), promote_type(S, T), result[k])
+                end
+                result[idx1[i]:idx2[i], k] = out
+            end
         else 
-            T = mapreduce(typeof, promote_type, output_iterator)
-        end
-        result[k] = Vector{T}(undef, size(result, 1))
-        for (i, output) in enumerate(output_iterator)
-            result[idx1[i]:idx2[i], k] = output
+            result[k] = Array{typeof(first)}(undef, size(result, 1))
+            result[idx1[1]:idx2[1], k] = first
+            for i in 2:length(g)
+                out = v(g[i])
+                S = typeof(out)
+                T = eltype(result[k])
+                if !(S <: T)
+                    result[k] = copyto!(similar(result[k]), promote_type(S, T), result[k])
+                end
+                result[idx1[i]:idx2[i], k] = out
+            end
         end
     end
-    result
+    return result
 end
 
 function transform_helper(x, args...)
--- a/src/DataFramesMeta.jl
+++ b/src/DataFramesMeta.jl
@@ -1,8 +1,7 @@
 module DataFramesMeta
 
 using DataFrames 
-import Tables: allocatecolumn
-
+using Tables
 # Basics:
 export @with, @where, @orderby, @transform, @by, @based_on, @select
 
@@ -411,7 +410,7 @@ function transform(g::GroupedDataFrame; kwargs...)
                 T = eltype(t)
                 if !(S <: T || promote_type(S, T) <: T)
                     t = copyto!(Tables.allocatecolumn(promote_type(S, T), size(result, 1)), 
-                        t[1:idx2[i-1]])
+                                t[1:idx2[i-1]])
                 end
                 t[idx1[i]:idx2[i]] = out
             end
@@ -424,7 +423,7 @@ function transform(g::GroupedDataFrame; kwargs...)
                 T = eltype(t)
                 if !(S <: T || promote_type(S, T) <: T)
                     t = copyto!(Tables.allocatecolumn(promote_type(S, T), size(result, 1)), 
-                        t[1:idx2[i-1]])
+                                t[1:idx2[i-1]])
                 end
                 t[idx1[i]:idx2[i]] .= out
             end
--- a/src/DataFramesMeta.jl
+++ b/src/DataFramesMeta.jl
@@ -402,6 +402,11 @@ function transform(g::GroupedDataFrame; kwargs...)
             t[idx1[1]:idx2[1]] = first
             for i in 2:length(g)
                 out = v(g[i])
+                if length(out) != size(g[i], 1)
+                    throw("If a function returns a vector, the result " * 
+                      "must have the same length as the groups it " *
+                      "operates on")
+                end
                 S = eltype(out)
                 T = eltype(t)
                 if !(S <: T || promote_type(S, T) <: T)
--- a/src/DataFramesMeta.jl
+++ b/src/DataFramesMeta.jl
@@ -387,54 +387,60 @@ end
 
 function transform(g::GroupedDataFrame; kwargs...)
     result = DataFrame(g)
-    idx2 = cumsum(Int[size(g[i],1) for i in 1:length(g)])
-    idx1 = [1; 1 .+ idx2[1:end-1]]
+    ends = cumsum(Int[size(g[i],1) for i in 1:length(g)])
+    starts = [1; 1 .+ ends[1:end-1]]
     for (k, v) in kwargs
         first = v(g[1])
         if first isa AbstractVector 
-            if length(first) != size(g[1], 1)
-                throw(ArgumentError("If a function returns a vector, the result " * 
-                                    "must have the same length as the groups it operates on"))
-            end
             t = Tables.allocatecolumn(eltype(first), size(result, 1))
-            t[idx1[1]:idx2[1]] = first
-            @inbounds for i in 2:length(g)
-                out = v(g[i])
-                if !(out isa AbstractVector)
-                    throw(ArgumentError("Return value must be an `AbstractVector` for all groups or for none of them"))
-                elseif length(out) != size(g[i], 1)
-                    throw(ArgumentError("If a function returns a vector, the result " * 
-                                        "must have the same length as the groups it operates on"))
-                end
-                S = eltype(out)
-                T = eltype(t)
-                if !(S <: T || promote_type(S, T) <: T)
-                    t = copyto!(Tables.allocatecolumn(promote_type(S, T), size(result, 1)), 
-                                1, t, 1, idx2[i-1])
-                end
-                t[idx1[i]:idx2[i]] = out
-            end
-        else
+            t = _transform!(t, 1, first, first, g, v, starts, ends)
+        else 
             t = Tables.allocatecolumn(typeof(first), size(result, 1))
-            @views fill!(t[idx1[1]:idx2[1]], first)
-            @inbounds for i in 2:length(g)
-                out = v(g[i])
-                if out isa AbstractVector
-                    throw(ArgumentError("Return value must be an `AbstractVector` for all groups or for none of them"))
-                end
-                S = typeof(out)
-                T = eltype(t)
-                if !(S <: T || promote_type(S, T) <: T)
-                    t = copyto!(Tables.allocatecolumn(promote_type(S, T), size(result, 1)), 
-                                1, t, 1, idx2[i-1])
-                end
-                @views fill!(t[idx1[i]:idx2[i]], out)
-            end
+            t = _transform!(t, 1, first, first, g, v, starts, ends)
         end
         result[k] = t
     end
     return result
 end
+
+function _transform!(t::AbstractVector, i::Int, previous::AbstractVector, current::AbstractVector, g::GroupedDataFrame, v::Function, starts::Vector, ends::Vector)
+    if !(current isa AbstractVector)
+        throw(ArgumentError("Return value must be an `AbstractVector` for all groups or for none of them"))
+    elseif length(current) != size(g[i], 1)
+        throw(ArgumentError("If a function returns a vector, the result " * 
+                            "must have the same length as the groups it operates on"))
+    end
+    S = eltype(current)
+    T = eltype(previous)
+    if !(S <: T || promote_type(S, T) <: T)
+        t = copyto!(Tables.allocatecolumn(promote_type(S, T), length(t)), 
+                    1, t, 1, ends[i-1])
+    end
+    t[starts[i]:ends[i]] = current
+    if i != length(g) 
+        return _transform!(t, i + 1, current, v(g[i+1]), g, v, starts, ends)
+    else 
+        return t
+    end
+end 
+
+function _transform!(t::AbstractVector, i::Int, previous::Any, current::Any, g::GroupedDataFrame, v::Function, starts::Vector, ends::Vector)
+    if (current isa AbstractVector)
+        throw(ArgumentError("Return value must be an `AbstractVector` for all groups or for none of them"))
+    end
+    S = typeof(current)
+    T = typeof(previous)
+    if !(S <: T || promote_type(S, T) <: T)
+        t = copyto!(Tables.allocatecolumn(promote_type(S, T), length(t)), 
+                    1, t, 1, ends[i-1])
+    end
+    @views fill!(t[starts[i]:ends[i]], current)
+    if i != length(g) 
+        return _transform!(t, i + 1, current, v(g[i+1]), g, v, starts, ends)
+    else 
+        return t
+    end
+end 
 
 function transform_helper(x, args...)
     quote
--- a/src/DataFramesMeta.jl
+++ b/src/DataFramesMeta.jl
@@ -400,13 +400,11 @@ function transform(g::GroupedDataFrame; kwargs...)
             t[idx1[1]:idx2[1]] = first
             for i in 2:length(g)
                 out = v(g[i])
-                if length(out) != size(g[i], 1)
+                if !(out isa AbstractVector)
+                    throw("Return value must be an `AbstractVector` for all groups or for none of them")
+                elseif length(out) != size(g[i], 1)
                     throw("If a function returns a vector, the result " * 
                           "must have the same length as the groups it operates on")
-                end
-                if !(out isa AbstractVector)
-                    throw("return value must not change its kind (single " *
-                          "value, named tuple, vector or data frame) across groups")
                 end
                 S = eltype(out)
                 T = eltype(t)
@@ -416,14 +414,13 @@ function transform(g::GroupedDataFrame; kwargs...)
                 end
                 t[idx1[i]:idx2[i]] = out
             end
-        else 
+        else
             t = Tables.allocatecolumn(typeof(first), size(result, 1))
-            t[idx1[1]:idx2[1]] .= first
+            t[idx1[1]:idx2[1]] = fill(first, size(g[1],1))
             for i in 2:length(g)
                 out = v(g[i])
                 if out isa AbstractVector
-                    throw("return value must not change its kind (single " *
-                          "value, named tuple, vector or data frame) across groups")
+                    throw("Return value must be an `AbstractVector` for all groups or for none of them")
                 end
                 S = typeof(out)
                 T = eltype(t)
@@ -431,7 +428,7 @@ function transform(g::GroupedDataFrame; kwargs...)
                     t = copyto!(Tables.allocatecolumn(promote_type(S, T), size(result, 1)), 
                                 1, t, 1, idx2[i-1])
                 end
-                t[idx1[i]:idx2[i]] .= out
+                t[idx1[i]:idx2[i]] = fill(out, size(g[i],1))
             end
         end
         result[k] = t
--- a/src/DataFramesMeta.jl
+++ b/src/DataFramesMeta.jl
@@ -405,9 +405,29 @@ end
 
 function _transform!(t::AbstractVector, first::AbstractVector, start::Int, 
                      g::GroupedDataFrame, v::Function, starts::Vector, ends::Vector)
+    @inline function fill_column_vec!(t::AbstractVector, out, startpoint::Int, endpoint::Int, len::Int)
+        if !(out isa AbstractVector)
+            throw(ArgumentError("Return value must be an `AbstractVector` for all groups or" *
+                                "for none of them"))
+        elseif length(out) != len
+            throw(ArgumentError("If a function returns a vector, the result " * 
+                                "must have the same length as the groups it operates on"))    
+        end     
+        elout = eltype(out)
+        T = eltype(t)
+        newtype = promote_type(elout, T)
+        if elout <: T || newtype <: T
+           t[startpoint:endpoint] = out
+            return nothing 
+        else 
+            return newtype
+        end
+        return nothing
+    end 
+
     # handle the first case 
-    newtype = fill_column_vec!(t, first, starts[start], ends[start], size(g[start], 1))
-    @assert newtype === nothing
+    newtype_first = fill_column_vec!(t, first, starts[start], ends[start], size(g[start], 1))
+    #@assert newtype_first === nothing
     @inbounds for i in (start+1):length(g)
         out = v(g[i])
         newtype = fill_column_vec!(t, out, starts[i], ends[i], size(g[i], 1))
@@ -422,9 +442,24 @@ end
 
 function _transform!(t::AbstractVector, first::Any, start::Int, 
                      g::GroupedDataFrame, v::Function, starts::Vector, ends::Vector)
+    @inline function fill_column_any!(t::AbstractVector, out, startpoint::Int, endpoint::Int)
+        if out isa AbstractVector
+            throw(ArgumentError("Return value must be an `AbstractVector` for all groups or" *
+                                 "for none of them"))
+        end     
+        typout = typeof(out)
+        T = eltype(t)
+        newtype = promote_type(typout, T)
+        if typout <: T || newtype <: T
+            t[startpoint:endpoint] .= Ref(out)
+            return nothing
+        else 
+            return newtype
+        end
+    end 
     # handle the first case 
-    newtype = fill_column_any!(t, first, starts[start], ends[start])
-    @assert newtype === nothing
+    newtype_first = fill_column_any!(t, first, starts[start], ends[start])
+    #@assert newtype_first === nothing
     @inbounds for i in (start+1):length(g)
         out = v(g[i])
         newtype = fill_column_any!(t, out, starts[i], ends[i])
@@ -435,42 +470,8 @@ function _transform!(t::AbstractVector, first::Any, start::Int,
          end
     end
     return t
+
 end
-
-function fill_column_vec!(t::AbstractVector, out, startpoint::Int, endpoint::Int, len::Int)
-    if !(out isa AbstractVector)
-        throw(ArgumentError("Return value must be an `AbstractVector` for all groups or" *
-                            "for none of them"))
-    elseif length(out) != len
-        throw(ArgumentError("If a function returns a vector, the result " * 
-                            "must have the same length as the groups it operates on"))    
-    end     
-    elout = eltype(out)
-    T = eltype(t)
-    newtype = promote_type(elout, T)
-    if elout <: T || newtype <: T
-        t[startpoint:endpoint] = out
-        return nothing 
-    else 
-        return newtype
-    end
-end 
-
-function fill_column_any!(t::AbstractVector, out, startpoint::Int, endpoint::Int)
-    if out isa AbstractVector
-        throw(ArgumentError("Return value must be an `AbstractVector` for all groups or" *
-                             "for none of them"))
-    end     
-    typout = typeof(out)
-    T = eltype(t)
-    newtype = promote_type(typout, T)
-    if typout <: T || newtype <: T
-        t[startpoint:endpoint] .= Ref(out)
-        return nothing
-    else 
-        return newtype
-    end
-end 
 
 function transform_helper(x, args...)
     quote
--- a/src/DataFramesMeta.jl
+++ b/src/DataFramesMeta.jl
@@ -407,7 +407,7 @@ end
 
 function _transform!(t::AbstractVector, first::AbstractVector, start::Int, 
                      g::GroupedDataFrame, v::Function, starts::Vector, ends::Vector)
-    @inline function fill_column_vec!(t::AbstractVector, out, startpoint::Int, endpoint::Int, 
+    @inline function fill_column!(t::AbstractVector, out, startpoint::Int, endpoint::Int, 
                                       len::Int)
         if !(out isa AbstractVector)
             throw(ArgumentError("Return value must be an `AbstractVector` for all groups or" *
@@ -416,10 +416,9 @@ function _transform!(t::AbstractVector, first::AbstractVector, start::Int,
             throw(ArgumentError("If a function returns a vector, the result " * 
                                 "must have the same length as the groups it operates on"))    
         end     
-        elout = eltype(out)
+        eltypout = eltype(out)
         T = eltype(t)
-        newtype = promote_type(elout, T)
-        if elout <: T || newtype <: T
+        if eltypout <: T || (newtype = promote_type(eltypout, T)) <: T
            t[startpoint:endpoint] = out
             return nothing 
         else 
@@ -429,11 +428,11 @@ function _transform!(t::AbstractVector, first::AbstractVector, start::Int,
     end 
 
     # handle the first case 
-    newtype_first = fill_column_vec!(t, first, starts[start], ends[start], size(g[start], 1))
-    #@assert newtype_first === nothing
+    newtype_first = fill_column!(t, first, starts[start], ends[start], size(g[start], 1))
+    @assert newtype_first === nothing
     @inbounds for i in (start+1):length(g)
         out = v(g[i])
-        newtype = fill_column_vec!(t, out, starts[i], ends[i], size(g[i], 1))
+        newtype = fill_column!(t, out, starts[i], ends[i], size(g[i], 1))
         if newtype !== nothing
              t = copyto!(Tables.allocatecolumn(newtype, length(t)), 
                          1, t, 1, ends[i-1])
@@ -445,15 +444,14 @@ end
 
 function _transform!(t::AbstractVector, first::Any, start::Int, 
                      g::GroupedDataFrame, v::Function, starts::Vector, ends::Vector)
-    @inline function fill_column_any!(t::AbstractVector, out, startpoint::Int, endpoint::Int)
+    @inline function fill_column!(t::AbstractVector, out, startpoint::Int, endpoint::Int)
         if out isa AbstractVector
             throw(ArgumentError("Return value must be an `AbstractVector` for all groups or" *
                                  "for none of them"))
         end     
         typout = typeof(out)
         T = eltype(t)
-        newtype = promote_type(typout, T)
-        if typout <: T || newtype <: T
+        if typout <: T || (newtype = promote_type(typout, T)) <: T
             t[startpoint:endpoint] .= Ref(out)
             return nothing
         else 
@@ -461,11 +459,11 @@ function _transform!(t::AbstractVector, first::Any, start::Int,
         end
     end 
     # handle the first case 
-    newtype_first = fill_column_any!(t, first, starts[start], ends[start])
-    #@assert newtype_first === nothing
+    newtype_first = fill_column!(t, first, starts[start], ends[start])
+    @assert newtype_first === nothing
     @inbounds for i in (start+1):length(g)
         out = v(g[i])
-        newtype = fill_column_any!(t, out, starts[i], ends[i])
+        newtype = fill_column!(t, out, starts[i], ends[i])
         if newtype !== nothing
              t = copyto!(Tables.allocatecolumn(newtype, length(t)), 
                          1, t, 1, ends[i-1])
@@ -473,7 +471,6 @@ function _transform!(t::AbstractVector, first::Any, start::Int,
          end
     end
     return t
-
 end
 
 function transform_helper(x, args...)
--- a/src/DataFramesMeta.jl
+++ b/src/DataFramesMeta.jl
@@ -416,7 +416,7 @@ function transform(g::GroupedDataFrame; kwargs...)
             end
         else
             t = Tables.allocatecolumn(typeof(first), size(result, 1))
-            t[idx1[1]:idx2[1]] = fill(first, size(g[1],1))
+            @views fill!(t[idx1[1]:idx2[1]], first)
             for i in 2:length(g)
                 out = v(g[i])
                 if out isa AbstractVector
@@ -428,7 +428,7 @@ function transform(g::GroupedDataFrame; kwargs...)
                     t = copyto!(Tables.allocatecolumn(promote_type(S, T), size(result, 1)), 
                                 1, t, 1, idx2[i-1])
                 end
-                t[idx1[i]:idx2[i]] = fill(out, size(g[i],1))
+                @views fill!(t[idx1[i]:idx2[i]], out)
             end
         end
         result[k] = t
--- a/src/DataFramesMeta.jl
+++ b/src/DataFramesMeta.jl
@@ -1,7 +1,7 @@
 module DataFramesMeta
 
 using DataFrames 
-using Tables
+using Tables: allocatecolumn
 # Basics:
 export @with, @where, @orderby, @transform, @by, @based_on, @select
 
@@ -410,7 +410,7 @@ function transform(g::GroupedDataFrame; kwargs...)
                 T = eltype(t)
                 if !(S <: T || promote_type(S, T) <: T)
                     t = copyto!(Tables.allocatecolumn(promote_type(S, T), size(result, 1)), 
-                                t[1:idx2[i-1]])
+                                1, t, 1, idx2[i-1])
                 end
                 t[idx1[i]:idx2[i]] = out
             end
@@ -423,7 +423,7 @@ function transform(g::GroupedDataFrame; kwargs...)
                 T = eltype(t)
                 if !(S <: T || promote_type(S, T) <: T)
                     t = copyto!(Tables.allocatecolumn(promote_type(S, T), size(result, 1)), 
-                                t[1:idx2[i-1]])
+                                1, t, 1, idx2[i-1])
                 end
                 t[idx1[i]:idx2[i]] .= out
             end
--- a/src/DataFramesMeta.jl
+++ b/src/DataFramesMeta.jl
@@ -405,6 +405,10 @@ function transform(g::GroupedDataFrame; kwargs...)
                     throw("If a function returns a vector, the result " * 
                           "must have the same length as the groups it operates on")
                 end
+                if !(out isa AbstractVector)
+                    throw("return value must not change its kind (single " *
+                          "value, named tuple, vector or data frame) across groups")
+                end
                 S = eltype(out)
                 T = eltype(t)
                 if !(S <: T || promote_type(S, T) <: T)
@@ -418,6 +422,10 @@ function transform(g::GroupedDataFrame; kwargs...)
             t[idx1[1]:idx2[1]] .= first
             for i in 2:length(g)
                 out = v(g[i])
+                if out isa AbstractVector
+                    throw("return value must not change its kind (single " *
+                          "value, named tuple, vector or data frame) across groups")
+                end
                 S = typeof(out)
                 T = eltype(t)
                 if !(S <: T || promote_type(S, T) <: T)
--- a/src/DataFramesMeta.jl
+++ b/src/DataFramesMeta.jl
@@ -394,9 +394,11 @@ function transform(g::GroupedDataFrame; kwargs...)
     for (k, v) in kwargs
         first = v(g[1])
         if first isa AbstractVector
-            t = _transform!(Tables.allocatecolumn(eltype(first), size(result, 1)), first, 1, g, v, starts, ends)
+            t = _transform!(Tables.allocatecolumn(eltype(first), size(result, 1)), 
+                            first, 1, g, v, starts, ends)
         else 
-            t = _transform!(Tables.allocatecolumn(typeof(first), size(result, 1)), first, 1, g, v, starts, ends)
+            t = _transform!(Tables.allocatecolumn(typeof(first), size(result, 1)), 
+                            first, 1, g, v, starts, ends)
         end
         result[k] = t
     end
@@ -405,7 +407,8 @@ end
 
 function _transform!(t::AbstractVector, first::AbstractVector, start::Int, 
                      g::GroupedDataFrame, v::Function, starts::Vector, ends::Vector)
-    @inline function fill_column_vec!(t::AbstractVector, out, startpoint::Int, endpoint::Int, len::Int)
+    @inline function fill_column_vec!(t::AbstractVector, out, startpoint::Int, endpoint::Int, 
+                                      len::Int)
         if !(out isa AbstractVector)
             throw(ArgumentError("Return value must be an `AbstractVector` for all groups or" *
                                 "for none of them"))
--- a/src/DataFramesMeta.jl
+++ b/src/DataFramesMeta.jl
@@ -392,54 +392,79 @@ function transform(g::GroupedDataFrame; kwargs...)
     lengths = [ends[i] - starts[i] + 1 for i in 1:length(starts)]
     for (k, v) in kwargs
         first = v(g[1])
-        if first isa AbstractVector 
-            t = Tables.allocatecolumn(eltype(first), size(result, 1))
-            t[starts[1]:ends[1]] = first
+        if first isa AbstractVector
+            t = _transform!(Tables.allocatecolumn(eltype(first), size(result, 1)), first, 1, g, v, starts, ends)
         else 
-            t = Tables.allocatecolumn(typeof(first), size(result, 1))
-            @views fill!(t[starts[1]:ends[1]], first)
-        end
-        @inbounds for i in 2:length(g)
-            t = _transform!(t, i, v(g[i]), starts, ends, lengths)
+            t = _transform!(Tables.allocatecolumn(typeof(first), size(result, 1)), first, 1, g, v, starts, ends)
         end
         result[k] = t
     end
     return result
 end
 
-function _transform!(t::AbstractVector, i::Int, out::AbstractVector, starts::Vector, ends::Vector, lengths::Vector)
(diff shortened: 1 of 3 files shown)


pdeffebach:
@bkamins 

This is likely the function to use from DataFrames

https://github.com/JuliaData/DataFrames.jl/blob/56a9b720b9bc61d778fd900c15d6b85309517136/src/abstractdataframe/abstractdataframe.jl#L859

```
@generated function promote_col_type(cols::AbstractVector...)
    T = mapreduce(x -> Missings.T(eltype(x)), promote_type, cols)
    if CategoricalArrays.iscatvalue(T)
        T = CategoricalArrays.leveltype(T)
    end
    if any(col -> eltype(col) >: Missing, cols)
        if any(col -> col <: AbstractCategoricalArray, cols)
            return :(CategoricalVector{Union{$T, Missing}})
        else
            return :(Vector{Union{$T, Missing}})
        end
    else
        if any(col -> col <: AbstractCategoricalArray, cols)
            return :(CategoricalVector{$T})
        else
            return :(Vector{$T})
        end
    end
end
```

The way it deals with `missing`s is weird, and maybe is a holdover from before the new small union optimizations. So maybe there is something to be done about this function first while we are at it for this PR.

bkamins:
As far as I understand its main difference from `promote_Type` is that it tries to preserve categorical type if you mix categorical and non-categorical vector.

pdeffebach:
Something like 

```
d = v(gi) for i in g # simplified cause we have a spread_scalar function somewhere
T = DataFrames.promote_col_type(d...)
result[k] = T(reduce(vcat, d))
```

I *think* that since `DataFrames.promote_col_type` is a `@generated` function, the type-finding step of this is actually quite efficient, with the loop happening at compile time.

pdeffebach:
This is proving to be a pain. 

1. `fill` doesn't work with CategoricalArrays quite like it should. You can't get the `pool` to be the same for the vector as it is for the individual element. Though I will continue looking at constructors code
2. It's hard to get the DataFrames function above working with a generator or other lazy object. 

I almost want to do PRs to `CategoricalArrays` so that `vcat(v(ig) for ig in g)` "just works" without worrying about categorical arrays. But I am sure that Categorical Arrays is a lot more complicated otherwise `vcat` would already just work.

bkamins:
@nalimilan do you have a second to have look at it? You probably have the answer ready having designed and implemented all this 😄. Thanks.

pdeffebach:
For clarity, this is the super contrived example I am imagining. 

```
df = DataFrame(id = [1, 2, 3, 1, 2, 3], year = [95, 95, 95, 96, 96, 96], x = rand(6), some_personal_variable = CategoricalArray([1, 2, missing, 1, missing, missing]))

6×4 DataFrames.DataFrame
│ Row │ id │ year │ x         │ some_personal_variable │
├─────┼────┼──────┼───────────┼────────────────────────┤
│ 1   │ 1  │ 95   │ 0.232321  │ 1                      │
│ 2   │ 2  │ 95   │ 0.0617226 │ 2                      │
│ 3   │ 3  │ 95   │ 0.970737  │ missing                │
│ 4   │ 1  │ 96   │ 0.0731739 │ 1                      │
│ 5   │ 2  │ 96   │ 0.555002  │ missing                │
│ 6   │ 3  │ 96   │ 0.963342  │ missing                │
```

`some_personal_variable` is time-invariant, like birth location. But we collected it inconsistently. So individual `2` has it for year `96` but not `96`. We want to "spread" that value across other years, with a function like `collect(skipmissing(:some_personal_variable))[1]` on a grouped dataframe. We do this in a transform because we are too lazy to to a `by` (stata `collapse`) operation and then perform a `m:1` merge on the collapse data. 

We want a command that 

1. Knows that my function returns a scalar and spreads the values into a vector accordingly
2. Promotes types in the appropriate way

pdeffebach:
Just got a chance to work on this. 

I think all my worrying about types with categorical arrays is overblown! As far as I can tell `reduce(vcat,...)` seems to do type promotion the right way, in the example I described above. 

I think this is ready for another review.

src/DataFramesMeta.jl
```diff
@@ -387,17 +386,29 @@ function transform(d::Union{AbstractDataFrame, AbstractDict}; kwargs...)
...
     idx2 = cumsum(Int[size(g[i],1) for i in 1:length(g)])
     idx1 = [1; 1 + idx2[1:end-1]]
+
+    function spread_scalar(x::Vector, length::Int)
+        return x 
+    end
+
+    function spread_scalar(x::CategoricalArrays.CategoricalValue, length::Int)
+        vec = CategoricalArray(fill(x, length))
+        levels!(vec, CategoricalArrays.index(CategoricalArrays.pool(x)))
```
bkamins: why this line is needed? I understand that without it `vec` has only one level? Yes?

src/DataFramesMeta.jl
```diff
@@ -387,17 +386,29 @@ function transform(d::Union{AbstractDataFrame, AbstractDict}; kwargs...)
...
 end
 
+
 function transform(g::GroupedDataFrame; kwargs...)
     result = DataFrame(g)
     idx2 = cumsum(Int[size(g[i],1) for i in 1:length(g)])
     idx1 = [1; 1 + idx2[1:end-1]]
+
+    function spread_scalar(x::Vector, length::Int)
+        return x
```
bkamins: what if `x` does not have length `length` (also maybe not use `length` as name as it clashes with `length` function).

src/DataFramesMeta.jl
```diff
@@ -387,17 +386,29 @@ function transform(d::Union{AbstractDataFrame, AbstractDict}; kwargs...)
     return result
 end
 
+
 function transform(g::GroupedDataFrame; kwargs...)
     result = DataFrame(g)
     idx2 = cumsum(Int[size(g[i],1) for i in 1:length(g)])
     idx1 = [1; 1 + idx2[1:end-1]]
```
bkamins: I think `idx1` and `idx2` are not needed?

src/DataFramesMeta.jl
pdeffebach: If we have a `CategoricalArray[1,2,1,2,1]` , I was thinking that if you take the first element of each group, filling in the rest of the group with 1 value, we still want to have the new vector have the same levels as the original one. 

Without that line you also get this weird case where the `pool` of the array and the `pool` of the elements are different.

src/DataFramesMeta.jl
pdeffebach: Good point on changing the `length` name. 

`length` is the number of observations in each group. So we are telling the function how many times to replicate the result of a `vector -> scalar` function.

src/DataFramesMeta.jl
pdeffebach: true, sorry.

src/DataFramesMeta.jl
bkamins: This is what I thought. Just wanted to be sure.

src/DataFramesMeta.jl
bkamins: But this is why I am asking - in line 395 you assume that `v` returned a vector not a scalar and you keep it unchanged. What if the length of this vector is not correct?

pdeffebach: You just get a 

```
ERROR: ArgumentError: New columns must have the same length as old columns
```

Perhaps we should add an error about only doing only `vector -> vector (right length)` or `vector -> scalar`.

bkamins: You will not when by accident the total length of vectors is OK, but the vectors themselves are not of correct length (e.g. you have two groups of length 3, 6 in total, and vectors are of lengths 2 and 4). I think we should strictly check that if a `Vector` is returned it has the length equal to the length of the group.

pdeffebach: Okay, so 
```
length(x) == obs_in_group ? x : error...
```

bkamins: Looks good to me 👍.

pdeffebach:
Sorry maybe I'm wrong but what am I doing? We should just call `@based_on`, because it seems to do the exact thing we want.

pdeffebach:
Apologize. the `@based_on` docstring had me confused. If you have two operations, one going to scalar and the other to vector, the scalar result is spread out, which made it seem like transform. But also `@based_on` drops columns. 

Very excited to get `describe` working with a grouped dataframe, because the `dplyr` strategy of performing grouped operations just to get summary statistics is annoying.

src/DataFramesMeta.jl
```diff
@@ -387,17 +386,26 @@ function transform(d::Union{AbstractDataFrame, AbstractDict}; kwargs...)
...
 end
 
+
 function transform(g::GroupedDataFrame; kwargs...)
     result = DataFrame(g)
-    idx2 = cumsum(Int[size(g[i],1) for i in 1:length(g)])
-    idx1 = [1; 1 + idx2[1:end-1]]
+
+    function spread_scalar(x::Vector, obs_in_group::Int)
+        length(x) == obs_in_group ? x : error("Functions must return either a vector the same length as each groupor a scalar")
```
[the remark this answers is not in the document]
pdeffebach: fixed!

src/DataFramesMeta.jl
```diff
@@ -387,18 +386,33 @@ function transform(d::Union{AbstractDataFrame, AbstractDict}; kwargs...)
...
+        levels!(vec, CategoricalArrays.index(CategoricalArrays.pool(x)))
+    end
+
+    function spread_scalar(x::Any, obs_in_group::Int)
+        fill(x, obs_in_group)
+    end
+
+    for (k, v) in kwargs
+        spreading_helper = x -> spread_scalar(v(x), size(x, 1))
+        result[k] = reduce(vcat, spreading_helper(ig) for ig in g)
```
bkamins: why it is better than simply:
```
result[k] = reduce(vcat, spread_scalar(v(ig), size(ig, 1)) for ig in g)
```

bkamins:
Excellent. I have left small comments. Can you copy-paste here the result of `describe` (or maybe even add it to the tests - so that people see that it can be used this way)?

pdeffebach:
Just added those things. 

w.r.t. describe, I have to make some PRs to DataFrames to get it working. I was just saying that this `dpyr`-eque workflow often uses grouped operations to get a glance summary statistics, and it doesn't have to be that way.

test/grouping.jl
```diff
@@ -21,4 +21,8 @@ d = DataFrame(n = 1:20, x = [3, 3, 3, 3, 1, 1, 1, 2, 1, 1, 2, 1, 1, 2, 2, 2, 3,
 g = groupby(d, :x, sort=true)
 @test @based_on(g, nsum = sum(:n))[:nsum] == [99, 84, 27]
 
+d = DataFrame(a = [1,1,1,2,2], b = [1,2,3,missing,missing])
```
nalimilan: If you expect a particular behavior for `CategoricalArray`, it should be tested (and a dependency on it be added only for testing).

src/DataFramesMeta.jl
```diff
@@ -378,7 +378,6 @@ end
 ## transform & @transform
 ##
 ##############################################################################
-
```
nalimilan: No need to remove this line, nor to add one below?

src/DataFramesMeta.jl
```diff
@@ -387,18 +386,32 @@ function transform(d::Union{AbstractDataFrame, AbstractDict}; kwargs...)
...
+        vec = CategoricalArray(fill(x, obs_in_group))
+        levels!(vec, CategoricalArrays.index(CategoricalArrays.pool(x)))
+    end
+
+    function spread_scalar(x::Any, obs_in_group::Int)
+        fill(x, obs_in_group)
+    end
+
+    for (k, v) in kwargs
+        result[k] = reduce(vcat, spread_scalar(v(ig), size(ig, 1)) for ig in g)
```
nalimilan: In theory using a generator as you do is better, but in practice I'm afraid that's going to be slow because the optimized `reduce(vcat, ...)` method from https://github.com/JuliaLang/julia/pull/27188 doesn't exist for generators. For a large number of groups, allocating a new copy for each new entry is going to kill performance.

I've filed https://github.com/JuliaLang/julia/issues/28691 to track this. In the meantime, better keep the existing approach which creates columns and fills them manually, with a reference to that issue.

pdeffebach: > In the meantime, better keep the existing approach which creates columns and fills them manually, with a reference to that issue.

Thanks. The original problem is that if the first group returns `[1,2,3]` and the second group returns `[missing, missing, missing]`, there is an error because the original vector allocated would be of type `Int`. 

Do you want me to re-work this PR so that `promote_type` works better and we allocate the right vector the first time? 

Or should we put this PR on hold and wait for `vcat` to work better?

nalimilan: OK. Then the solution is to do something like `mapreduce(eltype, promote_type, A)` to choose the best element type, as in https://github.com/JuliaLang/julia/pull/27188.

I don't think we can wait for `vcat` to improve, since it's not clear when it will happen (maybe it needs to wait for 2.0 to avoid breakage). See https://github.com/JuliaLang/julia/issues/18472.

src/DataFramesMeta.jl
```diff
@@ -387,18 +386,32 @@ function transform(d::Union{AbstractDataFrame, AbstractDict}; kwargs...)
...
+        if length(x) == obs_in_group 
+            return x 
+        else
+            throw(DimensionError("If a function returns a vector, the result " * 
+                                 "must have the same length as the groups it " *
+                                 "operates on"))
         end
     end
+
+    function spread_scalar(x::CategoricalArrays.CategoricalValue, obs_in_group::Int)
```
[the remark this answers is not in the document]
nalimilan: Actually, `(x for x in 1:obs_in_group)` isn't indexable, so my suggestion won't work. I guess the best solution is to have an `if x isa AbstractArray` branch in the `for` loop below, rather than dispatching on methods.

pdeffebach:
Based on the most recent comments, I think I might rebase and start anew. To recap: the goal was to fix three problems relating to `@transform` on a grouped dataframe, which is a grouped operation, but without the collapse. 

1. The original code called `eltype` on the returned value, meaning that functions returning strings got a `char` array. So when it tried to fill in that array with the output, an error would be thrown.
2. The type of the new vector is determined by the first group only, so a function that returned all missing for a group, for example, would throw an error. 
3. Categorical arrays aren't preserved since we allocate a new array. 

I thought all of these problems could be avoided, and the code would be simpler, by 

1. Making `vector -> scalar` functions return a vector via my `spread_by` operation. 
2. Using `vcat`'s type promotion rules to avoid allocating a return vector of a specific type. This seemed especially worthwhile given `CategoricalArray`'s odd type promotion rules. 

My current proposal for this is to keep the current set up (allocating a vector of a certain type, then filling it in as needed), but with two changes. 
1. Better type promotion function using `reduce` as Milan suggested. 
2. Use dispatch to decide if something is 
    1. A string
    2. A scalar in general
  And handle those cases separately. 

We can worry about the behavior of `CategoricaArray`s later, and give more thought to what behavior we want upstream with regars to type promotion.

pdeffebach:
I successfully rebased and squashed the commits.. I think. The PR now looks more similar to what the existing code was. 

For each key-value pair in `kwargs` I create a generator `(v(gi) for gi in g)` and then use that to use `mapreduce` and find the right type promotion. 

I don't actually use dispatch on this. Rather, I just use plain old `if...else`. Given that none of this is type stable, I don't think its an issue. And I don't really know how I would dispatch the output from the first element of the generator.

nalimilan:
Sorry if I missed this in my previous comments, but there's a problem with that approach: you compute transformations twice, which is going to be much slower. The solution which is generally used these days (by `map` in Base but also in e.g. Tables.jl and IndexedTables.jl) is to allocate a vector based on the eltype of the first result, and then check for each group whether its eltype is a subtype of the vector's eltype. If not, allocate a new vector (choosing the eltype using `promote_type`), and copy the old data to it. In general that's quite efficient since for a `Union{T,Missing}` vector only one (partial) copy will be made (and generally it will happen quite soon so not many elements have to be copied).

I'm currently preparing a PR using this strategy for `by` in DataFrames. The code is a bit more involved since the user can return a DataFrame (and not just one vector per transformation), and because I'm calling separate functions which operate on tuples of resulting columns to ensure some type stability (but not full type stability unfortunately). It would be interesting to try this approach here, but even without the type stability tricks you can take inspiration from it to reallocate vectors when needed. See in particular [these lines](https://github.com/JuliaData/DataFrames.jl/blob/4ff408a75711ff13062a165de2d4fa7dc797a276/src/groupeddataframe/grouping.jl#L221-L227).

nalimilan:
Actually, there's a very interesting challenge here which could give incredible performance with some transformations to how the code works. If instead of passing `g[i]` to the anonymous function `v` we passed it directly the columns it operates on, and if we called that from a helper function dedicated to each transformation taking the said columns and applying it to all groups, we would have fully type-stable code and `v` would probably be inlined. In short, that's the best possible code one could write, even by hand. That would particularly increase performance when the number of groups is large.

Now, the `v` method taking a tuple of functions already exists, but we need to get the names of the columns it expects. Apparently this can be obtained by extracting the names of arguments via something like `Symbol.(getindex.(Base.arg_decl_parts(methods(v).ms[2])[2][2:end], 1))`. So it looks like it wouldn't be that hard.

pdeffebach:
Thanks for the feedback. It looks like I had two misonceptions

1. I thought generators were faster. I thought it was a way to have Julia strore stuff but not allocate a contiguous array for it. 
2. I assumed array re-typing is expensive. I'm glad to hear its not. 

I will try the approach you showed me. 

On the other hand, this kind of operation could arguably be in `by` or something similar. So perhaps eventually DataFramesMeta could just call a `by` operation.

nalimilan:
>     1. I thought generators were faster. I thought it was a way to have Julia strore stuff but not allocate a contiguous array for it.

They are fast in the sense that they do not allocate a vector with all the results. But the downside is that they have to reevaluate the call each time you access them.

>     2. I assumed array re-typing is expensive. I'm glad to hear its not.

Re-typing is expensive since it requires copying all already processed elements. But that's less expensive than other solutions. Hopefully at some point we'll have a way of converting an `Array{T}` to an `Array{Union{T,Missing}}` without making a copy.

> I will try the approach you showed me.

Cool!

> On the other hand, this kind of operation could arguably be in `by` or something similar. So perhaps eventually DataFramesMeta could just call a `by` operation.

Actually DataFramesMeta can do things more efficiently than DataFrames since thanks to macros it knows which columns are involved in a computation and can specialize on them. OTC DataFrames's `by` passes a full data frame to the user-provided function, which creates a type instability (https://github.com/JuliaData/DataFrames.jl/issues/1256). But you're right that `@by` could probably take advantage of the same approach as `@transform(g::GroupedDataFrame, ...)`. AFAICT these are really the same operation.

pdeffebach:
> AFAICT these are really the same operation.

I meant that `@transform(g::GroupedDataFrame, ..)` spreads the results so that there is no `collapse` at any point, while all the `by` operations in DataFrames always collapse so you get a dataframe where each observation is a group. In Base DataFrames I think  you would need a `by` and a `join`. Which is fine, because thats what DataFramesMeta is for.

pdeffebach:
Cool I did the `copyto!` method described above. Hopefully this PR has made some progress. 

Note that I use `df[idx1:idx2,, k] = scalar` so that will have to change when we get `setindex` sorted out for DataFrames.

src/DataFramesMeta.jl
```diff
@@ -392,10 +391,35 @@ function transform(g::GroupedDataFrame; kwargs...)
     idx1 = [1; 1 .+ idx2[1:end-1]]
     for (k, v) in kwargs
         first = v(g[1])
-        result[k] = Array{eltype(first)}(undef, size(result, 1))
-        result[idx1[1]:idx2[1], k] = first
-        for i in 2:length(g)
-            result[idx1[i]:idx2[i], k] = v(g[i])
+        if first isa AbstractVector 
+            if length(first) != size(g[1], 1)
```
nalimilan: AFAICT this check should be done for all groups, right?

src/DataFramesMeta.jl
```diff
@@ -392,10 +391,35 @@ function transform(g::GroupedDataFrame; kwargs...)
...
+                throw("If a function returns a vector, the result " * 
+                      "must have the same length as the groups it " *
+                      "operates on")
+            end
+            result[k] = Array{eltype(first)}(undef, size(result, 1))
```
nalimilan: It should be faster to also store `result[k]` in a variable here and work directly on it below, as a data frame lookup is a bit costly. You'd also need to do that when reallocating the vector of course.

src/DataFramesMeta.jl
```diff
@@ -392,10 +391,35 @@ function transform(g::GroupedDataFrame; kwargs...)
...
+            result[idx1[1]:idx2[1], k] = first
+            for i in 2:length(g)
+                out = v(g[i])
+                S = eltype(out)
+                T = eltype(result[k])
+                if !(S <: T)
+                    result[k] = copyto!(similar(result[k]), promote_type(S, T), result[k])
```
nalimilan: You need to provide the indices of values which have been filled, as copying values which are still undefined will throw an error (not for `Int`, but e.g. for `String` columns).

src/DataFramesMeta.jl
pdeffebach: Probably? I ommitted it because the extra check seemed slow. 

Its the same function and similar input for each group (different selections of the same vector). Unless there are some really degenerate functions which change output size erratically based on the size of the input vector, it probably isn't necessary.

src/DataFramesMeta.jl
pdeffebach: fixed

src/DataFramesMeta.jl
nalimilan: You can never know what the user-provided function will return, and we really don't want to return invalid results. Are these checks so expensive? In my tests, they have a noticeable impact, but not so large either.

nalimilan:
> It looks like DataFramesMeta is still on REQUIREE, but we need to add Tables as a dependency for `Tables.allocatecolumn`. Isn't there a Bot that does this for us?

No, this needs to be done manually. The bot translates REQUIRES to Project.toml.

pdeffebach:
Okay I made the requested changes and added Tables to REQUIRE and the grouping test. I'm not sure what version caps of REQUIRE to put though.

src/DataFramesMeta.jl
pdeffebach: Fixed

src/DataFramesMeta.jl
pdeffebach: added a check

src/DataFramesMeta.jl
```diff
@@ -1,6 +1,7 @@
 module DataFramesMeta
 
-using DataFrames
+using DataFrames 
+import Tables: allocatecolumn
```
nalimilan: `import` is for when you want to override a function. Here it would be `using`. But since it's used only once better fully qualify it.

src/DataFramesMeta.jl
```diff
@@ -392,11 +392,44 @@ function transform(g::GroupedDataFrame; kwargs...)
...
+                      "must have the same length as the groups it " *
+                      "operates on")
+            end
+            t = Tables.allocatecolumn(eltype(first), size(result, 1))
+            t[idx1[1]:idx2[1]] = first
+            for i in 2:length(g)
+                out = v(g[i])
+                if length(out) != size(g[i], 1)
+                    throw("If a function returns a vector, the result " * 
+                      "must have the same length as the groups it " *
```
nalimilan: Incorrect indentation.

src/DataFramesMeta.jl
```diff
@@ -392,11 +392,44 @@ function transform(g::GroupedDataFrame; kwargs...)
...
+                      "operates on")
+                end
+                S = eltype(out)
+                T = eltype(t)
+                if !(S <: T || promote_type(S, T) <: T)
+                    t = copyto!(Tables.allocatecolumn(promote_type(S, T), size(result, 1)), 
+                        t[1:idx2[i-1]])
```
nalimilan: That's not what I meant (this will create a copy). Do `copy!(Tables.allocatecolumn(...), 1, t, 1, idx2[i-1])` (hopefully I got it right).

Also fix indentation.

test/grouping.jl
```diff
@@ -22,4 +23,8 @@ d = DataFrame(n = 1:20, x = [3, 3, 3, 3, 1, 1, 1, 2, 1, 1, 2, 1, 1, 2, 2, 2, 3,
...
+g = groupby(d, :a)
+@test isequal(@transform(g, t = mean(:b))[:t], [2.0, 2.0, 2.0, missing, missing])
```
nalimilan: Can you also test what happens when the function returns vectors of different types? For example, return either `[1]` or `[1.0]` depending on the group.

pdeffebach: Added.

src/DataFramesMeta.jl
pdeffebach: fixed.

src/DataFramesMeta.jl
pdeffebach: Ah i see now. Thanks. Fixed this.

src/DataFramesMeta.jl
pdeffebach: fixed.

pdeffebach:
Okay things seem ready. Let me know if you want me to look into the Tables versioning.

test/grouping.jl
nalimilan: That function always returns vectors of the same type. :-/

nalimilan:
DataFrames doesn't indicate any version requirement for Tables.jl, so I guess it's OK to do that here too.

test/grouping.jl
pdeffebach: does the line below this work? I thought I had it in that last commit...

nalimilan: Well, no, it always return a value of the same type. You need something like `:b[1] == 1 ? [1] : [1.0]`.

nalimilan: ...or better `:b[1] == 1 ? [0.0] : [:b[1]]`, which is more likely to catch bugs since values change. BTW, the test would be better if there were more than two groups and values were not sorted.

pdeffebach: You are totally right. This behavior we need to pin down. 

If the first result is a scalar, then any 1-length vector down the line works. But if the first result if a vector of length one, it will throw an error that the length type doesnt match the group's length. 

We could do this vector-scalar check in every iteration of the loop instead of just the first one. 

```
d = DataFrame(n = 1:20, x = [3, 3, 3, 3, 1, 1, 1, 2, 1, 1, 2, 1, 1, 2, 2, 2, 3, 1, 1, 2])
g = groupby(d, :x)
@transform(g, t = :n[1] == 1 ? [0], [0.0]) # error because we check length for vector outputs
@transform(g, t = :n[1] == 1 ? 0 :  [0.0]) # no error
```

nalimilan: Yes, you can copy the error message I throw in that case at https://github.com/JuliaData/DataFrames.jl/pull/1520.

But that's not what I meant originally, I just got the length of the vectors wrong. So we need tests both for cases where the lengths don't match, and for cases where lengths match but types change (the latter shouldn't throw an error). BTW, to cover all code paths you should also check that an error is thrown when the first vector doesn't have the right length, and when the first vector has the right length but the second doesn't.

Finally, as noted in an old comment, it would be nice to test returning a `CategoricalArray` and a `CategoricalValue`/`CategoricalString` (I know, writing tests is the worst part of development...).

pdeffebach:
Added more error checks and am working on all the tets. Here is an error I ran into: 

```
julia> c = CategoricalArray([1,2,1,2])
4-element CategoricalArray{Int64,1,UInt32}:
 1
 2
 1
 2

julia> t = Any[1,2,3,4]
4-element Array{Any,1}:
 1
 2
 3
 4

julia> t[1:2] .= c[1]
ERROR: MethodError: no method matching length(::CategoricalValue{Int64,UInt32})
Closest candidates are:
  length(::Core.SimpleVector) at essentials.jl:571
  length(::Base.MethodList) at reflection.jl:728
  length(::Core.MethodTable) at reflection.jl:802
  ...
Stacktrace:
 [1] _similar_for(::UnitRange{Int64}, ::Type, ::CategoricalValue{Int64,UInt32}, ::Base.HasLength) at .\array.jl:532
 [2] _collect(::UnitRange{Int64}, ::CategoricalValue{Int64,UInt32}, ::Base.HasEltype, ::Base.HasLength) at .\array.jl:563
 [3] collect(::CategoricalValue{Int64,UInt32}) at .\array.jl:557
 [4] broadcastable(::CategoricalValue{Int64,UInt32}) at .\broadcast.jl:609
 [5] broadcasted(::Function, ::CategoricalValue{Int64,UInt32}) at .\broadcast.jl:1134
 [6] top-level scope at none:0
```

Broadcasted assignment doesn't seem to work for `CategoricalValue`s. Should I file an issue there?

pdeffebach:
On second thought, we should have all the scalar broadcasted `.=` be `.= Ref(x)` because
1. For the forseeable future, DataFramesMeta will be using `=` instead of `.=` because of its reliance on keyword arguments. 
2. The mental model of a `@transform` operation for a grouped DataFrame doesn't include boradcasted assignment. It could just as equally be `fill(x, length(g[i])` under the hood. When a user inputs a Tuple, for example, we can infer that they want each element of the array to be a Tuple. 

The question is how to make this explicit to the user.

test/grouping.jl
```diff
@@ -23,8 +23,23 @@ d = DataFrame(n = 1:20, x = [3, 3, 3, 3, 1, 1, 1, 2, 1, 1, 2, 1, 1, 2, 2, 2, 3,
...
+# Transform tests
+d = DataFrame(a = [1,1,1,2,2,3,3,1], 
+              b = Any[1,2,3,missing,missing,6.0,5.0,4], 
+              c = CategoricalArray([1,2,3,1,2,3,1,2]))
 g = groupby(d, :a)
-@test isequal(@transform(g, t = mean(:b))[:t], [2.0, 2.0, 2.0, missing, missing])
-@test isequal(@transform(g, t = fill(:b[1], length(:b))), [1, 1, 1, missing, missing])
+## Scalar output 
+# Type promotion Int -> Float
+@test @transform(g, t = :b[1])[1, :t] == 1.0
```
nalimilan: You'd need `===` to actually check the type. But that doesn't allow checking the eltype: you need two checks, one for values and one for the type of the array. While you're at it, better also check all values in the array.

Same below.

src/DataFramesMeta.jl
```diff
@@ -404,6 +404,10 @@ function transform(g::GroupedDataFrame; kwargs...)
                     throw("If a function returns a vector, the result " * 
                           "must have the same length as the groups it operates on")
                 end
+                if !(out isa AbstractVector)
+                    throw("return value must not change its kind (single " *
```
nalimilan: Actually the message needs to be adapted a bit here, since what matters is whether a vector is returned or not. So maybe this:
> return value must be an `` `AbstractVector` `` for all groups or for none of them.

Also, better check that result is a vector before checking its length, which may not be possible for non-vectors. And `elseif` would be appropriate.

src/DataFramesMeta.jl
```diff
@@ -417,6 +421,10 @@ function transform(g::GroupedDataFrame; kwargs...)
             t[idx1[1]:idx2[1]] .= first
             for i in 2:length(g)
                 out = v(g[i])
```
nalimilan: The length check is missing here.

nalimilan:
> Broadcasted assignment doesn't seem to work for `CategoricalValue`s. Should I file an issue there?

Good catch. I'll look into it. We should probably mark `CategoricalValue` as behaving like scalars.

> On second thought, we should have all the scalar broadcasted `.=` be `.= Ref(x)` because
> 
>     1. For the forseeable future, DataFramesMeta will be using `=` instead of `.=` because of its reliance on keyword arguments.
> 
>     2. The mental model of a `@transform` operation for a grouped DataFrame doesn't include boradcasted assignment. It could just as equally be `fill(x, length(g[i])` under the hood. When a user inputs a Tuple, for example, we can infer that they want each element of the array to be a Tuple.
> 
> 
> The question is how to make this explicit to the user.

Sounds right. I don't think the user needs to know anything about this, that's really an implementation detail. We could also do that using a combination of `fill!` and `view`, not sure whether one approach is faster than the other.

BTW, have you benchmarked this a bit?

pdeffebach:
I just did some benchmarking with 
```
t[idx1[1]:idx2[1]] .= first # and in the loop
```
vs. 
```
t[idx1[1]:idx2[1]] = fill(first, size(g[1], 1)) # and in the loop
```
Results indicate we should use `fill`
```
 df = DataFrame(a = [1:400; 1:400], b = randn(800))
 g = groupby(df, :a)
 function manygroups(g::GroupedDataFrame)
    @transform(g, t = :b[1])
end
using BenchmarkTools
@btime manygroups($g);

#= With Ref
julia> @btime manygroups($g);
  690.468 μs (6641 allocations: 290.84 KiB)
julia> @btime manygroups($g);
  683.470 μs (6641 allocations: 290.84 KiB)
=#

#= With fill
julia> @btime manygroups($g);
  423.145 μs (6930 allocations: 351.61 KiB)
julia> @btime manygroups($g);
  418.480 μs (6930 allocations: 351.61 KiB)
=#
```

pdeffebach:
Let me know if there are any performance issues to think about (`@inbounds`?). 

Sorry I didn't get a chance to implement [this](https://github.com/JuliaStats/DataFramesMeta.jl/pull/101#issuecomment-422313466) but my impression is that that concept could be applied more generally to DataFramesMeta, right? 

Something like 
1. pass through and figure out which columns to be used
2. put it in a type stable object, like a TypedDataFrame 
3. Do operations on that
4. put back in the output datafame.

nalimilan:
`fill!` with a view would be even better than `fill` since it avoids an allocation. Also try with more rows per group, as the performance difference will be more visible. BTW, when I talked about benchmarking, I was also thinking about comparing with the current implementation.

> Sorry I didn't get a chance to implement [this](https://github.com/JuliaStats/DataFramesMeta.jl/pull/101#issuecomment-422313466) but my impression is that that concept could be applied more generally to DataFramesMeta, right?
> 
> Something like
> 
>     1. pass through and figure out which columns to be used
> 
>     2. put it in a type stable object, like a TypedDataFrame
> 
>     3. Do operations on that
> 
>     4. put back in the output datafame.

Yes, I think that's very promising for DataFramesMeta. But let's leave that for another PR.

test/grouping.jl
```diff
@@ -22,4 +23,31 @@ d = DataFrame(n = 1:20, x = [3, 3, 3, 3, 1, 1, 1, 2, 1, 1, 2, 1, 1, 2, 2, 2, 3,
...
 @test @based_on(g, nsum = sum(:n))[:nsum] == [99, 84, 27]
 
+# Transform tests
+d = DataFrame(a = [1,1,1,2,2,3,3,1], 
+              b = Any[1,2,3,missing,missing,6.0,5.0,4], 
+              c = CategoricalArray([1,2,3,1,2,3,1,2]))
+g = groupby(d, :a, sort = false)
+## Scalar output 
+# Type promotion Int -> Float
+@test isequal(@transform(g, t = :b[1])[:t],
```
nalimilan: `isequal` doesn't check types. Writing `1.0` and `6.0` is even misleading since it's strictly equivalent to `1` and `6`. You need either to store the result of the operation and check its type manually. Using `.===` as you did below is a clever trick, but it's not enough to ensure that the element type is e.g. `Union{Float64,Missing}` rather than `Any`.

test/grouping.jl
```diff
@@ -22,4 +23,31 @@ d = DataFrame(n = 1:20, x = [3, 3, 3, 3, 1, 1, 1, 2, 1, 1, 2, 1, 1, 2, 2, 2, 3,
...
+@test (@transform(g, t = isequal(:b[1], 1) ? :b[1] : "a")[:t] .=== 
+                  [1, 1, 1, 1,"a" ,"a" ,"a" ,"a"]) |> all
+## Vector output 
+# Normal use
+@test isequal(@transform(g, t = :b .- mean(:b))[:t],
+              [-1.5, -0.5, 0.5, 1.5, missing, missing, 0.5, -0.5])
+# Type promotion
+@test (@transform(g, t = isequal(:b[1], 1) ? fill(1, length(:b)) : fill(2.0, length(:b)))[:t] .=== 
+                  [1.0, 1.0, 1.0, 1.0, 2.0, 2.0, 2.0, 2.0]) |> all
+# Vectors of different types
```
nalimilan: Vectors are also of different types above. What's different here is that they promote to `Any`.

test/grouping.jl
```diff
@@ -22,4 +23,31 @@ d = DataFrame(n = 1:20, x = [3, 3, 3, 3, 1, 1, 1, 2, 1, 1, 2, 1, 1, 2, 2, 2, 3,
...
+@test (@transform(g, t = isequal(:b[1], 1) ? :b : fill("a", length(:b)))[:t] .=== 
+                  [1, 2, 3, 4, "a", "a", "a", "a"]) |> all
+# Categorical Categorical Array 
+# Scalar
+@test @transform(g, t = :c[1])[:t] == CategoricalArray([1, 1, 1, 1, 1, 1, 3, 3])
```
nalimilan: Here too, `CategoricalArray` on the RHS has no effect, you need to check the type.

test/grouping.jl
nalimilan: You probably don't need `.===` if you use `typeof`, at least not if all numbers are expected to be of the same type (which is the case here if I'm not mistaken).

test/grouping.jl
pdeffebach: should be fixed.

test/grouping.jl
pdeffebach: should be fixed.

pdeffebach:
We have performance regressions. Here is a gist with a script I ran to test things:

https://gist.github.com/pdeffebach/931b3edc4a9a134f7e3a96bfde7068c8#file-testgroupedtransform-jl

Some of the changes we made *should* have been faster
* Using a temporary vector instead of calling the `result` DataFrame each time. 
* `@view`

But we added more safety checks
* `length` of output vector
* check types for type promotion.

nalimilan:
Mmm, it's a bit disappointing, but we have indeed only made the code more complex and more robust, so the absence of performance improvements (and the clear regression when vector result with many groups) isn't too surprising. I guess we should merge this even with the regression since it's more correct, and improve it later. Without applying the more complex strategy developed above, moving the loop to a separate function specialized on the vector type should really make this faster.

test/grouping.jl
```diff
@@ -22,4 +23,40 @@ d = DataFrame(n = 1:20, x = [3, 3, 3, 3, 1, 1, 1, 2, 1, 1, 2, 1, 1, 2, 2, 2, 3,
 g = groupby(d, :x, sort=true)
 @test @based_on(g, nsum = sum(:n))[:nsum] == [99, 84, 27]
 
+# Transform tests
+d = DataFrame(a = [1,1,1,2,2,3,3,1], 
+              b = Any[1,2,3,missing,missing,6.0,5.0,4], 
+              c = CategoricalArray([1,2,3,1,2,3,1,2]))
+g = groupby(d, :a, sort = false)
```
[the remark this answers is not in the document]
pdeffebach: fixed.

test/grouping.jl
```diff
@@ -22,4 +23,40 @@ d = DataFrame(n = 1:20, x = [3, 3, 3, 3, 1, 1, 1, 2, 1, 1, 2, 1, 1, 2, 2, 2, 3,
...
+@test all(t .=== correct) && typeof(t) == typeof(correct)
+# Type promotion
+t = @transform(g, t = isequal(:b[1], 1) ? fill(1, length(:b)) : fill(2.0, length(:b)))[:t] 
+correct = Float64[1.0, 1.0, 1.0, 1.0, 2.0, 2.0, 2.0, 2.0]
+@test all(t .=== correct) && typeof(t) == typeof(correct)
+# Vectors of different types
+t = @transform(g, t = isequal(:b[1], 1) ? :b : fill("a", length(:b)))[:t]
+correct = Any[1, 2, 3, 4, "a", "a", "a", "a"]
+@test all(t .=== correct) && typeof(t) == typeof(correct)
+# Categorical Categorical Array
```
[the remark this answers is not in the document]
pdeffebach: fixed.

test/grouping.jl
```diff
@@ -22,4 +23,40 @@ d = DataFrame(n = 1:20, x = [3, 3, 3, 3, 1, 1, 1, 2, 1, 1, 2, 1, 1, 2, 2, 2, 3,
...
+# Transform tests
+d = DataFrame(a = [1,1,1,2,2,3,3,1], 
+              b = Any[1,2,3,missing,missing,6.0,5.0,4], 
+              c = CategoricalArray([1,2,3,1,2,3,1,2]))
+g = groupby(d, :a, sort = false)
+## Scalar output 
+# Type promotion Int -> Float
+t = @transform(g, t = :b[1])[:t] 
+correct = [1.0, 1.0, 1.0, 1.0, missing, missing, 6.0, 6.0]
+@test all(t .=== correct) && typeof(t) == typeof(correct)
```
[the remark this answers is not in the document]
pdeffebach: Fixed, except for categorical arrays because their type signatures are complicated.

test/grouping.jl
```diff
@@ -22,4 +23,40 @@ d = DataFrame(n = 1:20, x = [3, 3, 3, 3, 1, 1, 1, 2, 1, 1, 2, 1, 1, 2, 2, 2, 3,
...
+# Type promotion Number -> Any
+t = @transform(g, t = isequal(:b[1], 1) ? :b[1] : "a")[:t]
+correct = Any[1, 1, 1, 1,"a" ,"a" ,"a" ,"a"]
```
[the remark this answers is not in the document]
pdeffebach: fixed.

src/DataFramesMeta.jl
```diff
@@ -392,11 +391,47 @@ function transform(g::GroupedDataFrame; kwargs...)
...
+                    throw("If a function returns a vector, the result " * 
+                          "must have the same length as the groups it operates on")
+                end
+                S = eltype(out)
+                T = eltype(t)
+                if !(S <: T || promote_type(S, T) <: T)
+                    t = copyto!(Tables.allocatecolumn(promote_type(S, T), size(result, 1)), 
+                                1, t, 1, idx2[i-1])
+                end
+                t[idx1[i]:idx2[i]] = out
```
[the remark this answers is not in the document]
pdeffebach: Added, for no other reason than to show we are serious.

src/DataFramesMeta.jl
```diff
@@ -392,11 +391,47 @@ function transform(g::GroupedDataFrame; kwargs...)
     idx1 = [1; 1 .+ idx2[1:end-1]]
     for (k, v) in kwargs
         first = v(g[1])
-        result[k] = Array{eltype(first)}(undef, size(result, 1))
-        result[idx1[1]:idx2[1], k] = first
-        for i in 2:length(g)
-            result[idx1[i]:idx2[i], k] = v(g[i])
+        if first isa AbstractVector 
+            if length(first) != size(g[1], 1)
+                throw("If a function returns a vector, the result " *
```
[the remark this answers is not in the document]
pdeffebach: Did `throw(ArgumentError())` because that's what you have in the DataFrames PR.

pdeffebach:
Putting all that code in a function would be easy to implement. But the functions (the `v`s in `(k, v)` are all still anonymous functions on `AbstractDataFrame` objects, so there isn't much to be gained in the loop.

nalimilan:
> Putting all that code in a function would be easy to implement. But the functions (the `v`s in `(k, v)` are all still anonymous functions on `AbstractDataFrame` objects, so there isn't much to be gained in the loop.

Yeah, but in the other PR it definitely made a difference. It can cut by half the number of calls which go thourgh dynamic dispatch.

test/grouping.jl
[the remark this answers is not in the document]
nalimilan: At least you can also do `isequal(t, correct)`. Also, `t isa CategoricalVector{String}` is enough, you don't need to specify all parameters.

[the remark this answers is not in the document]
pdeffebach: Fixed. and the `isequal` has been changed.

pdeffebach:
Should we add it in a function now? Or should we wait until the next PR. I'm not 100% sure what you mean by putting it in a function.

nalimilan:
Depends on whether you can find the time or not. The idea is to have an external function to most of the work for each group as in https://github.com/JuliaData/DataFrames.jl/pull/1520. It would take as arguments `g`, `v`, `i` and `t` and consist mostly of the `for` loop. That function would have one method for `first::AbstractVector` and another method for other cases. Each time `allocatecolumn` needs to be called, it would call itself recursively with the newly allocated column, ensuring type stability of iterations.

pdeffebach:
I would like to see this through so let's add the recursive function. 

Do you mean something like this? 
```
function _transform(first::AbstractVector, g::GroupedDataFrame, v::Function, i::Int, t::Vector, N::Int, starts, ends)
    out = v(g[i])
    # check that its a vector and length is right here...
    S = eltype(out)
    T = eltype(t)
    if !(S <: T || promote_type(S, T) <: T)
       # Problem: We have to calculate v(g[i]) again for however many times we promote
        _transform(first, g, v, i, Tables.allocatecolumn(promote_type(S, T), N))
    end
    t[starts[i]:ends[i]] = out
    # _transform(first, g, v, i+1, t) make it truly recursive? 
    return t 
end 
```

nalimilan:
Yes, something like that. Thanks to the `return`, you don't need the line commented out at the bottom to make it "truly recursive". But you need to assign `out` to `t` before calling `_transform`, and pass `i+1` to it rather than `i`; that avoids calling `v` twice.

pdeffebach:
I just wrote a Recursive implementation that seems good to me (maybe). The good news is that we didn't hurt performance, as seen my comment on the gist above. Given that I probably made a number of errors in this re-write which will hurt performance, I am cautiously optimistic about this.

src/DataFramesMeta.jl
```diff
@@ -388,19 +387,61 @@ end
...
+    end
+    S = eltype(current)
+    T = eltype(previous)
+    if !(S <: T || promote_type(S, T) <: T)
+        t = copyto!(Tables.allocatecolumn(promote_type(S, T), length(t)), 
+                    1, t, 1, ends[i-1])
+    end
+    t[starts[i]:ends[i]] = current
+    if i != length(g) 
+        return _transform!(t, i + 1, current, v(g[i+1]), g, v, starts, ends)
```
nalimilan: The idea is that you'd call `_transform` only when re-allocating the column. Here you call it for each group, which will trigger a stack overflow for a large number of groups (and is probably slower).

pdeffebach: Okay. I think I was misunderstanding .

We still have a for-loop. The only difference, really, is that we push the type promotion into a function. Which only ever has to run recursively once or twice.

pdeffebach:
Thanks for the feedback. I'm still a bit confused though. The implementation I have here is as follows 

```
function _transform(t, out, indexes)
    if we need to promote
         make promoted_vector 
         _transform(promoted_vector, out, indexes)
    end
    t[indexes] = out
    return t # otherwise we return a small view
end 

function transform(g, v)
    first = v(g[1])
    allocate array t
     input first value
     for i in 2:length(g)
          t = _transform(t, v(g[i], indexes)
     end
end
```

I hope I am understanding #1520 correctly, because I think this is more or less the same logic used in `combine!` there. 

The issue is that the function `_transform` sometimes modifies the input vector `t`, in the case of no type promotion, and sometimes returns a new vector all together. We also change the type of `t` throughout the loop, but it seems like that is inevitable. Thankfully Julia assigns by reference so the extra `t = _transform` isn't allocating. But it needs to be there in the case where we promote type. 

>  But you need to assign `out` to `t` before calling` _transform`, and pass `i+1` to it rather than `i`; that avoids calling v twice.

I don't think that works because we need to either increment recursively for all iterations or none of them. If we have a `for` loop, we have to stick to a single counter and not call out of order. 

Let me know if this works. Thanks.

pdeffebach:
Ah I see it! `colstart` there means you only restart the loop at the necessary location! That's clever, thanks.

pdeffebach:
I think this is following the implementation in #1520, more or less. Performance is better than before without the error checks, and about on par with the error checks. 

Let me know if there are big performance traps I am stepping in!

src/DataFramesMeta.jl
```diff
@@ -388,19 +387,86 @@ end
...
+            t = _transform!(Tables.allocatecolumn(typeof(first), size(result, 1)), first, 1, g, v, starts, ends)
         end
+        result[k] = t
     end
     return result
 end
 
+function _transform!(t::AbstractVector, first::AbstractVector, start::Int, g::GroupedDataFrame, v::Function, starts::Vector, ends::Vector)
+    # handle the first case 
+    j = fill_column_vec!(t, first, starts[start], ends[start], size(g[start], 1))
```
nalimilan: `j` is a bit weird here. I used that name in the other PR because it was an index. Maybe `promoted` just like in the other place? Or even better `newT`/`newtype` (in both places): `promoted` sounds like a Boolean.

src/DataFramesMeta.jl
```diff
@@ -388,19 +387,86 @@ end
...
+            t = _transform!(Tables.allocatecolumn(eltype(first), size(result, 1)), first, 1, g, v, starts, ends)
+        else 
+            t = _transform!(Tables.allocatecolumn(typeof(first), size(result, 1)), first, 1, g, v, starts, ends)
         end
+        result[k] = t
     end
     return result
 end
 
+function _transform!(t::AbstractVector, first::AbstractVector, start::Int, g::GroupedDataFrame, v::Function, starts::Vector, ends::Vector)
```
nalimilan: Wrap all lines at 92 chars.

src/DataFramesMeta.jl
```diff
@@ -388,19 +387,86 @@ end
...
+         throw(ArgumentError("Return value must be an `AbstractVector` for all groups or for none of them"))
+    elseif length(out) != len
+        throw(ArgumentError("If a function returns a vector, the result " * 
+                            "must have the same length as the groups it operates on"))    
+    end     
+    elout = eltype(out)
+    T = eltype(t)
+    promoted = promote_type(elout, T)
+    if (elout <: T || promoted <: T)
+        t[startpoint:endpoint] = out
```
nalimilan: Better put the `return nothing` here for clarity.

src/DataFramesMeta.jl
```diff
@@ -388,19 +387,86 @@ end
...
+    else 
+        return promoted
+    end
+    return nothing 
+end 
+
+function fill_column_any!(t::AbstractVector, out, startpoint::Int, endpoint::Int)
+    if (out isa AbstractVector)
```
nalimilan: No parentheses.
```suggestion
    if out isa AbstractVector
```

src/DataFramesMeta.jl
```diff
@@ -388,19 +387,86 @@ end
...
+    if !(out isa AbstractVector)
+         throw(ArgumentError("Return value must be an `AbstractVector` for all groups or for none of them"))
+    elseif length(out) != len
+        throw(ArgumentError("If a function returns a vector, the result " * 
+                            "must have the same length as the groups it operates on"))    
+    end     
+    elout = eltype(out)
+    T = eltype(t)
+    promoted = promote_type(elout, T)
+    if (elout <: T || promoted <: T)
```
nalimilan: No parentheses:
```suggestion
    if elout <: T || promoted <: T
```

src/DataFramesMeta.jl
```diff
@@ -388,19 +387,86 @@ end
...
+end 
+
+function fill_column_any!(t::AbstractVector, out, startpoint::Int, endpoint::Int)
+    if (out isa AbstractVector)
+         throw(ArgumentError("Return value must be an `AbstractVector` for all groups or for none of them"))
+    end     
+    typout = typeof(out)
+    T = eltype(t)
+    promoted = promote_type(typout, T)
+    if (typout <: T || promoted <: T)
```
nalimilan: No parentheses.
```suggestion
    if typout <: T || promoted <: T
```

src/DataFramesMeta.jl
```diff
@@ -388,19 +387,86 @@ end
...
+        @views t[startpoint:endpoint] .= Ref(out)
```
nalimilan: `@views` isn't needed on assignment.
```suggestion
        t[startpoint:endpoint] .= Ref(out)
```

src/DataFramesMeta.jl
```diff
@@ -388,19 +387,86 @@ end
...
+end
+
+function _transform!(t::AbstractVector, first::Any, start::Int, g::GroupedDataFrame, v::Function, starts::Vector, ends::Vector)
+    # handle the first case 
+    j = fill_column_any!(t, first, starts[start], ends[start])
+    @assert j === nothing
+    for i in (start+1):length(g)
+        out = v(g[i])
+        promoted = fill_column_any!(t, out, starts[i], ends[i])
+        if !(promoted === nothing)
```
nalimilan:
```suggestion
        if promoted !== nothing
```

src/DataFramesMeta.jl
```diff
@@ -388,19 +387,86 @@ end
...
 end
 
+function _transform!(t::AbstractVector, first::AbstractVector, start::Int, g::GroupedDataFrame, v::Function, starts::Vector, ends::Vector)
+    # handle the first case 
+    j = fill_column_vec!(t, first, starts[start], ends[start], size(g[start], 1))
+    @assert j === nothing
+    for i in (start+1):length(g)
+        out = v(g[i])
+        promoted = fill_column_vec!(t, out, starts[i], ends[i], size(g[i], 1))
+        if !(promoted === nothing)
```
nalimilan:
```suggestion
        if promoted !== nothing
```

src/DataFramesMeta.jl
```diff
@@ -1,7 +1,6 @@
 module DataFramesMeta
 
-using DataFrames
-
+using DataFrames, Tables
```
nalimilan: Re-add empty line.

test/grouping.jl
nalimilan:
```suggestion
# Vectors whose eltypes promote to Any
```

pdeffebach:
Thanks for the feedback!

pdeffebach:
The performance for the vector fill leaves something to be desired, it seems, using the gist [here](https://gist.github.com/pdeffebach/931b3edc4a9a134f7e3a96bfde7068c8#file-testgroupedtransform-jl). With this new commit the "fewer vectors" case is ~ 180 on this branch and ~ 150 on master. 

I'm not sure what's going on, though.

nalimilan:
Maybe add `@inline fill_column_vec!`? The reason why I used separate functions in my PR is that it allowed specializing on the number of columns to copy. But here since there's a single column there's no advantage and the dynamic dispatch can have some overhead. You could even move them inside the function, the code would be clearer.

pdeffebach:
I added `@noinline` and things improved some. It's worth nothing that for the "many groups" case, 5000 as opposed to 1000, there are performance gains of approx. 10%

src/DataFramesMeta.jl
```diff
@@ -405,9 +405,29 @@ end
 
 function _transform!(t::AbstractVector, first::AbstractVector, start::Int, 
                      g::GroupedDataFrame, v::Function, starts::Vector, ends::Vector)
+    @inline function fill_column_vec!(t::AbstractVector, out, startpoint::Int, endpoint::Int, len::Int)
```
nalimilan: What I mean when I said you could move the code inside the function is that if you use `@inline`, you can just drop the function barrier and put the code directly in the parent function.

src/DataFramesMeta.jl
```diff
@@ -405,9 +405,29 @@ end
...
+            return newtype
+        end
+        return nothing
+    end 
+
     # handle the first case 
-    newtype = fill_column_vec!(t, first, starts[start], ends[start], size(g[start], 1))
-    @assert newtype === nothing
+    newtype_first = fill_column_vec!(t, first, starts[start], ends[start], size(g[start], 1))
+    #@assert newtype_first === nothing
```
nalimilan: I don't expect this check to be costly. Is it?

src/DataFramesMeta.jl
pdeffebach: It should probably still be a function, since we call it twice per `_transform!` function. Once for the first case and once for the rest. We need to call it twice because we have to compute `first` in `transform(::GroupedDataFrame,...)` so not treating the first case separately would require calculating `first` twice.

src/DataFramesMeta.jl
```diff
@@ -388,19 +388,94 @@ end
...
+        if !(out isa AbstractVector)
+            throw(ArgumentError("Return value must be an `AbstractVector` for all groups or" *
+                                "for none of them"))
+        elseif length(out) != len
+            throw(ArgumentError("If a function returns a vector, the result " * 
+                                "must have the same length as the groups it operates on"))    
+        end     
+        elout = eltype(out)
+        T = eltype(t)
+        newtype = promote_type(elout, T)
```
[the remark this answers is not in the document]
pdeffebach: wow thats really smart that julia allows for that. But alas there isn't any speed gain from this:

```
        if typout <: T || (newtype = promote_type(typout, T)) <: T
            t[startpoint:endpoint] .= Ref(out)
            return nothing
        else 
            return newtype
        end
```

src/DataFramesMeta.jl
nalimilan: Right. Should be OK as-is then.

src/DataFramesMeta.jl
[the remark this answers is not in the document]
nalimilan: Ok, too bad.

tshort:
+1 for me. Thanks for all the work @pdeffebach, and thanks for the great help and interaction, @nalimilan.

PR #107, opened 2018-08-22
Remove import of Compat

bkamins:
Removes import of Compat in `src` and `test`.


bkamins:
Actually here we need to remove only one `Compat` import, as other imports are removed in #106.

PR #113, opened 2018-10-14
lightquery compatibility

bramtayl:


nalimilan:
Can you explain what this achieves?

Also, I don't understand why we need functions accepting `Tuple{Function, Expr}` but which only use the first element.

bramtayl:
Ok nevermind I have a better way to do this

Issue #110, opened 2018-09-18
Add a @transform!  mutating version


jlperla:
In order to port over Stata and R code, it is nice to have succinct calculations of new columns based on existing values.  Right now,
```julia
using DataFrames, DataFramesMeta
df = DataFrame(a=[1; 2; 3], b=[10; 12; 13])
f(a, b) = a + b
df =@transform(df, c = f.(:a, :b))
```
But it would be great if this could instead be:
```julia
using DataFrames, DataFramesMeta
df = DataFrame(a=[1; 2; 3], b=[10; 12; 13])
f(a, b) = a + b
@transform!(df, c = f.(:a, :b))
```

Is this a relatively easy change for someone to add a PR?

jlperla:
Thanks for the idea.  Actually, let me make it even better, since variadic makes a lot of sense here.
```julia
using DataFrames, DataFramesMeta
df = DataFrame(a=[1; 2; 3], b=[10; 12; 13])
f(a, b) = a + b
@transform!(df, c = f.(:a, :b), d = :a + :b)
```
instead of
```julia
using DataFrames, DataFramesMeta
df = DataFrame(a=[1; 2; 3], b=[10; 12; 13])
f(a, b) = a + b
df =@transform(df, c = f.(:a, :b))
df =@transform(df, d = :a + :b)
```

pdeffebach:
One thing to note is that if you are doing a mutating transforms over and over again, `@with` is probably a more concise macro to use. 

```
@with df begin 
    df.t = f.(:a, :b)
    df.t = :a + :b
end
```

pdeffebach:
This would actually a bit tough for a grouped data frame. Because `transform(g::GroupedDataFrame, args...)` uses `DataFrame(g)` in the first line, it's always going to make a copy. Perhaps we could do 

```
d = sort!(g.parent, g.cols)
..all the other stuff
```

pdeffebach:
I just created a PR for this, but I think we should have more discussion about whether or not it is the right strategy. 

It's nice that this is an opinionated package where there is one way of doing each thing. I might prefer to have just one, albeit copying, macro for each operation than face a splintered API. 

Despite opening the PR I might wonder if we can convince users to get used to copying, and make sure that everything is as performant as possible.

matthieugomez:
[...]

PR #119, opened 2019-01-03
WIP add mutating macros

pdeffebach:
This is in reference to #110 . I hope to add 
- [x] `@transform!`
- [ ]  `@where!`
- [ ] `@select!` 
- [ ] `@orderby!`

### DIFF
--- a/src/DataFramesMeta.jl
+++ b/src/DataFramesMeta.jl
@@ -283,14 +283,6 @@ macro where(d, args...)
 end
 
 
-##############################################################################
-##
-## select - select columns
-##
-##############################################################################
-
-select(d::AbstractDataFrame, arg) = d[arg]
-
 
 ##############################################################################
 ##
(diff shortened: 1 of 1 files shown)


src/DataFramesMeta.jl
```diff
@@ -690,6 +705,18 @@ end
 ##
 ##############################################################################
 
+function select!(d::Union{AbstractDataFrame, AbstractDict}; kwargs...)
+    tokeep = [k for (k, v) in kwargs]
```
nalimilan: Could use a generator. Also don't we support integer column indices in addition to names?

src/DataFramesMeta.jl
```diff
@@ -378,14 +378,18 @@ end
...
     end
-    return result
+    return d
 end
 
+function transform(d::Union{AbstractDataFrame, AbstractDict}; kwargs...)
+    return transform!(copy(d))
+end
+
+# No mutating transform since transform returns a full DataFrame
```
nalimilan: As noted at https://github.com/JuliaData/DataFramesMeta.jl/issues/110#issuecomment-425736923, why not mutate the parent?

src/DataFramesMeta.jl
pdeffebach: Currently they are separate functions? there is a `select(d, args)` on its own somewhere. This method is for `@select(df, :a, :b, :z = :a + :b)`, which personally i wouldn't mind deprecating in favor of putting that variable creation in `@transform` only.

nalimilan: I just wondered whether `k` could be an integer here, resulting from e.g. `@select(df, 1, 2)`.

pdeffebach: I just looked into this and fixing it would probably a useful separate PR

@select(df, ...) takes you to:

1. `select(df, arg) = df[arg]` which works with only one integer
2. `select(df; kwargs...)`. In the process, if there is more than one argument, like `@select(df, :a, :b)` it gets transformed to `select(df, a = :a, b = :b)`. This doesn't work with integers at all. so `@select(df, 1, 2)` does not work. But the error message is not particularly insightful. It makes me think that the keyword argument creation step is failing silently. 

```
ERROR: MethodError: no method matching select(::DataFrame, ::Int64, ::Int64)
Closest candidates are:
  select(::AbstractDataFrame, ::Any) at C:\Users\Pdeffebach\.julia\dev\DataFramesMeta\src\DataFramesMeta.jl:292
  select(::Union{AbstractDataFrame, AbstractDict}; kwargs...) at C:\Users\Pdeffebach\.julia\dev\DataFramesMeta\src\DataFramesMeta.jl:722
Stacktrace:
 [1] top-level scope at C:\Users\Pdeffebach\.julia\dev\DataFramesMeta\src\DataFramesMeta.jl:752
```

pdeffebach:
Closed via #216

PR #122, opened 2019-01-19
Remove methods not belonging to DataFramesMeta.jl

bkamins:
This should resolve https://github.com/JuliaData/DataFramesMeta.jl/issues/121.

In general I would also remove `DataFrame(::GroupedDataFrame)` method. If you feel it is useful let us add it to DataFrames.jl.

@nalimilan - any opinion?

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,3 +1,3 @@
 julia 0.7
-DataFrames 0.18
-Tables 0.1.14
+DataFrames 0.17.1
+Tables 0.1.15
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,3 +1,3 @@
 julia 0.7
-DataFrames 0.13
-Tables
+DataFrames 0.18
+Tables 0.1.14
--- a/src/DataFramesMeta.jl
+++ b/src/DataFramesMeta.jl
@@ -805,6 +805,6 @@ end
 ##
 ##############################################################################
 
-DataFrame(g::GroupedDataFrame) = collect(g)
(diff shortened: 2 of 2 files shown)


bkamins:
sorry for the larger PR than indented my editor automatically strips trailing spaces from the code.

bkamins:
and in general if we want to keep DataFramesMeta.jl up to date with DataFrames.jl I think we should move it somewhat "closer" to it as I am always unsure if DataFramesMeta.jl will still work after we change DataFrames.jl. @nalimilan: do you have an opinion what should be our policy?

nalimilan:
I agree we should remove that method from here. It could make sense to have it in DataFrames. Is it defined for internal use or for convenience? If the latter, better deprecate it now in favor of `combine`, so that we can add it later in DataFrames if we want.

bkamins:
It is defined for convenience. But I would decide if we want it in DataFrames.jl. If yes then we can simply move it there and synchronize the releases without deprecation period.

So the question is if we want do define it in DataFrames.jl - I would say yes, as I do not see the reason why we should not do it (but I guess it might be related to your general design ideas in https://github.com/JuliaData/DataFrames.jl/issues/1256 so please decide).

nalimilan:
OK, let's move it to DataFrames then.

bkamins:
This should be merged only after https://github.com/JuliaData/DataFrames.jl/pull/1689 is merged and released.

bkamins:
I will still need to test it after we release DataFrames.jl

PR #131, opened 2019-07-13
Sync with DataFrames.jl 0.19

bkamins:
and fix many other things that stopped working as the time passed.

One significant change is that we stop supporting `AbstractDict` as DataFrames.jl diverged from dictionary API.


bkamins:
This PR will be ready to test and merge when DataFrames.jl 0.19 is out.

bkamins:
Thank you! 😄

PR #133, opened 2019-08-09
:z not defined in @select

EconometricsBySimulation:
`:z not defined in @select(df, x2 = 2 * :x, :y, :z)` example. Not useful for example anyways.


nalimilan:
Thanks!

PR #141, opened 2020-01-24
transform(::AbstractDataFrame) broadcasting

Byrth:
Change transform(::AbstractDataFrame) so it automatically broadcasts non-Array results.

transform(::GroupedDataFrame) already worked this way.

Although the behavior was inconsistent between the two in the current state, this will probably break things for people. Also, this is untested behavior (because the tests passed both before and after).


Byrth:
Doesn't work. Needs to bake.

bkamins:
Thank you for working on it anyway 😄.

PR #147, opened 2020-05-06
sync with DataFrames 0.21 release

bkamins:


bkamins:
This PR makes no changes. Only makes DataFramesMeta.jl work correctly with DataFrames.jl v0.21.

@pdeffebach - as discussed on slack it would be great if you moved forward and implemented more functionality for this package!

PR #151, opened 2020-06-13
Add warning to users about lack of support for `Not`, `Between`, and `All`

### LINKED ISSUE #150: `Not` does not work in `@linq` or `@select`
Currently, it seems like one can not use the `Not` selector in `@linq` or `@select` queries.

For example,

```julia
df = DataFrame(x=rand(3), y=rand(3))
@linq df |> select(Not(:x)) # <-- does not work
@select(df, Not(:x)) # <-- does not work
```
does not work as `:x` is replaced with the values of the `:x` column. 

Instead one needs to escape `:x` in the `Not` call.

```julia
@linq df |> select(Not(^(:x)))
# or...
@select(df, Not(^(:x)))
```

Is this the desired behavior? If so, it may need to be documented (is it?). Are there any cases in which replacing a symbol with the column data is needed in a call to `select`?

pdeffebach:
Fixes #150

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,13 @@ Add additional columns based on keyword arguments.
 @transform(df, newCol = cos.(:x), anotherCol = :x.^2 + 3*:x .+ 4)
 ```
 
+!!! note 
+    
+    Versions of DataFrames.jl `.0.19` and above support the operators `Between`, `All`,
+    and `Not` when selecting and transforming columns. DataFramesMeta does not currently
+    support this syntax. 
+
+
 ## `@byrow!`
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -79,6 +79,13 @@ Add additional columns based on keyword arguments.
 @transform(df, newCol = cos.(:x), anotherCol = :x.^2 + 3*:x .+ 4)
 ```
 
+!!! note 
+    
+    Versions of DataFrames.jl `.0.19` and above support the operators `Between`, `All`,
+    and `Not` when selecting and transforming columns. DataFramesMeta does not currently
```
bkamins, line 85:
```suggestion
    Versions of DataFrames.jl `0.19` and above support the operators `Between`, `All`,
    and `Not` when selecting and transforming columns. DataFramesMeta.jl does not currently
```

pdeffebach:
It's fine that it will be on master only, I don't think anyone is reading the source code markdown for docs. 

Yes! I think we can implement it as expanding to `Vector`s pretty easily. But I don't want to promise anything right now.

PR #152, opened 2020-06-19
Playing around with using `DataFrames.transform`

pdeffebach:
@bkamins 

This is a PR where I try to get a version of `@transform` which uses `DataFrames.transform` under the hood and still passes all tests. 

I have written out a few comments on how to approach this, to start. I will work on this incrementally and ping you for questions about how to best approach problems I encounter.


bkamins:
OK - just write to me in this issue. I get notified about questions posted here.

pdeffebach:
Okay, the first issue I've encountered consistently is that I don't know how to make the `keys` of the `membernames` dictionary become a vector of symbols rather than `Expr`. Take a look at what's going on in `transform_helper`. 

```
julia> @transform(df, y = :x1 + :x2)
t = quote
    #= /home/peterwd/Documents/Development/DataFramesMeta/src/DataFramesMeta.jl:492 =#
    function var"##300"(var"##302", var"##301")
        #= /home/peterwd/Documents/Development/DataFramesMeta/src/DataFramesMeta.jl:493 =#
        var"##301" + var"##302"
    end
    #= /home/peterwd/Documents/Development/DataFramesMeta/src/DataFramesMeta.jl:497 =#
    Any[:(:x2), :(:x1)] => var"##300"
end
ERROR: ArgumentError: idxs[1] has type Expr; only Integer, Symbol, or string values allowed when indexing by vector
```

I really can't find the combination of `:(` and `$` that will make these evaluate to `Symbols` so that we get `transform(df, [:x2, :x1] => var"##300")`. 

Second, We need to figure out how to define the function before it enters the transform call. Even if I got the vector of symbols to work above, we still never evaluated the expression defining the temporary function. I'm not sure where this should be done.

bkamins:
So this is a non trivial task unfortunately. What you need to do is to transform the following expression:
```
julia> macro transform(args...)
       return args
       end
@transform (macro with 1 method)

julia> dump(@transform(df, y = :x1 + :x2))
Tuple{Symbol,Expr}
  1: Symbol df
  2: Expr
    head: Symbol =
    args: Array{Any}((2,))
      1: Symbol y
      2: Expr
        head: Symbol call
        args: Array{Any}((3,))
          1: Symbol +
          2: QuoteNode
            value: Symbol x1
          3: QuoteNode
            value: Symbol x2
```
(this is what `@transform` gets in the input)

into this:
```
julia> dump(:(transform(df, [:x1, :x2] => (var1, var2) -> var1 + var2)))
Expr
  head: Symbol call
  args: Array{Any}((3,))
    1: Symbol transform
    2: Symbol df
    3: Expr
      head: Symbol call
      args: Array{Any}((3,))
        1: Symbol =>
        2: Expr
          head: Symbol vect
          args: Array{Any}((2,))
            1: QuoteNode
              value: Symbol x1
            2: QuoteNode
              value: Symbol x2
        3: Expr
          head: Symbol ->
          args: Array{Any}((2,))
            1: Expr
              head: Symbol tuple
              args: Array{Any}((2,))
                1: Symbol var1
                2: Symbol var2
            2: Expr
              head: Symbol block
              args: Array{Any}((2,))
                1: LineNumberNode
                2: Expr

julia> dump(:(transform(df, [:x1, :x2] => (var1, var2) -> var1 + var2)).args[3])
Expr
  head: Symbol call
  args: Array{Any}((3,))
    1: Symbol =>
    2: Expr
      head: Symbol vect
      args: Array{Any}((2,))
        1: QuoteNode
          value: Symbol x1
        2: QuoteNode
          value: Symbol x2
    3: Expr
      head: Symbol ->
      args: Array{Any}((2,))
        1: Expr
          head: Symbol tuple
          args: Array{Any}((2,))
            1: Symbol var1
            2: Symbol var2
        2: Expr
          head: Symbol block
          args: Array{Any}((2,))
            1: LineNumberNode
              line: Int64 1
              file: Symbol REPL[13]
            2: Expr
              head: Symbol call
              args: Array{Any}((3,))
                1: Symbol +
                2: Symbol var1
                3: Symbol var2
```
(this is what should be an output)

Note that I show two outputs here as `dump` truncates the printout.

Now in the output `var1` and `var2` should be dynamically generated inside a macro for the purposes of hygene.

Unfortunately I cannot offer a quick and short code that implements it, but the steps should probably be:
1. Take `@transform(df, y = :x1 + :x2)`
2. Identify the `:x1 + :x2` part as an expression
3. Find symbols in it.
4. from these symbols make an expression giving `[:x1, :x2]` vector
5. substitute `:x1` and `:x2` for temporary internal names inside a macro (my `var1` and `var2)
6. then transform `:x1 + :x2` into `(var1, var2) -> var1 + var2)`
7. Finally link point 5 and 6 into `[:x1, :x2] => (var1, var2) -> var1 + var2)`
8. Repeatedly make points 2 to 7 for each argument after the first in the macro call.
9. Compose values provided in point 8.

As you see (and probably already realized) - all this is hard (of course doable). So maybe the way to go is to leave `@transform` as it is (and just clean it up if there are some rough edges present).

pdeffebach:
I think I would really like to get this working using `DataFrames.transform` if possible, better to keep feature parity if it's a wrapper for that function. 

If you have time, take a look at the current iteration of `transform_helper`. 

1. handled, at least for one new variable
2. handled via existing machinery
3. handled via existing machinery
4. Not handled. I cannot seem to get `[:x1, :x2]` as a vector in the expression we wish to evaluate. 

I'm very close! As you can see from the current iteration of `transform_helper`. I just can't get the vector to be one of `Symbol`s rather than `QuoteNodes`. I think I need someone's help who is very good at metaprogramming. perhaps @piever can also join in and help me out if possible?

5. Handled via existing machinery
6. handled via existing machinery
7. Close to being correct, I realized that 

```
julia> transform(df, [:x1, :x2] => 
       function(v1, v2)
       v1 + v2
       end
       => :y
       )
```

is valid syntax, so we can keep existing machinery. 

8. Not yet handled but I think not hard

bkamins:
So let us start with 4:

This should be easy - just use `Expr` e.g.:
```
julia> Expr(:vect, :x1, :x2)
:([x1, x2])
```

piever:
I think this is pretty much the same as `parse_table_call!` in StatsPlots, see [here](https://github.com/JuliaPlots/StatsPlots.jl/blob/master/src/df.jl#L67). The idea is that you do a first pass through the expression and replace symbols (or things inside `cols()`) with a new variable (obtained with `gensym`). While you do this replacement, you `push!` to a `syms` and a `vars` vectors the symbols you replaced and the corresponding variable that was inserted.

Then, StatsPlots does it a bit differently, but here you should just use those `gensym` vars to create the anonymous function and the symbols you collected as the first argument of the `Pair`.

pdeffebach:
Thanks, I'm getting closer to understanding the issue. The problem is that `n` in the following is not a vector of symbols but a vector of expressions. This makes it hard to construct the `Symbol` vector I need in the `DataFrames.transform` call. 

Let me know if I should move this to discourse since I don't want to ask too many specific questions to too small an audience. 

```
    kw = args[1]
    
    funname = gensym()
    body = replace_syms!(kw.args[2], membernames)
    n = collect(keys(membernames)) # convert to Symbol somehow?
    @show n
    t = quote
        $n => function $funname($(values(membernames)...))
            $body 
        end
    end
    @show t
    
    quote 
        $DataFrames.transform($x,$t) 
    end
end
```

bkamins:
What are the expressions you are getting. If it is just a wrapper around a `Symbol` then this should help:
```
julia> Expr(:a).head
:a
```

Probably asking on Slack is fastest. But you can also ask here, but it would be easier if you asked an atomic issue, as now in order to say anything I would need to clone your whole branch locally and design my own tests (it is much easier to copy-paste a working code that is stand alone and show the issue you want to resolve).

bkamins:
(and what I have written is even more important if you ask on Slack/Discourse, as you should expect that people will not understand the context of your question)

pdeffebach:
Here is the set of functions that ensures things works

```
using DataFrames; import DataFramesMeta
function make_vec_to_fun(kw)
    output = kw.args[1]
    funname = gensym()

    body = DataFramesMeta.replace_syms!(kw.args[2], membernames)
    
    inputs = broadcast(s -> s.args[1], keys(membernames))

    t = quote
        $inputs => function $funname($(values(membernames)...))
            $body 
        end => $(QuoteNode(output))
    end
end

    t = [make_vec_to_fun(arg) for arg in args]

    quote 
        $DataFrames.transform($x, $(t...))
    end
end

macro transform(x, args...)
    esc(transform_helper(x, args...))
end

df = DataFrame(x1 = [1, 2], x2 = [3, 4])
@transform(df, y = :x1 + :x2, z = :x1)
```

bkamins:
Nice - I have played around with this and all looks good!

pdeffebach:
I think my current solution, above, using 

```
inputs = broadcast(s -> s.args[1], keys(membernames))
```

doesn't work. Here is a full MWE

```
function make_vec_to_fun(kw)
    output = kw.args[1]
    funname = gensym()

    body = DataFramesMeta.replace_syms!(kw.args[2], membernames)
    
    inputs = broadcast(s -> s.args[1], keys(membernames))

    t = quote
        $inputs => function $funname($(values(membernames)...))
            $body 
        end => $(QuoteNode(output))
    end
end

function transform_helper2(x, args...)

    t = [make_vec_to_fun(arg) for arg in args]

    quote 
        $DataFrames.transform($x, $(t...))
    end
end

macro transform2(x, args...)
    esc(transform_helper2(x, args...))
end

df = DataFrame(rand(2,2))
y = :x1
@transform2(df, z = cols(y)) # no method for `cols`
@transform2(df, z = cols(y) .+ :x2) # type `Symbol` has no field `args`
```

In the current `with_helper`, which escapes `cols(y)` just fine, we have 

```
function with_helper(d, body)
    funname = gensym()
    body = replace_syms!(body, membernames)
    if isempty(membernames)
        body
    else
        quote
            function $funname($(values(membernames)...))
                $body
            end
            $funname($((:($d[!, $key]) for key in keys(membernames))...)) # this is the important part
        end
    end
end
```

I feel like all we are missing is instead of 

```
inputs = broadcast(s -> s.args[1], keys(membernames))
```

we need something inside the quote that looks like 

```
:([$k for k in $(keys(membernames))]) => $funname...
```

But I have no idea the correct order of `:()` and `$` to get this to work. I feel like it should be feasible, since clearly a lot of work has gone into the escaping rules and making them work with the `quote`. Any thoughts on this are appreciated.

bkamins:
Have you tried?
```
$(Expr(:vect, keys(membernames)...)) => function $funname($(values(membernames)...))
```
(have not tested it though so there might be some hygiene issues)

pdeffebach:
Your advice works! I never would have gotten that, though I certainly tried a bunch of similar syntaxes. 

I think the next step is to work on `@select` a bit more. You wrote a function called `expandargs` which makes a kewword argument `x2 = :x2` in the call

```
@select(df, :x2)
```

but this isn't the most robust method, for instance it fails with `cols` and with `[:x1, :x2]`. Both of these should work. Hopefully I can make a function like `replace_syms!` that works on things other than keyword arguments so this just works.

bkamins:
I have never looked at what `expandargs` does, but I can check it if needed.

pdeffebach:
Great progress today! I think `expandargs` is not needed. Current implementation defines 

```
cols(x) = x
```

so that `cols` works when the expression is evaluated in the scope of the `DataFrames.transform` call.

I also got greedy since now that the `y = :x1 .+ :x2` is *never* evaluated as a keyword argument, we don't have to worry about keyword argument parsing, meaning it's super easy to make 

```
t = :y
@transform2(df, cols(t) = :x1 .+ :x2)
```

work. 

Additionally, we can now use `Between`, `Not`, etc. for non-keyword inputs, i.e. `select`ing columns. Since this is evaluated normally, we don't need `cols` for this. 

See full code and some examples below: 

```
cos(x) = x

function make_vec_to_fun(kw::Expr)

    if kw.head == :(=) || kw.head == :kw
        output = kw.args[1]
        
        funname = gensym()
        body = DataFramesMeta.replace_syms!(kw.args[2], membernames)
        @show typeof(kw.args[1])
        # inputs = broadcast(s -> s.args[1], keys(membernames))
        if kw.args[1] isa Symbol
            t = quote
                $(Expr(:vect, keys(membernames)...)) => function $funname($(values(membernames)...))
                    $body 
                end => $(QuoteNode(output))
            end
        elseif DataFramesMeta.onearg(kw.args[1], :cols)
            t = quote
                $(Expr(:vect, keys(membernames)...)) => function $funname($(values(membernames)...))
                    $body 
                end => $(output)
            end   
        end

        return t
    else
        println("hit this branch")
        return kw
    end
end

function make_vec_to_fun(kw::QuoteNode)
    return kw
end

function transform_helper2(x, args...)

    t = [make_vec_to_fun(arg) for arg in args]

    quote 
        $DataFrames.transform($x, $(t...))
    end
end

macro transform2(x, args...)
    esc(transform_helper2(x, args...))
end

function based_on_helper2(x, args...)

    t = [make_vec_to_fun(arg) for arg in args]

    quote 
        $DataFrames.combine($x, $(t...))
    end
end

macro based_on2(x, args...)
    esc(based_on_helper2(x, args...))
end

function select_helper2(x, args...)
    t = [make_vec_to_fun(arg) for arg in args]

    quote 
        $DataFrames.select($x, $(t...))
    end
end

macro select2(x, args...)
    esc(select_helper2(x, args...))
end

# df = DataFrame(rand(2,2))
# t = :x2
# s = :y
# @transform2(df, y = :x1 .+ :x2)
# @transform2(df, y = :x1 .+ cols(t))
# @transform2(df, [:x1, :x2])
# @transforms(df, [:x1, :x2], y = :x1 .+ :x2)
# @select2(df, Not(:x1))
# @select2(df, Not(t))
```

bkamins:
Great. Have you checked "corner case" names like `body`, `transform`, etc. as variable names to make sure we get hygiene right?

pdeffebach:
I'm going to open a PR to add a bunch of tests for existing DataFramesMeta functions that covers corner cases. 

Then I will start a PR in earnest with these functions.

bkamins:
OK. Great!

PR #155, opened 2020-07-17
Improve tests for `@select` and `@transform`

pdeffebach:
In preparation of using `DataFrames.transform` behind the scenes, I am writing a bunch of tests to 

1. Make sure we catch edge cases such as using reserved words 
2. Add `@test_throws` so we know exactly when we are adding new behavior or fixing current bugs. 

Please feel free to add tests as you see fit.

### DIFF
--- a/test/dataframes.jl
+++ b/test/dataframes.jl
@@ -98,7 +98,7 @@ s = [:i, :g]
     @test_throws ErrorException @eval @transform(df, cols(newvar) = :i)
     @test_throws MethodError @eval @transform(df, n = sum(Between(:i, :t)))
     @test @transform(df, n = :i).n === df.i
-    @test @transform(df, n = cols(s).i) == df.i
+    @test @transform(df, n = cols(s).i).n == df.i
 end
 
 @testset "@select" begin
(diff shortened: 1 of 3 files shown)


pdeffebach:
Okay I think this is ready for review. 

Remaining tests are 

* `transform` with a grouped data frame. But the tests for behavior are pretty good and the macro parsing stuff should work the exact same as `@transform` with a normal data frame
* `@where` which I don't have a strategy for replacing with `filter` but whose tests could still go in this PR
* `@orderby` which is like `@where`. Tests could be added but I don't have much of an idea how behavior will change. 
* `@with`, for which more tests could be added but whose behavior is at a low risk of changing, since it is just `replace_syms!` which `@transform2` etc. use without any modification.

bkamins:
CI fails

pdeffebach:
Sorry, `@eval` woes. It's fixed now, feel free to have a look.

test/dataframes.jl
```diff
@@ -4,64 +4,191 @@ module TestDataFrames
 using Test
 using DataFrames
 using DataFramesMeta
+using Statistics
+
+const ≅ = isequal
+
+df = DataFrame(
```
bkamins, line 11: it is better to wrap all variables inside a `@testset` (even if some definitions are repeated) as it makes `@testset` easier to maintain in the long run.

test/dataframes.jl
```diff
@@ -4,64 +4,191 @@ module TestDataFrames
...
+cq = :c
+
+gr = "g"
+ir = "i"
+tr = "t"
+yr = "y"
+cr = "c"
+
+@testset "@transform" begin
+    @test @transform(df, n = :i).n == df.i
```
bkamins, line 34: in several tests I would also check:
1. if the original `df` was not changed
2. if columns got copied (`!==`) and not just aliased (`===`)
3. not only the transformed column but if all other columns are retained

bkamins:
Nice tests - I have tried to check that you have not removed any old tests and it seems so, but just please confirm that this is the case.

test/dataframes.jl
bkamins: Same for other macros

pdeffebach: > if columns got copied (`!==`) and not just aliased (`===`)

It looks like that is not actually the behavior

```
julia> df = DataFrame(rand(2,2));

julia> @transform(df, n = :x1).n === df.x1
true
```

Presumably this is not actually behavior we want, so this goes in the `limits of` testset. Do you agree?

pdeffebach:
Okay i've pinned down some `===` behavior for `@transform` and `@select`. 

I can confirm that I didnt delete any tests. Just moved a lot of things.

test/dataframes.jl
bkamins: Yes - for sure we want `@tranform` to make a copy like `transform`, as this is the expected behavior (so I would put the current tests in the set of tests that are likely to be broken after the update).

In general we might consider adding `copycols` kwarg to `@transform` but this can be left for later.

test/dataframes.jl
pdeffebach: I've implmented that where I can but in many cases the use of `@eval` means i need the data in global scope.

bkamins: ah - right!

pdeffebach:
Bumping this. Would be great to get this merged so we can make real changes to `@transform`

bkamins:
Sure - but can you please resolve merge conflicts first and make sure all tests are correct against current master (without it I am not able to merge this). Thank you!

pdeffebach:
Tests pass! Ready for merging

bkamins:
Thank you!

pdeffebach:
Now the real work begins!

PR #154, opened 2020-07-10
Deprecates `@byrow!` in favor of `@byrow`.

henriquebecker91:
The `@byrow!` is simply duplicated and renamed to `@byrow` (also exported). The only difference is that one of them triggers the deprecation warning and the other does not. `@byrow!` documentation mention the deprecation and `@byrow` documentation just points to `@byrow!` documentation.

PR #157, opened 2020-07-23
Make `cols` in `@byrow!` escape propertly

### LINKED ISSUE #156: in `@byrow!`, `cols` escapes to the full vector instead of a scalar.
```
julia> df = DataFrame(a = ["a1", "a2"], b = [101, 102])
2×2 DataFrame
│ Row │ a      │ b     │
│     │ String │ Int64 │
├─────┼────────┼───────┤
│ 1   │ a1     │ 101   │
│ 2   │ a2     │ 102   │

julia> t = :b
:b

julia> @byrow! df begin
       :b = cols(t) + 100
       end
ERROR: MethodError: no method matching +(::Array{Int64,1}, ::Int64)
Closest candidates are:
  +(::Any, ::Any, ::Any, ::Any...) at operators.jl:529
  +(::Complex{Bool}, ::Real) at complex.jl:301
  +(::Missing, ::Number) at missing.jl:115
  ...
Stacktrace:
 [1] macro expansion at ./REPL[13]:2 [inlined]
 [2] macro expansion at /home/peterwd/.julia/packages/DataFramesMeta/c67UK/src/byrow.jl:50 [inlined]
 [3] ##263(::Array{Int64,1}, ::Array{Int64,1}) at /home/peterwd/.julia/packages/DataFramesMeta/c67UK/src/DataFramesMeta.jl:71
 [4] top-level scope at /home/peterwd/.julia/packages/DataFramesMeta/c67UK/src/DataFramesMeta.jl:73
 [5] top-level scope at /home/peterwd/.julia/packages/DataFramesMeta/c67UK/src/byrow.jl:49
```

Intuitively, `cols(t)` should behave just like if we used the `Symbol` literal `:b`, i.e. it should return a scalar. Instead, we get the whole column, `df.b`.

pdeffebach:
Fixes #156. 

Very far from a fix, still. Having trouble with the escaping rules as usual.


pdeffebach:
I have it working. However we should wait to merge the #154 and #155 to avoid conflicts.

src/byrow.jl
```diff
@@ -11,6 +11,10 @@ export @byrow!
 # ":(:(x))" with ":x[row]".
 function byrow_replace(e::Expr)
     # Traverse the syntax tree of e
+    if e.head == :call && e.args[1] == :cols
+       return Expr(:ref, e, :row)
+    end
+    #dump(e)
```
bkamins, line 17: probably not needed

src/byrow.jl
```diff
@@ -43,6 +47,9 @@ byrow_find_newcols(x, newcol_decl) = (x, Any[])
 
 function byrow_helper(df, body)
     e_body, e_newcols = byrow_find_newcols(body, Any[])
+
+    t = byrow_replace(e_body)
```
bkamins, line 51: also this is not needed - right?

bkamins:
it would be nice to add some tests :). Thank you for working on it.

pdeffebach:
Fixed the diff. I will rebase after the other PRs are merged and write tests then.

pdeffebach:
This is ready for a review. In particular

1. This doesn't work with multiple columns, which I think is okay since we decided it's undefined behavior in `@transform` etc. 
2. Any ideas for more tests?

pdeffebach:
Thanks for suggesting to put weird things in `:cols` because it has very odd behavior that I don't want. 

One solution might be to make a deprecation for multi-argument selectors everywhere. The relevant code is currently

```
    else
        quote
            function $funname($(values(membernames)...))
                $body
            end
            $funname($((:($d[!, $key]) for key in keys(membernames))...))
        end
    end
```

We could change `$d[!, $key]` to $`getproperty($d, $key)`

This would throw errors everywhere people are using multi-argument selectors, not just in `@byrow`.

bkamins:
Given the changes you want to do in DataFramesMeta.jl that will be fundamental maybe we do not have to do a deprecation I think (CC @nalimilan). The problem with `getproperty` is that it does not support integer indexing. I would rather do the check if `key isa DataFrames.ColumnIndex` and throw an error if it is not before calling `funname`

bkamins:
(in this I assume that you have a "grand strategy" how things should work in the future :) and this is one of the pieces of the puzzle)

pdeffebach:
I wish I had more of a grand strategy for specific behavior. What we really need is for `cols([:x, :y])` to return an iterator, so its at least somewhat close to working with `[df.x, df.y]`. 

In the `@byrow` context, a `DataFrameRow` or named tuple is natural and in a `@transform` context, a named tuple or a `eachcol` is natural. Either way I would not expect `cols([:x, :y])` to  return a DataFrame when `[:x, :y]` returns a vector of vectors.

nalimilan:
What does `cols` do currently when there are multiple columns? Anyway I would expect `n = [:a, :b]; cols(n)` to be equivalent to `[:a, :b]`, right?

pdeffebach:
> What does `cols` do currently when there are multiple columns? Anyway I would expect `n = [:a, :b]; cols(n)` to be equivalent to `[:a, :b]`, right?

It is not!

DataFramesMeta currently does

1. Look through the code for all `Symbol`s and things wrapped in `cols`. Creates a function `foo` and a collection of column selectors `X`
2. For each `x` in `X`, of these, call `df[:, x]`. This means that if `x` is `[:a, :b]` this results in a DataFrame being inputted into the function `foo`. 

The problem is that when the macros are run and collecting all the potential column selectors, we don't want to evaluate, for example `cols(n)` where `n` is `[:a, :b]`. 

Consequently, at the macro level, we can't create an expression that will give us `[df.a, df.b]`.

nalimilan:
>     2\. 
> 
> 
> The problem is that when the macros are run and collecting all the potential column selectors, we don't want to evaluate, for example `cols(n)` where `n` is `[:a, :b]`.
> 
> Consequently, at the macro level, we can't create an expression that will give us `[df.a, df.b]`.

Sorry, I don't understand that part.

pdeffebach:
Let's say we have the expression

```
@transform(df, y = sum([:a, :b])
```

DataFramesMeta doesn't have to know that `[:a, :b]` is a vector of symbols. All it does is go through the exact expression and when it sees a `:a` it replaces it with `df[:, :a]`, so the end expression evaluated is (roughly)

```
y = [df[:, :a], df[:, :b]]
```

When we use `cols(n)` where `n` is a vector of symbols, we can't do that "just replace the symbol" process. We have to know what `n` evaluates to before we do that.

nalimilan:
But can't we translate `cols(n)` to `if n isa Symbol ? df[!, n] elseif n isa AbstractVector map(n -> df[!, n]) else...`? That will be handled at runtime.

pdeffebach:
> But can't we translate `cols(n)` to `if n isa Symbol ? df[!, n] elseif n isa AbstractVector map(n -> df[!, n]) else...`? That will be handled at runtime.

Yes, we could do that. It wouldn't be type stable, though, which is a main benefit of DataFramesMeta, right? 

For the time being, I think we should figure out a way to throw an error in the case of multiple selectors in this exact issue, for `@byrow`. Then since the syntax will be reserved we can come up with a more general behavior later.

bkamins:
> It wouldn't be type stable, though, which is a main benefit of DataFramesMeta, right?

It is one of the benefits, but I think in the design of DataFramesMeta.jl we should focus on UX not speed (it does not mean we want to ignore it, but largely you can do things in a fast and type stable way by other means). But here maybe both objectives can be combined?

nalimilan:
I don't think the type unstability is a problem for performance here as long as the resulting objects are passed as arguments to transform. For example, translating `n = [:a, :b]; @transform(df, y = sum(cols(n)))` into `transform(df, [:a, :b] => ((a, b) -> sum([a, b])) => :y)` will give fast code even if the `[:a, :b] => ((a, b) -> sum([a, b])) => :y` part is generated at run time. Are there cases where that wouldn't be workable?

Anyway I agree it's fine to have this throw an error for now.

pdeffebach:
> `n = [:a, :b]; @transform(df, y = sum(cols(n)))` into `transform(df, [:a, :b] => ((a, b) -> sum([a, b])) => :y)` will give fast code even if the `[:a, :b] => ((a, b) -> sum([a, b])) => :y` part is generated at run time. Are there cases where that wouldn't be workable?

There are still issues with this proposal. Consider a proposal where `n` gets turned into a vector of vectors. 

```
n = [:a, :b]
@transform(df, y = :c + sum(cols(n)))
```

The current DataFramesMeta parsing will make a function with two arguments. One is a single vector (from `:c`), the other is a vector of vectors, from `cols(n)`. 

There is no way to express this in our current DataFrames mini-language. 

```
transform(df, [:c, [:a, :b]] => fun => :y) # fails due to heterogenous inputs
transform(df, All(:c, [:a, :b]) => fun => :y) # fails because will pass 3 arguments to a 2 argument function
```

This is why I propose #2328, where we have a "default" multi-argument selector that for any iterator of `Symbol`s. 

Nonetheless, today I will work for a way to throw an error with `cols(n)` in `@byrow`

pdeffebach:
I added a little wrapper function for indexing into a vector in order to reserve `cols` with multi-argument selectors in `@byrow`. I think it's a reasonable solution. 

I also updated the docstring and ensured tests pass. 

I would welcome a review! Thank you.

bkamins:
Looks good. Maybe if you are changing this also change `_N = length($df[!, 1])` in `byrow_helper` to use `nrow`, as what we have now errors for data frame with no columns.

pdeffebach:
> Looks good. Maybe if you are changing this also change `_N = length($df[!, 1])` in `byrow_helper` to use `nrow`, as what we have now errors for data frame with no columns.

I'll do this in another PR. The code currently has bad macro hygine (`_N` gets brought into scope) and I need to fix that as well.

src/byrow.jl
```diff
@@ -12,9 +12,23 @@ export @byrow
 # ":(:(x))" with ":x[row]".
 function byrow_replace(e::Expr)
     # Traverse the syntax tree of e
+    if onearg(e, :cols)
+
+       # return Expr(:ref, e, :row)
```
nalimilan, line 17:
```suggestion
```

src/byrow.jl
```diff
@@ -12,9 +12,23 @@ export @byrow
...
+        return Expr(:call, :(DataFramesMeta.getsingleindex), e, :row)
+    end
+
     Expr(e.head, (isempty(e.args) ? e.args : map(byrow_replace, e.args))...)
 end
 
+function getsingleindex(df::AbstractDataFrame, idx)
```
nalimilan, line 24: Use short function syntax here and below?

test/dataframes.jl
```diff
@@ -69,6 +69,40 @@ df2 = @byrow df begin
     end
 end
 
+df = DataFrame(A = 1:3, B = [2, 1, 2])
+
```
nalimilan, line 73:
```suggestion
```

test/dataframes.jl
```diff
@@ -69,6 +69,40 @@ df2 = @byrow df begin
...
+    @eval TestDataFrames n = [:A, :B]
+    @test_throws ArgumentError @eval @byrow df begin cols(n) end
+
+    @eval TestDataFrames n = [1, 2]
+    @test_throws ArgumentError @eval @byrow df begin cols(n) end
+end
+
+
+
```
nalimilan, line 105:
```suggestion
```

nalimilan (commented):
DataFramesMeta's design isn't fresh enough in my mind to really comment on the implementation, but throwing an error sounds reasonable anyway.

test/dataframes.jl
pdeffebach: needs to be in global scope because of `@eval`.

pdeffebach:
Thanks! Made the changes.

nalimilan (approved):

bkamins:
@pdeffebach - OK to merge?

pdeffebach:
Yes! please merge

bkamins:
Thank you!

PR #161, opened 2020-08-28
Update README.md for Lazy.jl and Pipe.jl

xiaodaigh:


README.md
```diff
@@ -189,6 +190,26 @@ x_thread = @> begin
     @select(:meanX, :meanY, var = :b)
 end
 ```
+
+Please note that `Lazy.jl` exports the function `groupby` which would clash
```
bkamins, line 194: can you pleae remove `\`` in all places where a package like Lazy.jl is referred to (as it is a package name not a code)

xiaodaigh: done

bkamins (approved):
Looks good. A review from a native speaker is welcome.

nalimilan (approved):
Thanks. I've made the heading more general since the section mentions several macros now.

bkamins:
Thank you!

PR #163, opened 2020-08-29
Make DataFrames piping functions the backend for @transform, @select, @by, and @based_on

pdeffebach:
This is obviously inspired by [DataFramesMacros](https://github.com/matthieugomez/DataFramesMacros.jl) from @matthieugomez.

This PR introduces two macros, `@col` and `@row`. Here is the docstring introducting their behavior. 

    @col(kw)

    
`@col` transforms an expression of the form `z = :x + :y` into it's equivalent in 
DataFrames's "mini-language". Functions act column-wise. For a row-wise functions, see 
`@row`

### Details

Parsing follows the same convention as other DataFramesMeta macros, such as `@with`. All 
terms in the expression that are `Symbols` are treated as columns in the DataFrame, except 
`Symbol`s wrapped in `^`. To use a variable representing a column name, wrap the variable 
in `cols`. 

`@col` constructs an anonymous function based off the given expression. It then creates
a `source => fun => destination` pair that is suitable for the `select`, `transform`, and 
`combine` functions in DataFrames. 

### Examples

```
julia> @col z = :x + :y
[:x, :y] => (##595 => :z)
```

In the above example, `##595` is an anonymous function equivelent to the following 

```
(_x, _y) -> _x + _y
```

```
julia> df = DataFrame(x = [1, 2], y = [3, 4]);

julia> DataFrames.transform(df, @col z = :x .* :y)
2×3 DataFrame
│ Row │ x     │ y     │ z     │
│     │ Int64 │ Int64 │ Int64 │
├─────┼───────┼───────┼───────┤
│ 1   │ 1     │ 3     │ 3     │
│ 2   │ 2     │ 4     │ 8     │

julia> DataFrames.transform(df, [:x, :y] => ((_x, _y) -> _x .* _y) => :z)
2×3 DataFrame
│ Row │ x     │ y     │ z     │
│     │ Int64 │ Int64 │ Int64 │
├─────┼───────┼───────┼───────┤
│ 1   │ 1     │ 3     │ 3     │
│ 2   │ 2     │ 4     │ 8     │

```

This PR represents the initial step in having `DataFrames.transform`,  `DataFrames.select`, and `DataFrames.combine` be the backend for most DataFramesMeta functionality. In the future, these macros, or, the function that transform the expressions in these macros, will transform all inputs to `@transform` etc. 

Design decisions left to discuss

1. Many people have expressed interest in not forcing people to write `:x` and instead write `x` for `df.x`, similar to `dplyr` and Stata (`@transform(z = x + y)`). I support this, conditional on air-tight escaping rules. Should we start that transition now in the development of these macros?
2. Naming. Obviously `@col` is a less-than-ideal name given `Cols()` in DataAPI and `cols` for variable escaping. We should bikeshed over the name a bit. 
3. Multi-column selectors are broken for the time being, given you can't do `DataFrames.transform(df, [:x, [:y, :z]] => fun => :a)`. See discussions in [DataFrames](https://github.com/JuliaData/DataFrames.jl/issues/2328) and [DataFramesMeta](https://github.com/JuliaData/DataFramesMeta.jl/pull/157). 

I think this is a solid first step!

cc @bkamins, @nalimilan, @matthieugomez

cc DataFramesMeta users @tk3369

### DIFF
--- a/src/DataFramesMeta.jl
+++ b/src/DataFramesMeta.jl
@@ -93,20 +93,20 @@ macro col(kw)
     esc(fun_to_vec(kw))
 end
 
-# `combinetable` needs to be `true` when we have syntax of the form
+# `nolhs` needs to be `true` when we have syntax of the form
 # `@based_on(gd, fun(:x, :y))` where `fun` returns a `table` object.
 # We don't create the "new name" pair because new names are given
 # by the table.
-function fun_to_vec(kw::Expr; combinetable = false)
-    if kw.head === :(=) || kw.head === :kw || combinetable
+function fun_to_vec(kw::Expr; nolhs = false)
+    if kw.head === :(=) || nolhs
         membernames = Dict{Any, Symbol}()
-        if combinetable
+        if nolhs
             body = replace_syms!(kw, membernames)
         else
             body = replace_syms!(kw.args[2], membernames)
         end
 
-        if combinetable
+        if nolhs
             t = quote
                 $(Expr(:vect, keys(membernames)...)) =>
                 ($(Expr(:tuple, values(membernames)...)) -> $body)
@@ -116,7 +116,7 @@ function fun_to_vec(kw::Expr; combinetable = false)
             t = quote
                 $(Expr(:vect, keys(membernames)...)) =>
                 ($(Expr(:tuple, values(membernames)...)) -> $body) =>
-                $(QuoteNode(output))
+                $(output)
             end
         end
         return t
@@ -518,7 +518,7 @@ function based_on_helper(x, args...)
         !(first(args) isa QuoteNode) &&
         !(first(args).head == :(=) || first(args).head == :kw)
 
-        t = fun_to_vec(first(args); combinetable = true)
+        t = fun_to_vec(first(args); nolhs = true)
         quote
             $DataFrames.combine($t, $x)
         end
@@ -600,7 +600,7 @@ function by_helper(x, what, args...)
         !(first(args) isa QuoteNode) &&
         !(first(args).head == :(=) || first(args).head == :kw)
 
-        t = fun_to_vec(first(args); combinetable = true)
+        t = fun_to_vec(first(args); nolhs = true)
         quote
             $DataFrames.combine($t, $groupby($x, $what))
         end
--- a/src/DataFramesMeta.jl
+++ b/src/DataFramesMeta.jl
@@ -52,14 +52,14 @@ DataFrames's `source => fun => destination` syntax.
 
 ### Details
 
-Parsing follows the same convention as other DataFramesMeta macros, such as `@with`. All
-terms in the expression that are `Symbols` are treated as columns in the data frame, except
+Parsing follows the same convention as other DataFramesMeta.jl macros, such as `@with`. All
+terms in the expression that are `Symbol`s are treated as columns in the data frame, except
 `Symbol`s wrapped in `^`. To use a variable representing a column name, wrap the variable
 in `cols`.
 
-`@col` constructs an anonymous function based off the given expression. It then creates
+`@col` constructs an anonymous function `fun` based on the given expression. It then creates
 a `source => fun => destination` pair that is suitable for the `select`, `transform`, and
-`combine` functions in DataFrames.
+`combine` functions in DataFrames.jl.
 
 ### Examples
 
--- a/src/DataFramesMeta.jl
+++ b/src/DataFramesMeta.jl
@@ -95,60 +95,7 @@ macro col(kw)
     esc(fun_to_vec(kw; byrow = false))
 end
 
-"""
-    @row(kw)
-
-
-`@row` transforms an expression of the form `z = :x + :y` into it's equivalent in
-DataFrames's "mini-language". Functions act row-wise. For column-wise functions, see
-`@col`
-
-### Details
-
-Parsing follows the same convention as other DataFramesMeta macros, such as `@with`. All
-terms in the expression that are `Symbols` are treated as columns in the DataFrame, except
-`Symbol`s wrapped in `^`. To use a variable representing a column name, wrap the variable
-in `cols`.
-
-`@row` constructs an anonymous function based off the given expression. It then creates
-a `source => fun => destination` pair that is suitable for the `select`, `transform`, and
-`combine` functions in DataFrames.
-
-### Examples
-
-```julia
-julia> @row z = :x + :y
-[:x, :y] => (ByRow{var"###607"}(##607) => :z)
-
-```
-
-In the above example, `##607` is an anonymous function equivalent to the following
-
-```julia
-(_x, _y) -> _x + _y
-```
-
-The function is wrapped in `ByRow` indicating that DataFrames applies the function
-to each row of the data frame.
-
-```julia
-julia> df = DataFrame(x = [1, 2], y = [3, 4]);
-
-julia> DataFrames.transform(df, @row z = :x * :y)
-2×3 DataFrame
-│ Row │ x     │ y     │ z     │
-│     │ Int64 │ Int64 │ Int64 │
-├─────┼───────┼───────┼───────┤
-│ 1   │ 1     │ 3     │ 3     │
-│ 2   │ 2     │ 4     │ 8     │
-
-"""
-macro row(kw)
-    esc(fun_to_vec(kw; byrow = true))
-end
-
-
-function fun_to_vec(kw::Expr; byrow = false)
+function fun_to_vec(kw::Expr)
     # If expression is of the form
     #
     # `z = :x + :y` or
@@ -171,19 +118,11 @@ function fun_to_vec(kw::Expr; byrow = false)
         funname = gensym()
         body = replace_syms!(kw.args[2], membernames)
 
-        if byrow == false
-           t = quote
-               $(Expr(:vect, keys(membernames)...)) => function $funname($(values(membernames)...))
-                   $body
-               end => $(QuoteNode(output))
-           end
-        else
-            t = quote
-                $(Expr(:vect, keys(membernames)...)) => $(ByRow)(function $funname($(values(membernames)...))
-                    $body
-                end) => $(QuoteNode(output))
-            end
-        end
+        t = quote
+            $(Expr(:vect, keys(membernames)...)) => (function $funname($(values(membernames)...))
+                $body
+            end) => $(QuoteNode(output))
+    end
         return t
     else
         throw(ArgumentError("Expressions not of the form `y = f(:x)` currently disallowed."))
@@ -574,9 +513,11 @@ end
 ##############################################################################
 
 function based_on_helper(x, args...)
-    with_args =
-        with_anonymous(:($DataFrame($(map(replace_equals_with_kw, args)...))))
-    :(DataFrames.combine($with_args, $x))
+    t = [fun_to_vec(arg) for arg in args]
+
+    quote
+        $DataFrames.combine($x, $(t...))
+    end
 end
 
 """
@@ -644,8 +585,11 @@ end
 ##############################################################################
 
 function by_helper(x, what, args...)
-    :(DataFrames.combine($(with_anonymous(:($DataFrame($(map(replace_equals_with_kw, args)...))))),
-              DataFrames.groupby($x, $what)))
+    t = [fun_to_vec(arg) for arg in args]
+
+    quote
+        $DataFrames.combine($groupby($x, $what), $(t...))
+    end
 end
 
 """
@@ -729,40 +673,11 @@ end
 ##
 ##############################################################################
 
-
-function select(d::AbstractDataFrame; kwargs...)
-    result = typeof(d)()
-    for (k, v) in kwargs
-        result[!, k] = v
-    end
-    return result
-end
-
-function replace_equals_with_kw(e)
-    if e.head == :(=)
-        Expr(:kw, e.args[1], e.args[2])
-    else
-        e
-    end
-end
-
-expandargs(x) = x
-expandargs(q::QuoteNode) = Expr(:kw, q.value, q)
-function expandargs(e::Expr)
-    if e.head == :quote
-        Expr(:kw, e.args[1], e)
-    else
-        replace_equals_with_kw(e)
-    end
-end
-
 function select_helper(x, args...)
-    DF = gensym()
-    select_args = with_helper(DF, :($select($DF, $(map(expandargs, args)...))))
+    t = [fun_to_vec(arg) for arg in args]
+
     quote
-        let $DF = $x
-            $(with_helper(DF, :($select($DF, $(map(expandargs, args)...)))))
-        end
+        $DataFrames.select($x, $(t...))
     end
 end
 
--- a/src/DataFramesMeta.jl
+++ b/src/DataFramesMeta.jl
@@ -3,7 +3,7 @@ module DataFramesMeta
 using DataFrames, Tables
 
 # Basics:
-export @with, @where, @orderby, @transform, @by, @based_on, @select
+export @with, @where, @orderby, @transform, @by, @based_on, @select, @col, @row
 
 include("linqmacro.jl")
 include("byrow.jl")
@@ -43,6 +43,178 @@ replace_syms!(e::Expr, membernames) =
     else
         mapexpr(x -> replace_syms!(x, membernames), e)
     end
+
+"""
+    @col(kw)
+
+    
+`@col` transforms an expression of the form `z = :x + :y` into it's equivalent in 
+DataFrames's "mini-language". Functions act column-wise. For a row-wise functions, see 
+`@row`
+
+### Details
+
+Parsing follows the same convention as other DataFramesMeta macros, such as `@with`. All 
+terms in the expression that are `Symbols` are treated as columns in the DataFrame, except 
+`Symbol`s wrapped in `^`. To use a variable representing a column name, wrap the variable 
+in `cols`. 
+
+`@col` constructs an anonymous function based off the given expression. It then creates
+a `source => fun => destination` pair that is suitable for the `select`, `transform`, and 
+`combine` functions in DataFrames. 
+
+### Examples
+
+```
+julia> @col z = :x + :y
+[:x, :y] => (##595 => :z)
+```
+
+In the above example, `##595` is an anonymous function equivelent to the following 
+
+```
+(_x, _y) -> _x + _y
+```
+
+```
+julia> df = DataFrame(x = [1, 2], y = [3, 4]);
+
+julia> DataFrames.transform(df, @col z = :x .* :y)
+2×3 DataFrame
+│ Row │ x     │ y     │ z     │
+│     │ Int64 │ Int64 │ Int64 │
+├─────┼───────┼───────┼───────┤
+│ 1   │ 1     │ 3     │ 3     │
+│ 2   │ 2     │ 4     │ 8     │
+
+julia> DataFrames.transform(df, [:x, :y] => ((_x, _y) -> _x .* _y) => :z)
+2×3 DataFrame
+│ Row │ x     │ y     │ z     │
+│     │ Int64 │ Int64 │ Int64 │
+├─────┼───────┼───────┼───────┤
+│ 1   │ 1     │ 3     │ 3     │
+│ 2   │ 2     │ 4     │ 8     │
+
+```
+
+"""
+macro col(kw)
+    esc(fun_to_vec(kw; byrow = false))
+end
+
+"""
+    @row(kw)
+
+    
+`@row` transforms an expression of the form `z = :x + :y` into it's equivalent in 
+DataFrames's "mini-language". Functions act row-wise. For column-wise functions, see 
+`@col`
+
+### Details
+
+Parsing follows the same convention as other DataFramesMeta macros, such as `@with`. All 
+terms in the expression that are `Symbols` are treated as columns in the DataFrame, except 
+`Symbol`s wrapped in `^`. To use a variable representing a column name, wrap the variable 
+in `cols`. 
+
+`@row` constructs an anonymous function based off the given expression. It then creates
+a `source => fun => destination` pair that is suitable for the `select`, `transform`, and 
+`combine` functions in DataFrames. 
+
+### Examples
+
+```
+julia> @row z = :x + :y
+[:x, :y] => (ByRow{var"###607"}(##607) => :z)
+
+```
+
+In the above example, `##607` is an anonymous function equivelent to the following 
+
+```
+(_x, _y) -> _x + _y
+```
+
+The function is wrapped in `ByRow` indicating that DataFrames applies the function 
+to each row of the data frame.  
+
+```
+julia> df = DataFrame(x = [1, 2], y = [3, 4]);
+
+julia> DataFrames.transform(df, @row z = :x * :y)
+2×3 DataFrame
+│ Row │ x     │ y     │ z     │
+│     │ Int64 │ Int64 │ Int64 │
+├─────┼───────┼───────┼───────┤
+│ 1   │ 1     │ 3     │ 3     │
+│ 2   │ 2     │ 4     │ 8     │
+
+julia> DataFrames.transform(df, [:x, :y] => ByRow((_x, _y) -> _x * _y) => :z)
+2×3 DataFrame
+│ Row │ x     │ y     │ z     │
+│     │ Int64 │ Int64 │ Int64 │
+├─────┼───────┼───────┼───────┤
+│ 1   │ 1     │ 3     │ 3     │
+│ 2   │ 2     │ 4     │ 8     │
+"""
+macro row(kw)
+    esc(fun_to_vec(kw; byrow = true))
+end
+
+
+function fun_to_vec(kw::Expr; byrow = false)
+    # z = :x + :y
+    if kw.head == :(=) || kw.head == :kw
+        # New column to be created
+        output = kw.args[1]
+
+        # membernames:
+        # Dict(:x => _x, :y => _y)
+        #
+        # body:
+        # :(function(_x, _y) _x + _y end) 
+        membernames = Dict{Any, Symbol}()
+        funname = gensym()
+        body = replace_syms!(kw.args[2], membernames)
+
+        if byrow == false
+            # z = :x + :y
+            if kw.args[1] isa Symbol
+                t = quote
+                    $(Expr(:vect, keys(membernames)...)) => function $funname($(values(membernames)...))
+                        $body 
+                    end => $(QuoteNode(output))
+                end
+            # n = :z
+            # cols(n) = :x + :y
+            elseif DataFramesMeta.onearg(kw.args[1], :cols)
+                t = quote
+                    $(Expr(:vect, keys(membernames)...)) => function $funname($(values(membernames)...))
+                        $body 
+                    end => $(output.args[2])
+                end   
+            end
+        else 
+            if kw.args[1] isa Symbol
+                t = quote
+                    $(Expr(:vect, keys(membernames)...)) => $(ByRow)(function $funname($(values(membernames)...))
+                        $body 
+                    end) => $(QuoteNode(output))
+                end
+            elseif DataFramesMeta.onearg(kw.args[1], :cols)
+                t = quote
+                    $(Expr(:vect, keys(membernames)...)) => $(ByRow)(function $funname($(values(membernames)...))
+                        $body 
+                    end) => $(output.args[2])
+                end   
+            end
+        end
+        return t
+    else
+        Throw(ArgumentError("Invalid expression input"))
+    end
+end
+
 
 protect_replace_syms!(e, membernames) = e
 function protect_replace_syms!(e::Expr, membernames)
--- a/src/DataFramesMeta.jl
+++ b/src/DataFramesMeta.jl
@@ -461,10 +461,15 @@ end
 
 
 function transform_helper(x, args...)
+
     t = [fun_to_vec(arg) for arg in args]
 
     quote
-        $DataFrames.transform($x, $(t...))
+        out = $DataFrames.transform($x, $(t...))
+        if $x isa GroupedDataFrame
+            sort!(out, $groupcols($x))
+        end
+        out
     end
 end
 
--- a/src/DataFramesMeta.jl
+++ b/src/DataFramesMeta.jl
@@ -514,8 +514,8 @@ function based_on_helper(x, args...)
     if length(args) == 1 &&
         !(first(args) isa QuoteNode) &&
         !(first(args).head == :(=) || first(args).head == :kw)
+
         t = fun_to_vec(first(args); combinefun = true)
-        @show t
         quote
             $DataFrames.combine($t, $x)
         end
@@ -595,6 +595,7 @@ function by_helper(x, what, args...)
     if length(args) == 1 &&
         !(first(args) isa QuoteNode) &&
         !(first(args).head == :(=) || first(args).head == :kw)
+
         t = fun_to_vec(first(args); combinefun = true)
         quote
             $DataFrames.combine($t, $groupby($x, $what))
--- a/src/DataFramesMeta.jl
+++ b/src/DataFramesMeta.jl
@@ -3,7 +3,7 @@ module DataFramesMeta
 using DataFrames, Tables
 
 # Basics:
-export @with, @where, @orderby, @transform, @by, @based_on, @select, @col
+export @with, @where, @orderby, @transform, @by, @based_on, @select
 
 include("linqmacro.jl")
 include("byrow.jl")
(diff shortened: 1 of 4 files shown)


bkamins:
@pdeffebach - I would recommend to make docstrings explicit (i.e. they should give an exact specification what happens without referring to external references - of course such references can be retained but it is impossible to be 100% precise with them). Would this be possible? Thank you!

src/DataFramesMeta.jl
```diff
@@ -44,6 +44,178 @@ replace_syms!(e::Expr, membernames) =
...
+
+### Examples
+
+```
+julia> @col z = :x + :y
+[:x, :y] => (##595 => :z)
+```
+
+In the above example, `##595` is an anonymous function equivelent to the following 
+
```
nalimilan, line 74:
```suggestion
In the above example, `##595` is an anonymous function equivalent to the following:
```

src/DataFramesMeta.jl
```diff
@@ -44,6 +44,178 @@ replace_syms!(e::Expr, membernames) =
...
+julia> DataFrames.transform(df, @col z = :x .* :y)
+2×3 DataFrame
+│ Row │ x     │ y     │ z     │
+│     │ Int64 │ Int64 │ Int64 │
+├─────┼───────┼───────┼───────┤
+│ 1   │ 1     │ 3     │ 3     │
+│ 2   │ 2     │ 4     │ 8     │
+
+julia> DataFrames.transform(df, [:x, :y] => ((_x, _y) -> _x .* _y) => :z)
```
nalimilan, line 90: This one probably isn't needed here, I think everybody can see what the previous one means. :-)

src/DataFramesMeta.jl
```diff
@@ -44,6 +44,178 @@ replace_syms!(e::Expr, membernames) =
...
+### Examples
+
+```
+julia> @row z = :x + :y
+[:x, :y] => (ByRow{var"###607"}(##607) => :z)
+
+```
+
+In the above example, `##607` is an anonymous function equivelent to the following 
+
```
nalimilan, line 133:
```suggestion
In the above example, `##607` is an anonymous function equivalent to the following:

```

src/DataFramesMeta.jl
```diff
@@ -44,6 +44,178 @@ replace_syms!(e::Expr, membernames) =
...
+`Symbol`s wrapped in `^`. To use a variable representing a column name, wrap the variable 
+in `cols`. 
+
+`@col` constructs an anonymous function based off the given expression. It then creates
+a `source => fun => destination` pair that is suitable for the `select`, `transform`, and 
+`combine` functions in DataFrames. 
+
+### Examples
+
+```
```
nalimilan, line 68: Add `julia` or `jldoctest`?

src/DataFramesMeta.jl
```diff
@@ -44,6 +44,178 @@ replace_syms!(e::Expr, membernames) =
...
+│ 1   │ 1     │ 3     │ 3     │
+│ 2   │ 2     │ 4     │ 8     │
+"""
+macro row(kw)
+    esc(fun_to_vec(kw; byrow = true))
+end
+
+
+function fun_to_vec(kw::Expr; byrow = false)
+    # z = :x + :y
```
nalimilan, line 166: Explain that this is the example you use to illustrate?

src/DataFramesMeta.jl
```diff
@@ -44,6 +44,178 @@ replace_syms!(e::Expr, membernames) =
...
+                t = quote
+                    $(Expr(:vect, keys(membernames)...)) => function $funname($(values(membernames)...))
+                        $body 
+                    end => $(output.args[2])
+                end   
+            end
+        else 
+            if kw.args[1] isa Symbol
+                t = quote
+                    $(Expr(:vect, keys(membernames)...)) => $(ByRow)(function $funname($(values(membernames)...))
```
nalimilan, line 200: Since everything is the same except `ByRow`, would it be possible to do something like `identity` when `!byrow` so that code can be shared?

nalimilan:
Cool. Though maybe these shouldn't be exported until we have a clear idea of the API we want to provide? IIUC all macros will apply this automatically, right?

> * Many people have expressed interest in not forcing people to write `:x` and instead write `x` for `df.x`, similar to `dplyr` and Stata (`@transform(z = x + y)`). I support this, conditional on air-tight escaping rules. Should we start that transition now in the development of these macros?

I'd rather discuss this separately, unless it's tightly linked with the implementation. IIUC, once all the functions will have been implemented using these macros, it should be easier to switch the semantics for all of them at once.

> * Naming. Obviously `@col` is a less-than-ideal name given `Cols()` in DataAPI and `cols` for variable escaping. We should bikeshed over the name a bit.

Yeah. `@bycol` and `@byrow` would make sense to parallel `ByRow` in DataFrames, but we already have `@byrow`, so...

> * Multi-column selectors are broken for the time being, given you can't do `DataFrames.transform(df, [:x, [:y, :z]] => fun => :a)`. See discussions in [DataFrames](https://github.com/JuliaData/DataFrames.jl/issues/2328) and [DataFramesMeta](https://github.com/JuliaData/DataFramesMeta.jl/pull/157).

Would it still be possible to support `All`, `Between`, `Not` when they are used alone?

src/DataFramesMeta.jl
pdeffebach: > Cool. Though maybe these shouldn't be exported until we have a clear idea of the API we want to provide? IIUC all macros will apply this automatically, right?

Yes, `@transform` will just be `transform(df, @col ...)` etc. 

I would also probably add `@combine` and `@filter` eventually so there is a 1:1 parallel. 


> Yeah. `@bycol` and `@byrow` would make sense to parallel `ByRow` in DataFrames, but we already have `@byrow`, so...

We should rename `@byrow` to `@eachrow` imo. Since it's use and syntax follows super closely with `for row in eachrow(df)...`

> Would it still be possible to support `All`, `Between`, `Not` when they are used alone?

Yes, it would be possible. However this PR has the ability to be a totally non-breaking, no behavior-changing change to DataFramesMeta. So I guess I would rather merge this without that feature and ensure that everything is the same and then add this behavior in a new PR.

pdeffebach:
Sorry, accidentally responded to all comments in a thread. 

@nalimilan, @bkamins Initially I thought this PR would be implementing an opt-in feature that we could use to test things out before we went ahead and replaced `@transform` entirely. 

But if we don't export these macros, should we just go all the way in this PR and make the change? 

We could make this non-breaking with the exception of `cols([:a, :b])`. 

I added tests in #155 that should cover everything for a non-breaking release. Then we can discuss new features individually and add them piece by piece with the new infrastructure.

matthieugomez:
Just two comments
- I decide to name these macros @cols and @rows because it was a way to have a macro (which requires users to type more characters) that  adds some information. One issue is that the naming breaks down for `filter` since the pair syntax is by row by default there (hence my view that the Pair syntax should always be column wise).
- Knowing whether to use `:x` or `x` to refer to columns is indeed a separate question and it would be nice to have a separate issue about that. For me, the nice thing with `x` is that it is more natural as a keyword argument (as in `x=mean(y)`), and that  it lends itself to a very easy replacement syntax using `$` e.g `c =:x ; @cols($c = mean(x))`.

bkamins:
My two general comments:
* there should be someone who bears the burden of pushing how should things work (@pdeffebach - I understand you take this role in DataFramesMeta.jl and thank you for this) - I can give an opinion, but someone must have a grand vision of everything
* I think with DataFramesMeta.jl - since it is not very mature yet - it is OK to move forward faster than eg. DataFrames.jl with consecutive breaking changes to allow evolution into a good solution

Now given this PR, what problem I had is the docstring. Do I understand it correctly that:

`@cols a = expr` rewrites it to `[col1, col2, ...] => ((x...) -> expr2) => :a`

and

`@rows a = expr` rewrites it to `[col1, col2, ...] => ByRow(((x...) -> expr2)) => :a`

where `col1, col2, ...` are variables captured in `expr` and `expr2` is `expr` with these variables replaced by references to `x` argument of an anonymous function?

src/DataFramesMeta.jl
```diff
@@ -44,6 +44,167 @@ replace_syms!(e::Expr, membernames) =
...
+        # :(function(_x, _y) _x + _y end)
+        membernames = Dict{Any, Symbol}()
+        funname = gensym()
+        body = replace_syms!(kw.args[2], membernames)
+
+        if byrow == false
+            # z = :x + :y
+            if kw.args[1] isa Symbol
+                t = quote
+                    $(Expr(:vect, keys(membernames)...)) => function $funname($(values(membernames)...))
```
matthieugomez, line 173: You can just create an anonymous function which avoids gensys() completely (that’s what I do in my package)

src/DataFramesMeta.jl
```diff
@@ -44,6 +44,167 @@ replace_syms!(e::Expr, membernames) =
...
+                        $body
+                    end => $(output.args[2])
+                end
+            end
+        else
+            if kw.args[1] isa Symbol
+                t = quote
+                    $(Expr(:vect, keys(membernames)...)) => $(ByRow)(function $funname($(values(membernames)...))
```
matthieugomez, line 189: Since the whole expression ends up being wrapped in `esc`, does not that mean that the symbol =>, or ByRow should be prefixed by Base/DataFrames in case the user redefined them?

nalimilan:
> But if we don't export these macros, should we just go all the way in this PR and make the change?

@pdeffebach Yeah, why not if you feel like it's doable.

> One issue is that these names don’t work well for `filter` since the pair syntax is by row by default there (hence my view that the Pair syntax should always be column wise).

@matthieugomez That's a good point. There are two issues to discuss: 1) how we could support both row-wise and column-wise operation in `@filter`; and 2) whether the inconsistency (within DataFramesMeta and/or with DataFrames depending on what we do by default) is a problem. Probably better file a separate issue to avoid mixing discussions.

bkamins:
`@filter` can do whatever it likes (mostly), `filter` is in Base with a concrete contract (elementwise processing)

pdeffebach:
Okay this PR now has the goal of having zero-to-none breaking behavior. For the functions `@transform`, `@select`, `@by`, and `@based_on`, we use `transform`, `select`, and `combine`. 

I have made the change for each function. Remaining tasks are 

1. grouped `@transform`. This will have to be breaking as the current implementation calls `DataFrame(gd)` and thus reorders groups
2. Being very specific about what to do when `@based_on(gd, fun)`. 

I will ping for review when I am done with this and have added documentation about proposed behavior.

bkamins:
This is fantastic. I will review it once you ping it is good to look at.

pdeffebach:
Tests pass locally! This PR is ready for a review @nalimilan @bkamins 

No tests were harmed in the making of this PR. It either adds new tests or changes the error message in my "limits of" tests. I am confident that we can merge this PR and the user will have no breaking behavior with the exception of 

```
n = [:x1, :x2]
df = DataFrame(rand(2,2))
@transform(df, y = sum(eachcol(cols(n))))
```

Previously `cols(n)` would evaluate to a `DataFrame` but this now throws an error. I am under the impression this is undefined behavior so it's okay to break this. 

There is some weird code in here, in particular this annoying logic relating to `combinefun`

```
function fun_to_vec(kw::Expr; combinefun = false)
    if kw.head == :(=) || kw.head == :kw || combinefun == true
        funname = gensym()
        if combinefun == false
            body = replace_syms!(kw.args[2], membernames)
        else
            body = replace_syms!(kw, membernames)
        end
        if combinefun == false
            output = kw.args[1]
            t = quote
                $(Expr(:vect, keys(membernames)...)) =>
                (function $funname($(values(membernames)...))
                    $body
                end) =>
                $(QuoteNode(output))
            end
        else
            t = quote
                $(Expr(:vect, keys(membernames)...)) =>
                (function $funname($(values(membernames)...))
                    $body
                end)
            end
        end
        return t
    else
        throw(ArgumentError("Expressions not of the form `y = f(:x)` currently disallowed."))
    end
end
```

The problem this solves is that `@based_on` allows the user to write

```
@based_on(df, (n1 = :x1, n2 = :x2))
```

Here, `:x1` and `:x2` refer to the column in `df`, but the expression doesn't the normal keyword argument structure. I solve this with the code you see above. It's not elegant, but it's not breaking. 

I would rather keep some awkward code here for now than worry about all the tests, deprecations, and documentation associated with a breaking change. I will leave that for a slew of minor PRs after this one. 

Thanks!

src/DataFramesMeta.jl
pdeffebach: I believe so. This function is assigned a `gensym()` in the scope of the call.

src/DataFramesMeta.jl
pdeffebach: Thanks for the tip. I added this. Though I'm not 100% the implementation under the hood is different. I think an anonymous function calls `gensys` already.

test/grouping.jl
```diff
@@ -196,28 +204,26 @@ gd = groupby(df, :g)
...
-	@test  (@transform(g, y = :n .- median(:n)))[1,:y] == -5.0
+
+	@test (@transform(g, y = :n .- median(:n)))[1,:y] == -5.0
 
 	d = DataFrame(a = [1,1,1,2,2,3,3,1],
 	              b = Any[1,2,3,missing,missing,6.0,5.0,4],
 	              c = CategoricalArray([1,2,3,1,2,3,1,2]))
-	g = groupby(d, :a)
+
+    g = groupby(d, :a)
```
nalimilan, line 226:
```suggestion
      g = groupby(d, :a)
```

src/DataFramesMeta.jl
```diff
@@ -3,7 +3,7 @@ module DataFramesMeta
 using DataFrames, Tables
 
 # Basics:
-export @with, @where, @orderby, @transform, @by, @based_on, @select
+export @with, @where, @orderby, @transform, @by, @based_on, @select, @col
```
nalimilan, line 6: Let's not export this for now? It's not clear whether it will be needed or not in the final API.

nalimilan (commented):
Sounds great. I just have a few minor comments.

test/dataframes.jl
```diff
@@ -88,37 +91,39 @@ s = [:i, :g]
...
-    @test_throws ArgumentError @eval @transform(df, Not([:i, :g]))
+    @test_throws LoadError @eval @transform(df, Between(:i, :t)).Between == df.i
+    @test_throws LoadError @eval  @transform(df, Not(:i)).Not == df.i
+    @test_throws LoadError @eval @transform(df, Not([:i, :g]))
     newvar = :n
-    @test_throws ErrorException @eval @transform(df, cols(newvar) = :i)
+    @test_throws ArgumentError @eval @transform(df, cols(newvar) = :i)
     @test_throws MethodError @eval @transform(df, n = sum(Between(:i, :t)))
-    @test @transform(df, n = :i).n === df.i
-    @test @transform(df, n = cols(s).i).n == df.i
```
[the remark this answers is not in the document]
pdeffebach: yes, buggy or undefined behavior. In the above example, `cols(s)` is a DataFrame. This PR disallows this.

src/DataFramesMeta.jl
```diff
@@ -519,9 +511,20 @@ end
 ##############################################################################
 
 function based_on_helper(x, args...)
-    with_args =
-        with_anonymous(:($DataFrame($(map(replace_equals_with_kw, args)...))))
-    :(DataFrames.combine($with_args, $x))
+    if length(args) == 1 &&
```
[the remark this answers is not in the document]
pdeffebach: We only allow one argument when people want to output a table object, similar to `combine`. In particular, with current `combine` the table output needs to go in the first argument, and we can't ensure that if there is more than one argument.

src/DataFramesMeta.jl
```diff
@@ -373,106 +459,12 @@ end
...
-             t = copyto!(Tables.allocatecolumn(newtype, length(t)),
-                         1, t, 1, ends[i-1])
-             _transform!(t, out, i, g, v, starts, ends)
-         end
-    end
-    return t
-end
 
 function transform_helper(x, args...)
+    t = [fun_to_vec(arg) for arg in args]
```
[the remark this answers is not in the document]
pdeffebach: added a generator.

src/DataFramesMeta.jl
```diff
@@ -44,6 +44,92 @@ replace_syms!(e::Expr, membernames) =
...
+"""
+macro col(kw)
+    esc(fun_to_vec(kw))
+end
+
+function fun_to_vec(kw::Expr; combinefun = false)
+    if kw.head == :(=) || kw.head == :kw || combinefun == true
+        membernames = Dict{Any, Symbol}()
+        funname = gensym()
+        if combinefun == false
```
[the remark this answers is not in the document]
pdeffebach: fixed!

src/DataFramesMeta.jl
pdeffebach: Okay not exported. Keeping it unexported though as the docstring is informative.

pdeffebach:
Thanks!

test/dataframes.jl
```diff
@@ -88,37 +91,39 @@ s = [:i, :g]
 
 @testset "limits of @transform" begin
     ## Test for not-implemented or strange behavior
-    @test_throws LoadError @eval @transform(df, :i)
-    @test_throws ErrorException @eval @transform(df, [:i, :g])
+    # @test_throws throws a `LoadError` when it
```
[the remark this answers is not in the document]
pdeffebach: This is a limitation of `@test_throws`. I added a comment and filed an issue [here](https://github.com/JuliaLang/julia/issues/37420).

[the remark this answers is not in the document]
pdeffebach: Update: It's because some errors happen when the macro expansion happens and others happen inside the functions.

src/DataFramesMeta.jl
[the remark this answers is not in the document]
nalimilan: OK, then I'd say it's worth adding a comment to explain this. Also `combinefun` may not be the best name for this, as `combine` accepts several forms. What's special AFAICT is that there's no output name provided, right?

src/DataFramesMeta.jl
```diff
@@ -44,6 +44,86 @@ replace_syms!(e::Expr, membernames) =
...
+
+```
+
+"""
+macro col(kw)
+    esc(fun_to_vec(kw))
+end
+
+function fun_to_vec(kw::Expr; combinefun = false)
+    if kw.head == :(=) || kw.head == :kw || combinefun == true
```
nalimilan, line 97:
```suggestion
    if kw.head === :(=) || kw.head === :kw || combinefun
```

test/grouping.jl
```diff
@@ -70,15 +70,20 @@ g = groupby(d, :x, sort=true)
     @test @based_on(gd, n = first(cols(tr) .* string.(cols(yr)))).n == ["av", "cy"]
     @test @based_on(gd, n = first(Symbol.(cols(yr), ^(:t)))).n == [:vt, :yt]
     @test @based_on(gd, n = first(Symbol.(cols(yr), ^(:body)))).n == [:vbody, :ybody]
-    @test @based_on(gd, body = cols(ir)).body == df.i 
+    @test @based_on(gd, body = cols(ir)).body == df.i
     @test @based_on(gd, transform = cols(ir)).transform == df.i
     @test @based_on(gd, (n1 = [first(cols(ir))], n2 = [first(cols(yr))])).n1 == [1, 4]
+
+    @test @based_on(gd, :i).i == df.i
+    @test @based_on(gd,:i, :g).g == df.g
```
nalimilan, line 78:
```suggestion
    @test @based_on(gd, :i, :g).g == df.g
```

test/grouping.jl
```diff
@@ -103,12 +106,10 @@ gd = groupby(df, :g)
 newvar = :n
 
 @testset "Limits of @based_on" begin
-    @test_throws LoadError @eval @based_on(gd, :i)
-    @test @based_on(gd, [:i, :g]) == DataFrame(g = df.g, x1 = df.i, x2 = df.g)
-    @test_throws ArgumentError @eval @based_on(gd, All())
-    @test_throws MethodError @eval @based_on(gd, Between(:i, :t)).Between == df.i
-    @test_throws ArgumentError @eval @based_on(gd, Not(:i)).Not == df.i
-    @test_throws ArgumentError @eval @based_on(gd, Not([:i, :g]))
+    @test @based_on(gd, [:i, :g]).i_g_function isa Vector{<:SubArray}
```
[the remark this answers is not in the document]
nalimilan: Still applies.

test/grouping.jl
```diff
@@ -163,15 +164,18 @@ end
     @test @by(df, "g", n = first(cols(tr) .* string.(cols(yr)))).n == ["av", "cy"]
     @test @by(df, "g", n = first(Symbol.(cols(yr), ^(:t)))).n == [:vt, :yt]
     @test @by(df, "g", n = first(Symbol.(cols(yr), ^(:body)))).n == [:vbody, :ybody]
-    @test @by(df, "g", body = cols(ir)).body == df.i 
+    @test @by(df, "g", body = cols(ir)).body == df.i
     @test @by(df, "g", transform = cols(ir)).transform == df.i
     @test @by(df, "g", (n1 = [first(cols(ir))], n2 = [first(cols(yr))])).n1 == [1, 4]
+
+    @test @by(df, :g, :i).i == df.i
+    @test @by(df, :g, :i, :g).g == df.g
```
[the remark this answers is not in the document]
nalimilan: Still applies too.

test/grouping.jl
```diff
@@ -196,22 +200,19 @@ gd = groupby(df, :g)
...
-	@test @by(df, :g, [:i, :g]) ≅ DataFrame(g = df.g, x1 = df.i, x2 = df.g)
-	@test_throws ArgumentError @eval @by(df, :g, All())
-	@test_throws MethodError @eval @by(df, :g, Between(:i, :t)).Between == df.i
-	@test_throws ArgumentError @eval @by(df, :g, Not(:i)).Not == df.i
-	@test_throws ArgumentError @eval @by(df, :g, Not([:i, :g]))
-	newvar = :n
-	@test_throws ArgumentError @eval @by(df, :g, cols(newvar) = mean(:i))
-	@test_throws MethodError @eval @by(df, :g, n = sum(Between(:i, :t)))
-	@test_throws LoadError @eval @by(df, :g; n = mean(:i))
+    @test @by(df, :g, [:i, :g]).i_g_function isa Vector{<:SubArray}
```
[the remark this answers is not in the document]
nalimilan: This one too.

test/grouping.jl
```diff
@@ -196,29 +204,27 @@ gd = groupby(df, :g)
...
+	@test (@transform(g, y = :n .- median(:n)))[1,:y] == -5.0
 
 	d = DataFrame(a = [1,1,1,2,2,3,3,1],
 	              b = Any[1,2,3,missing,missing,6.0,5.0,4],
 	              c = CategoricalArray([1,2,3,1,2,3,1,2]))
-	g = groupby(d, :a)
-	## Scalar output
+    g = groupby(d, :a)
+
+    ## Scalar output
```
nalimilan, line 227:
```suggestion
      g = groupby(d, :a)

      ## Scalar output
```

src/DataFramesMeta.jl
```diff
@@ -44,6 +44,86 @@ replace_syms!(e::Expr, membernames) =
         mapexpr(x -> replace_syms!(x, membernames), e)
     end
 
+"""
+    @col(kw)
+
+`@col` transforms an expression of the form `z = :x + :y` into it's equivalent in
+DataFrames's "mini-language".
```
nalimilan, line 51: "mini-language" is probably too strong... :-)
```suggestion
DataFrames's `source => fun => destination` syntax.
```

src/DataFramesMeta.jl
```diff
@@ -589,8 +591,21 @@ end
 ##############################################################################
 
 function by_helper(x, what, args...)
-    :(DataFrames.combine($(with_anonymous(:($DataFrame($(map(replace_equals_with_kw, args)...))))),
-              DataFrames.groupby($x, $what)))
+    # Only alow one argument when returning a Table object
```
nalimilan, line 594:
```suggestion
    # Only allow one argument when returning a Table object
```

src/DataFramesMeta.jl
```diff
@@ -519,9 +509,21 @@ end
 ##############################################################################
 
 function based_on_helper(x, args...)
-    with_args =
-        with_anonymous(:($DataFrame($(map(replace_equals_with_kw, args)...))))
-    :(DataFrames.combine($with_args, $x))
+    # Only alow one argument when returning a Table object
```
nalimilan, line 512:
```suggestion
    # Only allow one argument when returning a Table object
```

src/DataFramesMeta.jl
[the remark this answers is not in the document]
pdeffebach: Yeah what's special is that we don't have `y = f(:x)` but still want `:x` to represent the vector `df.x`. So we want to apply `replace_syms!` but not create a new column the normal way. 

This will definitely get re-worked in the future, but probably in a breaking way with some new syntax. This flag is just for the non-breaking-ness of the PR. 

I think I will rename it to `combinetable`.

test/grouping.jl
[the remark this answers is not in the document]
pdeffebach: This is in "limits" because its not behavior anyone expects and I don't want to consider this behavior part of the public API. (Hence "limits" instead of "errors"). Is that feasible?

Will still add a real test for equality and type

src/DataFramesMeta.jl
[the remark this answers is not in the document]
nalimilan: OK I think I still don't really get what happens. :-p

test/grouping.jl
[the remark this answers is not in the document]
nalimilan: OK. Though we should quickly turn this either into an error or into supported behavior.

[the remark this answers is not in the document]
pdeffebach: Yes. Tentatively, I propose the following rule

1. All expressions of the form `y = f(:x)` get the macro applied to them
2. All expressions that *don't* have that structure get evaluated as if they were in the basic DataFrames pipes, this is so `@select(df, Between(:a, :b))` works. 
3. If users want to have an expression that is *not* of the form `y = f(:x)` but still want to refer to `:x` as the column `df.x`, which only applies in `combine`. Then they have to do 

```
@based_on(df, @fun f(:x))
```

So if that is implemented 

```
@based_on(gd, [:i, :g])
```

would be the exact same as `combine(df, [:i, :g])` and if they really wanted a vector of the `SubArray`s they would have to do

```
@based_on(gd, @fun [:i, :g])
```

This is a very tentative proposal but would realistically get around this problem.

nalimilan (approved):
Looks good (at least for the parts I master)!

bkamins:
OK - I will try to review it today or tomorrow then.

test/grouping.jl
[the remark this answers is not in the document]
nalimilan: I'm not a big fan of special-casing `y = f(:x)`. What if we required people to write `All([:i, :g])` (or `Cols([i, :g])` if we add this) when they want to get these columns? Anyway you can always write `based_on(df, :i, :g)`, which is actually shorter.

(BTW, `@based_on` should really be deprecated in favor of `@combine` for consistency with DataFrames. I've never liked that name anyway.)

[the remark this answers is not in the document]
pdeffebach: I disagree. In DataFrames

```
combine(df, [:a, :b], :i, [:x, :y] => fun => :z 
```

The "selectors" mean different things than the `[:x, :y]` in the `Pair`. The `[:a, :b]` mean "keep these columns", whereas the `[:x, :y]` area things that get passed to a function. 

I think special casing `y = f(:x)` is consistent with the way `:x`, or inputs to `Pair`s more generally, is used in base data frames. 

More importantly, though, people want `@select(df, :a, y = f(:x))` to work and special casing `y = f(:x)` makes that feasible. 

Obviously we could `@select(df, :a)` passed into the whole split apply combine infrastructure, but then we would end up with `:a_fun` and have to worry about that.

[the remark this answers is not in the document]
nalimilan: OK. But then it's not really `y = f(x)` that would be special-cased, but just the fact that a keyword argument is passed (equivalent to a pair in DataFrames).

One issue though is that in DataFrames the pair syntax allows quite naturally to omit the output names in the case where you return a named tuple. Requiring `@fun` in DataFramesMeta wouldn't be as nice. That's not high priority, but it would be interesting to find another solution.

src/DataFramesMeta.jl
```diff
@@ -44,6 +44,90 @@ replace_syms!(e::Expr, membernames) =
         mapexpr(x -> replace_syms!(x, membernames), e)
     end
 
+"""
+    @col(kw)
+
+`@col` transforms an expression of the form `z = :x + :y` into it's equivalent in
```
bkamins, line 50: maybe add `[:x, :y] => ((x,y) -> +(x,y)) => :z` (or whatever it exactly is) and then say that it is DataFrames.jl syntax as you do later in this sentence.

src/DataFramesMeta.jl
```diff
@@ -44,6 +44,90 @@ replace_syms!(e::Expr, membernames) =
...
+DataFrames's `source => fun => destination` syntax.
+
+### Details
+
+Parsing follows the same convention as other DataFramesMeta macros, such as `@with`. All
```
bkamins, line 55:
```suggestion
Parsing follows the same convention as other DataFramesMeta.jl macros, such as `@with`. All
```

src/DataFramesMeta.jl
```diff
@@ -44,6 +44,90 @@ replace_syms!(e::Expr, membernames) =
...
+terms in the expression that are `Symbols` are treated as columns in the data frame, except
```
bkamins, line 56:
```suggestion
terms in the expression that are `Symbol`s are treated as columns in the data frame, except
```

src/DataFramesMeta.jl
```diff
@@ -44,6 +44,90 @@ replace_syms!(e::Expr, membernames) =
...
+`Symbol`s wrapped in `^`. To use a variable representing a column name, wrap the variable
+in `cols`.
+
+`@col` constructs an anonymous function based off the given expression. It then creates
```
bkamins, line 60:
```suggestion
`@col` constructs an anonymous function based on the given expression. It then creates
```

src/DataFramesMeta.jl
```diff
@@ -44,6 +44,90 @@ replace_syms!(e::Expr, membernames) =
...
+a `source => fun => destination` pair that is suitable for the `select`, `transform`, and
+`combine` functions in DataFrames.
```
bkamins, line 62:
```suggestion
`combine` functions in DataFrames.jl.
```

src/DataFramesMeta.jl
bkamins: oh - I see that you have added in the example. Then decide if to duplicate it or leave as is.

src/DataFramesMeta.jl
```diff
@@ -44,6 +44,90 @@ replace_syms!(e::Expr, membernames) =
...
+macro col(kw)
+    esc(fun_to_vec(kw))
+end
+
+# `combinetable` needs to be `true` when we have syntax of the form
+# `@based_on(gd, fun(:x, :y))` where `fun` returns a `table` object.
+# We don't create the "new name" pair because new names are given
+# by the table.
+function fun_to_vec(kw::Expr; combinetable = false)
+    if kw.head === :(=) || kw.head === :kw || combinetable
```
bkamins, line 101: what is the scenario for `:kw`?

src/DataFramesMeta.jl
```diff
@@ -44,6 +44,90 @@ replace_syms!(e::Expr, membernames) =
...
+            output = kw.args[1]
+            t = quote
+                $(Expr(:vect, keys(membernames)...)) =>
+                ($(Expr(:tuple, values(membernames)...)) -> $body) =>
+                $(QuoteNode(output))
+            end
+        end
+        return t
+    else
+        throw(ArgumentError("Expressions not of the form `y = f(:x)` currently disallowed."))
```
bkamins, line 124: I am not clear here. `:(y = f(:x))` has `head` equal to `:(=)`

bkamins:
As a general question: it seems that mostly we replace back-end, but frontend is unchanged (i.e. the tests are the same). Are there new things that are possible now in comparison to what was possible earlier?

tk3369:
@pdeffebach I'm happy to assist in alpha/beta testing when you're ready.

src/DataFramesMeta.jl
pdeffebach: Leaving as is.

src/DataFramesMeta.jl
pdeffebach: With a small tweak in the internals of `@byrow` we don't need to consider `:kw`. So we don't have to support it.

src/DataFramesMeta.jl
pdeffebach: This refers to disallowing thing like `@select(df, [:x2, :x4], Between(:x1, :x5))`. This is currently not allowed in `@select` so I didn't want to introduce that feature in this PR.

src/DataFramesMeta.jl
pdeffebach: Nevermind. The `:kw` path is actually important here for `@linq` to work. I am not sure why, however.

pdeffebach:
> As a general question: it seems that mostly we replace back-end, but frontend is unchanged (i.e. the tests are the same). Are there new things that are possible now in comparison to what was possible earlier?

The only thing I can think of is when we "spread" a scalar output across columns, i.e. `@transform(df, y = 5)` does not work, but will work with this PR. 

This has been added in tests for all 4 functions, `@transform`, `@select`, `@based_on`, and `@by`.

src/DataFramesMeta.jl
pdeffebach: See #169 https://github.com/JuliaData/DataFramesMeta.jl/issues/169

pdeffebach:
Just did a quick once-over. I think this is ready to be merged!

bkamins:
Thank you! Let us expose it on master to the users (so that we can potentially get a feedback before a release).

nalimilan:
Cool! Now we can start thinking about a deeper redesign!

PR #171, opened 2020-09-23
Allow cols on LHS

pdeffebach:
This will enable 

```
newname = :y 
df = DataFrame(x1 = [1], x2 = [2])
@transform(df, cols(newname) = :x1 + :x2)
```

This is a relatively uncontroversial and easy to implement feature! Tests will be added soon.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,10 +127,10 @@ end
 ## Working with column names programmatically with `cols`
 
 DataFramesMeta provides the special syntax, `cols`, for referring to 
-columns in a DataFrame via a `Symbol`, `String`, or column position. 
+columns in a data frame via a `Symbol`, string, or column position as either
+a literal or a variable. 
 
 ```julia
-
 df = DataFrame(A = 1:3, B = [2, 1, 2])
 
 nameA = :A
@@ -140,10 +140,9 @@ nameB = "B"
 df3 = @byrow df begin 
     :A = cols(nameB)
 end
-
 ```
 
-`cols` can also be used to create new columns in a DataFrame. 
+`cols` can also be used to create new columns in a data frame. 
 
 ```julia
 df = DataFrame(A = 1:3, B = [2, 1, 2])
@@ -158,11 +157,11 @@ df3 = @byrow df begin
     @newcol cols(nameC)::Vector{Int}
     cols(nameC) = :A
 end
-
 ```
 
 Note that `cols` is *not* a standard Julia function. It is only used to modify the 
-way that macros in DataFramesMeta escape arguments and has no behavior of its own. 
+way that macros in DataFramesMeta escape arguments and has no behavior of its own 
+outside of DataFramesMeta macros.
 
 
 ## LINQ-Style Queries and Transforms
--- a/README.md
+++ b/README.md
@@ -38,8 +38,6 @@ end
 
 @with(df, df[:x .> 1, ^(:y)]) # The ^ means leave the :y alone
 
-colref = :x
-@with(df, :y + cols(colref)) # Equivalent to df[:y] + df[colref]
 ```
 
 `@with` is the fundamental macro used by the other metaprogramming
@@ -125,6 +123,40 @@ df2 = @byrow df begin
     end
 end
 ```
+
+## Working with column names programmatically with `cols`
+
+DataFramesMeta provides the special syntax, `cols`, for referring to 
+columns in a DataFrame via a `Symbol`, `String`, or column position. 
+
+```julia
+
+df = DataFrame(A = 1:3, B = [2, 1, 2])
+
+nameA = :A
+
+df2 = @transform(df, C = :B - cols(nameA))
+
+nameB = "B"
+df3 = @byrow df begin 
+    :A = cols(nameB)
+end
+
+```
+
+`cols` can also be used to create new columns in a DataFrame. 
+
+```julia
+df = DataFrame(A = 1:3, B = [2, 1, 2])
+
+newcol = "C"
+@select(df, cols(newcol) = :A + :B)
+
+@by(df, :B, cols("A complicated"  * " new name") = first(:A))
+
+```
+
+
 
 ## LINQ-Style Queries and Transforms
 
--- a/README.md
+++ b/README.md
@@ -134,7 +134,6 @@ columns in a DataFrame via a `Symbol`, `String`, or column position.
 df = DataFrame(A = 1:3, B = [2, 1, 2])
 
 nameA = :A
-
 df2 = @transform(df, C = :B - cols(nameA))
 
 nameB = "B"
@@ -164,7 +163,6 @@ end
 
 Note that `cols` is *not* a standard Julia function. It is only used to modify the 
 way that macros in DataFramesMeta escape arguments and has no behavior of its own. 
-
 
 
 ## LINQ-Style Queries and Transforms
--- a/README.md
+++ b/README.md
@@ -154,7 +154,16 @@ newcol = "C"
 
 @by(df, :B, cols("A complicated"  * " new name") = first(:A))
 
+nameC = "C"
+df3 = @byrow df begin 
+    @newcol cols(nameC)::Vector{Int}
+    cols(nameC) = :A
+end
+
 ```
+
+Note that `cols` is *not* a standard Julia function. It is only used to modify the 
+way that macros in DataFramesMeta escape arguments and has no behavior of its own. 
 
 
 
--- a/README.md
+++ b/README.md
@@ -126,7 +126,7 @@ end
 
 ## Working with column names programmatically with `cols`
 
-DataFramesMeta.jl provides the special syntax, `cols`, for referring to 
+DataFramesMeta.jl provides the special syntax `cols` for referring to 
 columns in a data frame via a `Symbol`, string, or column position as either
 a literal or a variable. 
 
--- a/src/DataFramesMeta.jl
+++ b/src/DataFramesMeta.jl
@@ -99,23 +99,30 @@ end
 # We don't create the "new name" pair because new names are given
 # by the table.
 function fun_to_vec(kw::Expr; nolhs = false)
+    # nolhs: f(:x) where f returns a Table
+    # !nolhs, y = g(:x)
     if kw.head === :(=) || kw.head === :kw || nolhs
         membernames = Dict{Any, Symbol}()
         if nolhs
+            # act on f(:x)
             body = replace_syms!(kw, membernames)
         else
+            # act on g(:x)
             body = replace_syms!(kw.args[2], membernames)
         end
 
         if nolhs
+            # [:x] => _f
             t = quote
                 $(Expr(:vect, keys(membernames)...)) =>
                 ($(Expr(:tuple, values(membernames)...)) -> $body)
             end
          else
             if kw.args[1] isa Symbol
+                # cols(n) = f(:x) becomes [:x] => _f => n
                 output = QuoteNode(kw.args[1])
             elseif onearg(kw.args[1], :cols)
+                # y = f(:x) becomes [:x] => _f => :y
                 output = kw.args[1].args[2]
             end
             t = quote
--- a/src/DataFramesMeta.jl
+++ b/src/DataFramesMeta.jl
@@ -23,6 +23,7 @@ function addkey!(membernames, nam)
 end
 
 onearg(e, f) = e.head == :call && length(e.args) == 2 && e.args[1] == f
+
 mapexpr(f, e) = Expr(e.head, map(f, e.args)...)
 
 replace_syms!(x, membernames) = x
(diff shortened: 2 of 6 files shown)


bkamins (approved):

src/DataFramesMeta.jl
```diff
@@ -112,11 +112,15 @@ function fun_to_vec(kw::Expr; nolhs = false)
                 ($(Expr(:tuple, values(membernames)...)) -> $body)
             end
          else
-            output = kw.args[1]
+            if onearg(kw.args[1], :cols)
+                output = kw.args[1].args[2]
```
bkamins, line 116: is `QuoteNode` not needed here?

pdeffebach: We wrap in `QuoteNode` so that when it gets escaped it is treated as a Symbol. 

When a `Symbol` *not* wrapped in `QuotNode` is escaped, it looks for the variable in the current environment, which is what we want because the DataFrames expression is evaluated normally.

pdeffebach:
Still need 

1. Update readme
2. Same in `@byrow`
3. `@linq` tests

bkamins:
@pdeffebach - thank you for working on this. When you are finished with "internal design" it would be super helpful if you put somewhere a short note how things work now on implementation level (it can be even in a comment).

pdeffebach:
The relevant implementation details are in two places

First, for `@transform` etc. 

```
            if kw.args[1] isa Symbol
                output = QuoteNode(kw.args[1])
            elseif onearg(kw.args[1], :cols)
                output = kw.args[1].args[2]
            end
            t = quote
                $(Expr(:vect, keys(membernames)...)) =>
                ($(Expr(:tuple, values(membernames)...)) -> $body) =>
                $(output)
            end
```

In the above snippet, `kw` is the expression `y = f(:x)` or `cols(n) = f(:x)`

If the expression if `y = f(x)`, then it looks like 

```
julia> dump(:(y = f(:x)))
Expr
  head: Symbol =
  args: Array{Any}((2,))
    1: Symbol y
    2: Expr
      head: Symbol call
      args: Array{Any}((2,))
        1: Symbol f
        2: QuoteNode
          value: Symbol x
```

We see that `kw.args[1]` is a `Symbol`. In the expression, it is a literal. Then we wrap `y` in a `QuoteNode`, which takes it from `y` to `:y` and we get 

```
[:x] => _fun => :y 
```

If it is of the form `cols(n) = f(x)` then we have 

```
julia> dump(:(cols(n) = f(:x)))
Expr
  head: Symbol =
  args: Array{Any}((2,))
    1: Expr
      head: Symbol call
      args: Array{Any}((2,))
        1: Symbol cols
        2: Symbol n
    2: Expr
      head: Symbol block
      args: Array{Any}((2,))
        1: LineNumberNode
          line: Int64 1
          file: Symbol REPL[23]
        2: Expr
          head: Symbol call
          args: Array{Any}((2,))
            1: Symbol f
            2: QuoteNode
              value: Symbol x
```

Clearly `kw.args[1]` is an `Expr`. So we check if it's `onearg(kw.args[1], :cols)`, meaning it's of the form `cols(n)` and not `cols(n, p)`. 

When we are in this branch, we pluck the `n` out of that expression and leave it as a `Symbol`, so it gets evaluated like a normal literal would in the DataFrames call. This gives us 

```
[:x] => _fun => n
```

And whatever `n` is is the new name.

pdeffebach:
With `@byrow` the fix was even easier. I simply had to fix something I wrote in #157 . When we see 

```
cols(n) = :y 
```

we simply have to replace that expression with 

```
cols(n)[i] = :x[i]
```

and `@with` handles the rest via `replace_syms!`. 

Note that in order for this to work, I had to deprecate using `cols([:x, :y])` with `@with`. Previously it worked with `@with` (returning a `DataFrame`), but not `@transform` etc. and `@byrow`. 

Hopefully I will be able to add `cols([:x, :y])` return something more sensible soon. This is an acceptable deprecation because we are deprecating it elsewhere already.

pdeffebach:
Note: I still need to add this for `@newcol` macro in `@byrow`.

src/DataFramesMeta.jl
```diff
@@ -112,11 +113,15 @@ function fun_to_vec(kw::Expr; nolhs = false)
                 ($(Expr(:tuple, values(membernames)...)) -> $body)
             end
          else
-            output = kw.args[1]
+            if kw.args[1] isa Symbol
+                output = QuoteNode(kw.args[1])
+            elseif onearg(kw.args[1], :cols)
+                output = kw.args[1].args[2]
+            end
```
xiaodaigh, line 120: Some comments here about what each branch is handling would be nice.

src/DataFramesMeta.jl
```diff
@@ -143,6 +148,10 @@ function replace_dotted!(e, membernames)
     Expr(:., x_new, y_new)
 end
 
+getsinglecolumn(df, s::DataFrames.ColumnIndex) = df[!, s]
```
pdeffebach, line 158: @bkamins This is what I mean when I say that we deprecate `cols([:x, :y])` in side `@with`. 

I define a small helper function for getting a single column from the data frame. Below, I use this function when getting columns to the anonymous function created.

src/DataFramesMeta.jl
pdeffebach: added, thanks!

pdeffebach:
Okay I think this looks good! requesting a review fom @nalimilan and @bkamins and maybe @quinnj

pdeffebach:
Update: This does not work with `@linq`. Will work on a fix.

pdeffebach:
Update: `@linq` was not a problem, mistakes writing tests was. 

est pass. This is ready for a review.

README.md
```diff
@@ -126,6 +124,47 @@ df2 = @byrow df begin
 end
 ```
 
+## Working with column names programmatically with `cols`
+
+DataFramesMeta provides the special syntax, `cols`, for referring to 
+columns in a DataFrame via a `Symbol`, `String`, or column position.
```
bkamins, line 130: maybe replace `DataFrame` everywhere in the docs by "data frame" as you do not require a `DataFrame` but just `AbstractDataFrame` - right?

bkamins: same with `String` better write "string" as any string is allowed.

README.md
```diff
@@ -126,6 +124,47 @@ df2 = @byrow df begin
...
+df = DataFrame(A = 1:3, B = [2, 1, 2])
+
+nameA = :A
+df2 = @transform(df, C = :B - cols(nameA))
+
+nameB = "B"
+df3 = @byrow df begin 
+    :A = cols(nameB)
+end
+
```
bkamins, line 143: here and below this empty line I think can be removed at the end of code block

README.md
```diff
@@ -126,6 +124,47 @@ df2 = @byrow df begin
...
+`cols` can also be used to create new columns in a DataFrame. 
+
+```julia
+df = DataFrame(A = 1:3, B = [2, 1, 2])
+
+newcol = "C"
+@select(df, cols(newcol) = :A + :B)
+
+@by(df, :B, cols("A complicated"  * " new name") = first(:A))
```
bkamins, line 154:
```suggestion
@by(df, :B, cols("A complicated" * " new name") = first(:A))
```

README.md
```diff
@@ -126,6 +124,47 @@ df2 = @byrow df begin
...
+nameC = "C"
+df3 = @byrow df begin 
+    @newcol cols(nameC)::Vector{Int}
+    cols(nameC) = :A
+end
+
+```
+
+Note that `cols` is *not* a standard Julia function. It is only used to modify the 
+way that macros in DataFramesMeta escape arguments and has no behavior of its own.
```
bkamins, line 165:
```suggestion
way that macros in DataFramesMeta.jl escape arguments and has no behavior of its own. 
```

src/DataFramesMeta.jl
```diff
@@ -143,6 +155,10 @@ function replace_dotted!(e, membernames)
...
+getsinglecolumn(df, s) = throw(ArgumentError("Only indexing with Symbols and Strings" *
```
bkamins, line 159:
```suggestion
getsinglecolumn(df, s) = throw(ArgumentError("Only indexing with Symbols and strings" *
```

bkamins: also integers - right?

bkamins (approved):
Nice. Actually `cols` works quite nicely with strings.

README.md
```diff
@@ -126,6 +124,47 @@ df2 = @byrow df begin
 end
 ```
 
+## Working with column names programmatically with `cols`
+
+DataFramesMeta provides the special syntax, `cols`, for referring to
```
nalimilan, line 129:
```suggestion
DataFramesMeta provides the special syntax `cols`, for referring to 
```

README.md
```diff
@@ -126,6 +124,47 @@ df2 = @byrow df begin
 end
 ```
 
+## Working with column names programmatically with `cols`
+
+DataFramesMeta provides the special syntax, `cols`, for referring to 
+columns in a DataFrame via a `Symbol`, `String`, or column position. 
+
+```julia
+
```
nalimilan, line 133:
```suggestion
```

README.md
nalimilan: Also it should be mentioned that `cols` allows passing a variable holding the index, right?

src/DataFramesMeta.jl
```diff
@@ -143,6 +155,10 @@ function replace_dotted!(e, membernames)
...
+    "Is currently allowed with cols"))
```
nalimilan, line 160:
```suggestion
    "is currently allowed with cols"))
```
Also align with above string.

test/dataframes.jl
```diff
@@ -168,6 +180,13 @@ end
     @test DataFramesMeta.select(df, :i) == df.i
 
     @test @select(df, n = 1).n == fill(1, nrow(df))
+
+    @test @select(df, cols("new_column") = :i).new_column == df.i
+    @test @select(df, cols(n_str) = :i).new_column == df.i
+    @test @select(df, cols(n_sym) = :i).new_column == df.i
+    @test @select(df, cols(n_space) = :i)."new column" == df.i
+    @test @select(df, cols("new" * "_" * "column") = :i).new_column == df.i
+
```
nalimilan, line 189:
```suggestion
```

test/grouping.jl
```diff
@@ -78,6 +80,13 @@ g = groupby(d, :x, sort=true)
     @test @based_on(gd, :i, :g) ≅ select(df, :g, :i)
 
     @test @based_on(gd, :i, n = 1).n == fill(1, nrow(df))
+
+    @test @based_on(gd, cols("new_column") = 2).new_column == [2, 2]
+    @test @based_on(gd, cols(n_str) = 2).new_column == [2, 2]
+    @test @based_on(gd, cols(n_sym) = 2).new_column == [2, 2]
+    @test @based_on(gd, cols(n_space) = 2)."new column" == [2, 2]
+    @test @based_on(gd, cols("new" * "_" * "column") = 2)."new_column" == [2, 2]
+
```
nalimilan, line 89:
```suggestion
```

test/grouping.jl
```diff
@@ -144,14 +152,18 @@ end
...
+    n_space = "new column"
+
     @test @by(df, :g, n = mean(:i)).n == [2.0, 4.5]
     @test @by(df, :g, n = mean(:i) + mean(:g)).n == [3.0, 6.5]
     @test @by(df, :g, n = first(:t .* string.(:y))).n == ["av", "cy"]
     @test @by(df, :g, n = first(Symbol.(:y, ^(:t)))).n == [:vt, :yt]
     @test @by(df, :g, n = first(Symbol.(:y, ^(:body)))).n == [:vbody, :ybody]
     @test @by(df, :g, body = :i).body == df.i
     @test @by(df, :g, transform = :i).transform == df.i
+    # TODO: Make `t = :n1; (cols(t) = ...,)` work
```
nalimilan, line 166: Is this a TODO for now or for later? If the latter, is there a way to test the failure?

nalimilan (commented):
Looks great.

test/grouping.jl
pdeffebach: I've decided it isn't a failure. We already have lots of ways to make named tuples with variables as the names. I'm deleting this comment.

pdeffebach (commented):
Thanks for the comments!

README.md
pdeffebach: ths comma is necessary

README.md
pdeffebach: added.

README.md
nalimilan: AFAIK it has to be either "the special syntax `cols`" or "a special syntax, `cols`". :-)

pdeffebach: I know you always correct my writing! but I promise the comma belongs lol, [source](https://opinionator.blogs.nytimes.com/2012/05/21/the-most-comma-mistakes/)

nalimilan: Well we're exactly in the "I went to see the movie “Midnight in Paris” with my friend Jessie." case AFAICT! :-D

pdeffebach: I will change it to get this merged! But this is the *only* syntax with those properties, so we are in the "latest movie" case.

nalimilan: But it's not the only "special syntax" in the world, contrary to "Woody Allen’s latest movie"!

pdeffebach: Okay i see your point! 

With this settled, let's merge this! It will put us ahead of `dplyr` in this respect.

nalimilan (approved):

pdeffebach:
@bkamins please merge. I don't have write access yet it seems.

bkamins:
Thank you!

PR #182, opened 2020-10-02
Fix hygiene with @byrow

### LINKED ISSUE #172: Fix bug with `getproperty` in `@byrow`
```

julia> @byrow df begin
       :a = mean(df.b)
       end
ERROR: syntax: invalid syntax "df.(##255[row])" around REPL[4]:2
```

This is a problem because `QuoteNodes` show up in getproperty calls. 

```
julia> dump(:(d.x))
Expr
  head: Symbol .
  args: Array{Any}((2,))
    1: Symbol d
    2: QuoteNode
      value: Symbol x
```

This is a super easy fix, I think. `replace_dotted` already exists for other operations which cleans getproperty calls.

pdeffebach:
On master 

```

julia> _DF = 0; _N = 5;

julia> @byrow df begin 
       :A = :B
       end;

julia> _DF
3×2 DataFrame
│ Row │ A     │ B     │
│     │ Int64 │ Int64 │
├─────┼───────┼───────┤
│ 1   │ 2     │ 2     │
│ 2   │ 1     │ 1     │
│ 3   │ 2     │ 2     │

julia> _N
3
```

After this PR, I use `let` to make sure these variable stay inside `@byrow`. 

The fact that the user can use `_DF` and `_N` is weird, but I don't necessarily want to break it. People may like it. 

Additionally, this PR fixes #172 , which would accidentally escape the quotenode from a getproperty call. 

The following now does not error. 

```
    df4 = @byrow df begin
        :b = x.a
    end
```


bkamins:
I will have a look at it when CI passes

src/byrow.jl
```diff
@@ -52,13 +61,16 @@ function byrow_helper(df, body, deprecation_warning)
...
+    _df = gensym()
     quote
-        _N = length($df[!, 1])
-        _DF = @transform($df, $(e_newcols...))
-        $(with_helper(:_DF, :(for row = 1:_N
-            $(byrow_replace(e_body))
-        end)))
-        _DF
+        let $_df = $df
+            local _N = nrow($_df)
```
bkamins, line 67: what happens if you write to `_N` or `_DF`?

pdeffebach: Nothing bad. Since `_N` and `_DF` are only used when constructing new columns, something that happens before the anonymous function is evaluated, you don't break anything. 

You just don't get to use `_N` and `_DF` if you re-define them. 

Same with `row`. If the user has the correct mental model, that you are really writing `df.x[row]` then it should be obvious why things break.

bkamins (approved):

pdeffebach:
I went over this again and double checked that the scoping rules are the same as before. Adding the `let` block doesn't change anything about the scoping, given that the entire expression is wrapped in a function already. 

Let me know what kind of extra checks I should do before merging this.

bkamins:
I am OK to merge this.

pdeffebach:
Thanks!

PR #183, opened 2020-10-04
Fix mixing `Symbol`s and `String`s in `cols`

### LINKED ISSUE #181: Using strings in `cols`
We are missing tests for the following, which errors

```
@transform(df, z = cols("x") + :y)
```

DataFrames throws an error because we end up with 

```
["x", :y] => _f => :z 
```

which mixes a symbol and string in the input vector. 

i don't think think we should disallow using strings inside `cols`. We should probably turn everything into `Symbol`s  like 

```
Symbol.(["x", :y]) => _f = :z 
```

This means we will deviate from DataFrames, but this is an acceptable deviation imo.

pdeffebach:
Fixes #181 

This was an easy fix with easy tests to add. 

In the future this might get re-factored to allow for multiple column inputs in `cols`, but not today.

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,6 +162,59 @@ df3 = @byrow df begin
 end
 ```
 
+DataFramesMeta macros do not allow mixing of integer column references with references 
+of other types. This means `@transform(df, y = :A + cols(2))`, attempting to add the columns 
+`df[!, :A]` and `df[!, 2]`, will fail. This is because in DataFrames, the command 
+
+```julia
+transform(df, [:A, 2] => (+) => :y)
+``` 
+
+will fail, as DataFrames requires the "source" column identifiers in a 
+`source => fun => dest` pair to all have the same type. DataFramesMeta adds one exception
+to this rule. `Symbol`s and strings are allowed to be mixed inside DataFramesMeta macros. 
+Consequently, 
+
+```
+@transform(df, y = :A + cols("B"))
+```
+
+will not error even though 
+
+```
+transform(df, [:A, "B"] => (+) => :y)
+```
+
+will error in DataFrames. 
+
+For consistency, this restriction in the input column types also applies to `@with`
+and `@byrow`. You cannot mix integer column references with `Symbol` or string column 
+references in `@with` and `@byrow` in any part of the expression, but you can mix 
+`Symbol`s and strings. The following will fail:
+
+```julia
+df = DataFrame(A = 1:3, B = [2, 1, 2])
+@byrow df begin 
+    :A = cols(2)
+end
+
+@with df begin 
+    cols(1) + cols("A")
+end
+```
+
+while the following will work without error
+
+```julia
+@byrow df begin 
+    cols(1) = cols(2)
+end
+
+@with df begin 
+    cols(1) + cols(2)
+end
+```
+
 Note that `cols` is *not* a standard Julia function. It is only used to modify the 
 way that macros in DataFramesMeta.jl escape arguments and has no behavior of its own 
 outside of DataFramesMeta macros.
--- a/README.md
+++ b/README.md
@@ -136,8 +136,11 @@ df = DataFrame(A = 1:3, B = [2, 1, 2])
 nameA = :A
 df2 = @transform(df, C = :B - cols(nameA))
 
+nameA_string = "A"
+df2 = @transform(df, C = :B - cols(nameA_string))
+
 nameB = "B"
-df3 = @byrow df begin 
+df4 = @byrow df begin 
     :A = cols(nameB)
 end
 ```
--- a/src/DataFramesMeta.jl
+++ b/src/DataFramesMeta.jl
@@ -145,7 +145,7 @@ function repair_source(x)
     if isempty(x) || isconcretetype(eltype(x))
         return x
     elseif all(t -> t isa Union{AbstractString, Symbol}, x)
-        return [xi isa AbstractString ? Symbol(xi) : xi for xi in x]
+        return Symbol.(x)
     else
         throw(ArgumentError("Column references must be either all the same " *
                             "type or a a combination of `Symbol`s and strings"))
--- a/src/DataFramesMeta.jl
+++ b/src/DataFramesMeta.jl
@@ -116,7 +116,7 @@ function fun_to_vec(kw::Expr; nolhs = false)
         if nolhs
             # [:x] => _f
             t = quote
-                Symbol.($(source)) =>
+                DataFramesMeta.repair_source($(source)) =>
                 ($(Expr(:tuple, values(membernames)...)) -> $body)
             end
          else
@@ -128,7 +128,7 @@ function fun_to_vec(kw::Expr; nolhs = false)
                 output = kw.args[1].args[2]
             end
             t = quote
-                 Symbol.($(source)) =>
+                DataFramesMeta.repair_source($(source)) =>
                 ($(Expr(:tuple, values(membernames)...)) -> $body) =>
                 $(output)
             end
@@ -141,6 +141,16 @@ end
 
 fun_to_vec(kw::QuoteNode) = kw
 
+function repair_source(x)
+    if isconcretetype(eltype(x))
+        return x
+    elseif all(t -> t isa Union{AbstractString, Symbol}, x)
+        return [xi isa AbstractString ? Symbol(xi) : xi for xi in x]
+    else
+        # DataFrames can throw the error
+        return x
+    end
+end
 
 protect_replace_syms!(e, membernames) = e
 function protect_replace_syms!(e::Expr, membernames)
--- a/src/DataFramesMeta.jl
+++ b/src/DataFramesMeta.jl
@@ -116,7 +116,7 @@ function fun_to_vec(kw::Expr; nolhs = false)
         if nolhs
             # [:x] => _f
             t = quote
-                DataFramesMeta.repair_source($(source)) =>
+                DataFramesMeta.make_source_concrete($(source)) =>
                 ($(Expr(:tuple, values(membernames)...)) -> $body)
             end
          else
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -136,8 +136,11 @@ df = DataFrame(A = 1:3, B = [2, 1, 2])
 nameA = :A
 df2 = @transform(df, C = :B - cols(nameA))
 
+nameA_string = "A"
+df2 = @transform(df, C = :B - cols(nameA_string))
```
pdeffebach, line 140:
```suggestion
df3 = @transform(df, C = :B - cols(nameA_string))
```

bkamins:
what would happen if you passed `1` or `'1'` or `missing` instead of `Symbol` or `AbstractString`?

pdeffebach:
Thanks for this. Indeed my solution is too heavy-handed. 

```
df = DataFrame(a = 1)
```

* `1`: error, it looks up the symbol `:1`
* `'1'` same as above
* 'a': works, it looks up the symbol `:a` even though this is not a valid column selector
* `missing`: error, looks up symbol `:missing` (which is very weird)

Is there an internal DataFrames function I should use to sanitize this input?

pdeffebach:
This is actually going to be quite hard to support everything, since the implementation current is agnostic about the data frame that is actually being called. I will think on this more.

bkamins:
you can check if `source` is `ColumnIndex` or `MultiColumnIndex` maybe (both defined in DataFrames.jl)?

pdeffebach:
Thanks I will do that. I was also thinking

```
_names(df)[DataFrames.getindex(df, :a)]
```

or similar. 

The problem is that the current implementation only works with the expression `y = f(:x)` and doesn't rely on the data frame at all, and it's what makes something like 

```
DataFrames.transform(df, @cols y = f(:x))
```

a possibility.

I would have to break this correspondence to fix this. Perhaps it is worth it.

pdeffebach:
Here is a very ad-hoc solution

1. Define a function which takes in a dataframe and a vector of column identifiers. It then replaces all of the elements in the vector with integer column positions. 
2. `gensym` the name of the data frame, pass this symbol to the `fun_to_vec` function and write an expression which calls the above function when it returns.
3. Evaluate this expression, which works because we have passed around the relevant `gensym()`-ed symbol. 

Then the actual `transform` call is  

```
_df = df # a gensym()
transform(_df, DataFrames.repair_source(_df, [:a, "a")] => _f => :y)
```

I don't like this solution. It mixes run-time with parse time behavior. It also adds complexity that I don't want to add yet until I have figured out how multi-valued `cols` arguments will work. 

I would rather disallow integers. Disallowing integers means we can call a small function that just converts strings to symbols. If the user wants to use integers they can call `names(df)[4]` to get the 4th column. Thankfully this is easy to do inside `cols`. 

This is a breaking change, as currently you can do `cols(4)` in `@transform`. 


Details:

```
function repair_source(df, x)
    [xi isa Integer ? xi : DataFrames.columnindex(df, xi) for xi in x]
end

    _df = gensym()
    t = (fun_to_vec(arg, _df) for arg in args)

    for ti in t
        @show ti
    end

    quote
        $_df = $x
        $DataFrames.transform($_df, $(t...))
    end
end
```

bkamins:
With `DataFrames.columnindex` just remember that `0` is returned when a column is missing.

pdeffebach:
Do you think the solution is viable? Or do you think it is too ad-hoc. 

I think it breaks the "one-to-one" mental model between DataFramesMeta and DataFrames a bit too much by introdutcing the extra feature. 

Do you think this is a valid concern? Do you think users *really* want to do `@transform(df, y = :x + cols(1))`?

bkamins:
I would start with a usual question: what is a contract for `cols`?
I assumed it is that `cols(something)` is converted to `df[!, something]` verbatim.
Is this correct?
If yes then we do not need to do anything with what is inside `cols` - just pass it verbatim and let DataFrames.jl either resolve it or error.

> Do you think users `really` want to do `@transform(df, y = :x + cols(1))`?

No, but they might potentially want e.g.:
```
for i in 1:5
    @transform(df, y = :x + cols(i)) # of course something more useful here
end
```

pdeffebach:
I guess after the re-design, I would view `@transform` as a way to construct a `source => fun => dest` expression. So the contract with `cols` is "something that goes into the `source`. 

If this is the case then `:x + cols(1)` is clearly invalid, since `[:x, 1]` is an error. On the other hand `:x + cols("y")` should error, but is a small deviation. Importantly, you can reason about the resulting expression *without* knowing much about the data frame you are acting on.

bkamins:
Ah - right. I thought about the redesign only after I have commented. In this case, if you want to allow mixing `Symbol` and `AbstractString`, instead of `Symbol.(...)` I would write `(x -> x isa AbstractString ? Symbol(x) : x).(...)`

pdeffebach:
I will check if they are all the same type, and if so, return the vector. 

The only mix we allow is `[:x, "y"]`, as a special case. 

This way people can do `y = cols(4) + cols(5)`. This will be easy to reason about. If people complain, then I will re-think the design.

pdeffebach:
I have updated the implementation and think this is a reasonable fix. 

This change does *not* apply to `@with`-based macros, though, including `@byrow`. Perhaps I should enforce the same rules for consistency, even though the implementation is fundamentally different.

pdeffebach:
I will restrit `@with`, then, as well.

Maybe in the future DataFrames can expose its API for doing this process so `@with` can use it directly or other packages can opt-in.

pdeffebach:
Ready for a review! The `@with` code has been changed to not mix integers with other types.

README.md
```diff
@@ -136,8 +136,11 @@ df = DataFrame(A = 1:3, B = [2, 1, 2])
 nameA = :A
 df2 = @transform(df, C = :B - cols(nameA))
 
+nameA_string = "A"
```
bkamins, line 139: maybe make it a `julia> ....` and then print output (like we do in docstrings in DataFrames.jl)? It can be a separate PR of course later

src/DataFramesMeta.jl
```diff
@@ -139,6 +141,16 @@ end
 
 fun_to_vec(kw::QuoteNode) = kw
 
+function repair_source(x)
+    if isempty(x) || isconcretetype(eltype(x))
+        return x
+    elseif all(t -> t isa Union{AbstractString, Symbol}, x)
+        return [xi isa AbstractString ? Symbol(xi) : xi for xi in x]
```
bkamins, line 148: I think this is better for two reasons:
1. `Symbol(:x)` is no-op so no need for `if`
2. if `x` is not a vector then comprehension will convert it to a vector (but maybe you want it? How do you e.g. handle `Tuple`? Or maybe we should check that we got a vector?)

```suggestion
        return Symbol.(x)
```

bkamins (approved):
Looks good except the comments left.

README.md
pdeffebach: yeah I'm procrastinating some docs changes. I'm going to move things to a real documentation system soon so I won't be worried about the diffs.

src/DataFramesMeta.jl
pdeffebach: Thanks. 

w.r.t vector inputs, I'm using a `Vector` of column references in `@where` even though it's not strictly necessary. This is the right thing to do to avoid too much compilation.

pdeffebach:
Requesting a review from @nalimilan before merging.

src/DataFramesMeta.jl
```diff
@@ -139,6 +141,16 @@ end
 
 fun_to_vec(kw::QuoteNode) = kw
 
+function repair_source(x)
```
nalimilan, line 144: Is this appropriate? (It helps understanding the code.)
```suggestion
function repair_source(x::AbstractVector)
```

Could also use a more specific name like `to_symbols` or something similar.

src/DataFramesMeta.jl
```diff
@@ -163,15 +175,14 @@ function with_helper(d, body)
     membernames = Dict{Any, Symbol}()
     funname = gensym()
     body = replace_syms!(body, membernames)
-    if isempty(membernames)
```
nalimilan, line 166: Isn't this branch relevant anymore?

test/byrow.jl
```diff
@@ -99,13 +98,6 @@ end
     end
     @test df2 == DataFrame(A = [2, 1, 2], B = [2, 1, 2])
 
-    n = 1
-    df2 = @byrow df begin
-        cols(n) = :B
-    end
-    @test df2 == DataFrame(A = [2, 1, 2], B = [2, 1, 2])
```
nalimilan, line 106: Isn't this still allowed?

src/DataFramesMeta.jl
pdeffebach: I don't think so. That code is pre 0.6. My guess is splatting empty collections wasn't allowed back then or something. All tests pass, even ones with empty arguments. 

Plus, this branch is a source of inconsistent behavior, see #184

test/byrow.jl
pdeffebach: No, because of the change with `@with`. We now disallow mixing `:x` and `2` inside the inputs to the anonymous function. 

and `@byrow` is just 

`@with df begin 
    for row in 1:nrow(df)
        :x[row] = :y[row]
    end
end
```

The distinction between LHS and RHS isn't present in `@with` the way it is in `@transform`.

src/DataFramesMeta.jl
pdeffebach: I added the type annotation. 

However I think the name is good since it doesn't convert everything to `Symbol`s. If given a vector of integers, it leaves it untouched.

pdeffebach:
I have updated the readme with a complete discussion of the behavior and updated NEWS.md. 

I also renamed `repair_source` to `make_source_concrete` for clarity.

pdeffebach:
Bumping this, I thin kthe docs are good

bkamins:
Let us wait for @nalimilan to comment as he had some questions here.

nalimilan (approved):

bkamins:
Thank you!

PR #188, opened 2020-10-07
Reexport DataFrames

### LINKED ISSUE #185: Reexport DataFrames
It would probably make sense to reexport all of DataFrames, since you cannot do much with DataFramesMeta if you can't even use the `DataFrame` type.

pdeffebach:
fixes #185 

Less typing for me to get started also!

cc @nalimilan


bkamins (approved):

bkamins:
you need to add Reexport.jl version in `[compat]` I think

bkamins:
OK to merge when CI finishes.

pdeffebach:
Thanks!

pdeffebach:
All I did in this last commit is get rid of an unexported `DataFramesMeta.select` method which was used internally in the old macros, but not the new implementation.

src/DataFramesMeta.jl
```diff
@@ -378,15 +380,6 @@ macro where(d, args...)
 end
 
 
-##############################################################################
-##
-## select - select columns
-##
-##############################################################################
-
-select(d::AbstractDataFrame, arg) = d[!, arg]
```
nalimilan, line 387: Wow, this one was... embarrassing. Thanks for the PR!

PR #190, opened 2020-10-07
Delete Tables.jl as a dependency

### LINKED ISSUE #189: Don't rely on Tables
In #188 I realized that we depend on Tables.jl. But a quick search for `Tables.*` yields nothing. We probably don't need it as a direct dependency.

pdeffebach:
fixes #189


bkamins (approved):
Thank you!

PR #191, opened 2020-10-10
Re-do the `@orderby` backend

pdeffebach:
Before, there was some complicated `with_anonymous` machinery. 

Now the backends are `select` and `combine`. 

For an abstract data frame this is intuitive, just call `select` on all the arguments with a `nolhs = true` option and return the `sortperm` from that. 

For a grouped data frame, unfortunately `@orderby` currently re-orders the groups and returns a grouped dataframe. 

This annoying because if you do, say, 

```

julia> df = DataFrame(a = [1, 1, 2, 2], b = [6, 6, 7, 7]);

julia> gd = groupby(df, :a);

julia> @orderby(df, :b)
```

then the result of the anonymous function in `@orderby` is a vector for each group and since Julia has lexicographic ordering of vectors, you can get some surprising results. 

~This was never tested and I'm not sure who uses it. I am making a breaking change here by only allowing things inside `@orderby` to return scalars, i.e. `@orderby(df, :b)` will error, and `@orderby(df, mean(:b))` will work.~

Update: I have found a solution that is somewhat hacky but doable. Current behavior matches 100% old behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -264,7 +264,7 @@ The following operations are now included:
   GroupedDataFrame.
 
 - `orderby(g, d -> mean(d[:a]))` and `@orderby(g, mean(:a))` -- Sort
-  groups based on the given criteria. Returns a GroupedDataFrame.
+  observations by a given criteria. 
 
 - `DataFrame(g)` -- Convert groups back to a DataFrame with the same
   group orderings.
--- a/README.md
+++ b/README.md
@@ -126,7 +126,8 @@ end
 
 ## `@orderby`
 
-Sort rows by values in one of several columns or a transformation of columns.
+Sort rows in a `DataFrame` by values in one of several columns or a 
+transformation of columns.
 
 ```julia
 d = DataFrame(x = [3, 3, 3, 2, 1, 1, 1, 2, 1, 1], n = 1:10);
--- a/README.md
+++ b/README.md
@@ -124,6 +124,17 @@ df2 = @byrow df begin
 end
 ```
 
+## `@orderby`
+
+Re-order rows of a data frame 
+
+```julia
+d = DataFrame(x = [3, 3, 3, 2, 1, 1, 1, 2, 1, 1], n = 1:10);
+@orderby(d, -1 .* :n)
+g = groupby(d, :x);
+@orderby(g, :x, :n .- mean(:n))
+```
+
 ## Working with column names programmatically with `cols`
 
 DataFramesMeta.jl provides the special syntax `cols` for referring to 
--- a/README.md
+++ b/README.md
@@ -131,8 +131,7 @@ Sort rows by values in one of several columns or a transformation of columns.
 ```julia
 d = DataFrame(x = [3, 3, 3, 2, 1, 1, 1, 2, 1, 1], n = 1:10);
 @orderby(d, -1 .* :n)
-g = groupby(d, :x);
-@orderby(g, :x, :n .- mean(:n))
+@orderby(d, :x, :n .- mean(:x))
 ```
 
 ## Working with column names programmatically with `cols`
--- a/src/DataFramesMeta.jl
+++ b/src/DataFramesMeta.jl
@@ -1,6 +1,6 @@
 module DataFramesMeta
 
-using Tables, Reexport
+using Reexport
 
 @reexport using DataFrames
 
--- a/src/DataFramesMeta.jl
+++ b/src/DataFramesMeta.jl
@@ -410,19 +410,36 @@ function orderby(x::AbstractDataFrame, @nospecialize(args...))
 end
 
 function orderby(x::GroupedDataFrame, @nospecialize(args...))
-
     @warn "orderby behavior now returns a `DataFrame` instead of a `GroupedDataFrame`. " *
           "Group the returned data frame to restore old behavior" maxlog = 2
 
     t = DataFrames.select(x, args...; copycols = false, keepkeys = false)
+    @assert nrow(parent(x)) == nrow(t)
     parent(x)[sortperm(t), :]
 end
 
 """
     @orderby(d, i...)
 
-Sort by criteria. When using a grouped data frame, operation
-is performed at the group level.
+Sort by criteria. When given a `GroupedDataFrame`, operation
+is performed at the group level.`@orderby` always returns a fresh
+`DataFrame` regardless of whether the input is a `DataFrame` or a
+`GroupedDataFrame`.
+
+When given a `DataFrame`, `@orderby` applies the transformation
+given by its arguments (but does not create new columns) and sorts
+the given `DataFrame` on the result, returning a new `DataFrame`.
+
+When given a `GroupedDataFrame`, `@orderby` applies the transformation
+given by its arguments *by group*. Then it returns the `parent` of the
+input `GroupedDataFrame` which has been sorted by the transformations.
+The second example below shows the logic of `@orderby` with a
+`GroupedDataFrame`.
+
+When inputting a `GroupedDataFrame`, to return a `DataFrame` sorted
+first by group and then sorted by the tranformation, use the
+grouping columns of the the `GroupedDataFrame` as the first arguments
+of the`@orderby` call. See the final example below for a demonstration.
 
 ### Arguments
 
@@ -434,62 +451,60 @@ is performed at the group level.
 ```jldoctest
 julia> using DataFrames, DataFramesMeta, Statistics
 
-julia> d = DataFrame(n = 1:20, x = [3, 3, 3, 3, 1, 1, 1, 2, 1, 1,
-                                    2, 1, 1, 2, 2, 2, 3, 1, 1, 2]);
+julia> d = DataFrame(x = [3, 3, 3, 2, 1, 1, 1, 2, 1, 1], n = 1:10);
 
-julia> @orderby(d, :x - :n)
-20×2 DataFrame
-│ Row │ n     │ x     │
+julia> @orderby(d, -1 .* :n)
+10×2 DataFrame
+│ Row │ x     │ n     │
 │     │ Int64 │ Int64 │
 ├─────┼───────┼───────┤
-│ 1   │ 19    │ 1     │
-│ 2   │ 20    │ 2     │
-│ 3   │ 18    │ 1     │
-│ 4   │ 16    │ 2     │
-│ 5   │ 17    │ 3     │
-│ 6   │ 15    │ 2     │
-│ 7   │ 13    │ 1     │
-│ 8   │ 14    │ 2     │
-│ 9   │ 12    │ 1     │
-│ 10  │ 10    │ 1     │
-│ 11  │ 11    │ 2     │
-│ 12  │ 9     │ 1     │
-│ 13  │ 7     │ 1     │
-│ 14  │ 8     │ 2     │
-│ 15  │ 6     │ 1     │
-│ 16  │ 5     │ 1     │
-│ 17  │ 4     │ 3     │
-│ 18  │ 3     │ 3     │
-│ 19  │ 2     │ 3     │
-│ 20  │ 1     │ 3     │
+│ 1   │ 1     │ 10    │
+│ 2   │ 1     │ 9     │
+│ 3   │ 2     │ 8     │
+│ 4   │ 1     │ 7     │
+│ 5   │ 1     │ 6     │
+│ 6   │ 1     │ 5     │
+│ 7   │ 2     │ 4     │
+│ 8   │ 3     │ 3     │
+│ 9   │ 3     │ 2     │
+│ 10  │ 3     │ 1     │
 
 julia> g = groupby(d, :x);
 
-julia> @orderby(g, mean(:x))
-20×2 DataFrame
+julia> @linq g |>
+       transform(t = :n .- mean(:n)) |>
+       orderby(:n .- mean(:n))
+
+10×3 DataFrame
+│ Row │ x     │ n     │ t       │
+│     │ Int64 │ Int64 │ Float64 │
+├─────┼───────┼───────┼─────────┤
+│ 1   │ 3     │ 1     │ -1.0    │
+│ 2   │ 3     │ 2     │ 0.0     │
+│ 3   │ 3     │ 3     │ 1.0     │
+│ 4   │ 2     │ 4     │ -2.0    │
+│ 5   │ 1     │ 5     │ -2.4    │
+│ 6   │ 1     │ 6     │ -1.4    │
+│ 7   │ 1     │ 7     │ -0.4    │
+│ 8   │ 2     │ 8     │ 2.0     │
+│ 9   │ 1     │ 9     │ 1.6     │
+│ 10  │ 1     │ 10    │ 2.6     │
+
+julia> @orderby(d, :x, -1 .* :n)
+10×2 DataFrame
 │ Row │ n     │ x     │
 │     │ Int64 │ Int64 │
 ├─────┼───────┼───────┤
-│ 1   │ 5     │ 1     │
-│ 2   │ 6     │ 1     │
+│ 1   │ 10    │ 1     │
+│ 2   │ 9     │ 1     │
 │ 3   │ 7     │ 1     │
-│ 4   │ 9     │ 1     │
-│ 5   │ 10    │ 1     │
-│ 6   │ 12    │ 1     │
-│ 7   │ 13    │ 1     │
-│ 8   │ 18    │ 1     │
-│ 9   │ 19    │ 1     │
-│ 10  │ 8     │ 2     │
-│ 11  │ 11    │ 2     │
-│ 12  │ 14    │ 2     │
-│ 13  │ 15    │ 2     │
-│ 14  │ 16    │ 2     │
-│ 15  │ 20    │ 2     │
-│ 16  │ 1     │ 3     │
-│ 17  │ 2     │ 3     │
-│ 18  │ 3     │ 3     │
-│ 19  │ 4     │ 3     │
-│ 20  │ 17    │ 3     │
+│ 4   │ 6     │ 1     │
+│ 5   │ 5     │ 1     │
+│ 6   │ 8     │ 2     │
+│ 7   │ 4     │ 2     │
+│ 8   │ 3     │ 3     │
+│ 9   │ 2     │ 3     │
+│ 10  │ 1     │ 3     │
 ```
 
 """
--- a/src/DataFramesMeta.jl
+++ b/src/DataFramesMeta.jl
@@ -387,39 +387,28 @@ end
 ##
 ##############################################################################
 
-# takes calls `:a` and makes them `identity(:a)` so
-# that we always get new column names and can
-# discard the key columns when doing `sortperm`.
-_make_expr(e::Expr) = e
-_make_expr(e::QuoteNode) = Expr(:call, :identity, e)
-
 function orderby_helper(x, args...)
-    t = (fun_to_vec(_make_expr(arg); nolhs = true) for arg in args)
+    t = (fun_to_vec(arg; nolhs = true) for arg in args)
     quote
         $DataFramesMeta.orderby($x, $(t...))
     end
 end
 
-function orderby(x::AbstractDataFrame, args...)
+function orderby(x::AbstractDataFrame, @nospecialize(args...))
     t = DataFrames.select(x, args...; copycols = false)
     x[sortperm(t), :]
 end
 
-
-function orderby(x::GroupedDataFrame, args...)
-    out_gd = DataFrames.combine(x, args...; ungroup = false)
-
-
-   # if nrow(out_df) != length(x)
-   #     throw(ArgumentError("All inputs to `@orderby` should return a scalar"))
-   # end
-    x[sortperm([eachcol(sdf[!, Not(out_gd.cols)]) for sdf in out_gd])]
+function orderby(x::GroupedDataFrame, @nospecialize(args...))
+    t = DataFrames.select(x, args...; copycols = false)
+    parent(x)[sortperm(t), :]
 end
 
 """
     @orderby(d, i...)
 
-Sort by criteria. Normally used to sort groups in GroupedDataFrames.
+Sort by criteria. When using a grouped data frame, operation
+is performed at the group level.
 
 ### Arguments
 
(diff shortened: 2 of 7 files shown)


pdeffebach:
Additionally, it looks like `@orderby` wasn't really tested! So I have to add lots of tests.

bkamins:
I would prefer if you propose the solution that "makes sense" for the future and if this needs to break in the case of `GroupedDataFrame` just break it.

pdeffebach:
The solution I would like is 

```
julia> df = DataFrame(a = [1, 1, 1, 2, 2, 2], b = [1, 0, -1, 3, 0, -3]);

julia> gd = groupby(df, :a);

julia> @orderby(gd, :b .- mean(:b))
```

which would end up performing the operation `:b .- mean(:b)` the way it would work in `@select` and then calling 

```
out_df = @select(gd, :_x = :b .- mean(:b))
parent(gd)[sortperm(out_df))
```

bkamins:
Yes - this would be super cool. And I a similar thing will be supported in `where` in DataFrames.jl once I get to implementing it.

pdeffebach:
This is ready for a review. 

For posterity, I want to emphasize that this change reduces performance. @bkamins I was very surprised that `gd[[2, 1]]` is doesn't do any copies!

But I think this is the better and more intuitive interface. I don't really understand the original authors' preoccupation with reordering and selecting different groups in a DataFrame.

pdeffebach:
Tests are broken, will fix soon.

bkamins:
> @bkamins I was very surprised that gd[[2, 1]] is doesn't do any copies!

I have not designed it (expect for cleaning bugs), but I guess this was intentional to ensure we have a good performance.

> I don't really understand the original authors' preoccupation with reordering and selecting different groups in a DataFrame.

Again - I am not the author, but I guess after you order groups (which is fast), if you want to materialize a `DataFrame` you can always call `DataFrame` on the result (and if you do not need a materialized `DataFrame` you can leave things as they are - e.g. when you later want to run some aggregation function).

pdeffebach:
Apologies for the delay on this. It is now ready for a review

bkamins:
Just to be clear - why do you prefer `@orderby` to return a `DataFrame` not `GroupedDataFrame`?

pdeffebach:
1. Returning grouped data frame is inconsistent with `@select` and `@transform`
2. With the addition of `@select` and `@transform`, the order of groups doesn't matter as much, so i'm not sure the original use-case applies
3. something like `@orderby(groupby(df, :g), :x)` is very confusing. Current behavior compares the arrays `gd[1].x`, `gd[2].x` etc. This is lexicographic ordering because that's how arrays are compared in Julia. I think this is confusing. 
4. Implementation seems hard when using the new backend.

bkamins:
OK

src/DataFramesMeta.jl
```diff
@@ -410,6 +410,10 @@ function orderby(x::AbstractDataFrame, @nospecialize(args...))
 end
 
 function orderby(x::GroupedDataFrame, @nospecialize(args...))
+
+    @warn "orderby behavior now returns a `DataFrame` instead of a `GroupedDataFrame`. " *
+          "Group the returned data frame to restore old behavior" maxlog = 5
```
bkamins, line 415: probably `maxlog = 1` or `2` is enough?

pdeffebach: added.

pdeffebach:
@nalimilan Are you okay with me breaking this?

bkamins:
I would recommend that you just push forward with what you think it a good design (of course @nalimilan and I will gladly consult you). You have this freedom as DataFramesMeta.jl is not that big/mature - so I would recommend to shape it aggressively (the reason is that once it matures/becomes widely used - you will hit what we have in DataFrames.jl - that every decision, even ones that seem simple, are taking a lot of time).

pdeffebach:
That sounds good. I will break `@where` in a similar way, then. 

I will merge tomorrow if there are no objections

nalimilan:
I'm fine with you breaking thinks, and I agree it makes sense to make `@orderby` closer to `@select`. Though maybe we should think carefully about the consistency of the API in general? In particular, I wonder what's the relation between `@orderby` and `sort`. For `@select`, `transform` and `combine` there's a one-to-one mapping with DataFrames functions. Should it be the case for other macros too?

For `DataFrame`, `orderby` does the same thing as `sort` AFAICT, it just allows computing a new column and sorting on it in a single operation. For `GroupedDataFrame` there's the issue that `sort` should probably sort groups rather than rows, similar to what we discussed for `filter`. So maybe actually the current behavior of `@orderby` is what a hypothetical `@sort` would do?

Also, is the goal that all macros return a `DataFrame` by default, even when passed a `GroupedDataFrame`?

FWIW, in dplyr `arrange` doesn't take into account grouping at all when sorting, but preserves grouping, and this is explicitly mentioned as an exception in the dplyr API. Not sure whether they are happy about that, but in issues mentioning that on GitHub they didn't mention any regret.

bkamins:
> For `GroupedDataFrame` there's the issue that sort should probably sort groups rather than rows

I understand that this is what `@orderby` does now. That is why I was OK with it. The only thing was that `DataFrame` is returned, but I understand @pdeffebach wants `GroupedDataFrame` not to be sticky in DataFramesMeta.jl, but dropped always. Actually this is the point of the design that would be good to understand the reasons behind better. @pdeffebach - can you comment on this more?

pdeffebach:
Thanks, Milan, for the feedback. 

My motivation is that with current behavior, 

```
@orderby(groupby(df, :g), :x .- mean(:x))
```

gives kind of a non-sense result currently. It compares the vectors `:x .- mean(:x)` with lexicographic ordering. 

My proposal is actually very close to `arrange`

```
r$> t = tibble(a = c(1, 1, 2, 2), b = c(11, 10, 101, 100))                                                

r$> g = group_by(t, a)                                                                                    

r$> arrange(g, b - mean(b))                                                                               
# A tibble: 4 x 2
# Groups:   a [2]
      a     b
  <dbl> <dbl>
1     1    10
2     1    11
3     2   100
4     2   101
```

See that results are sorted *within* a group by the output of `b - mean(b)`. 

The behavior in this PR is the same as `arrange`, except where we get rid of grouping at the end. 

In relation to the rest of `dplyr`, `arrange` is inconsistent imo. And is a cause of frequent calls to `ungroup` in the middle of pipes.

Note that if we had a `filter` function which took in columns, I would maybe advocate that it shouldn't just filter groups. But since it takes in rows, there is no added benefit to different ways of selecting observations to keep. Additionally, because a `GroupedDataFrame` iterates through groups we have to keep consistency with base. 

In a similar way, `sort` should return a grouped data frame for consistency with base. 

I don't think that's the most intuitive behavior for chaining manipulations, though, and we get the benefit of a new name, `@orderby`, where we dont have to be consistent with base.

bkamins:
> See that results are sorted within a group by the output of `b - mean(b)`

I would recommend to make the docstring more precise and the example more explicit to show this (now you sort by the constant). Also if you want to do "within group sorting" I understand that this means you do not change the order of groups? In particular you use `mean(:x)` which is a constant within group if it is grouped by `:x`.

pdeffebach:
yes, sorry. I was wrong in the above post. The current implementation does *not* order within groups. 

If someone wants to order within groups, they can include the grouping variable

```
@orderby(df, :g, :b .- mean(:b))
```

I like this, as it is more explicit.

nalimilan:
I agree that the current behavior doesn't make much sense. But we should have a general plan regarding:
1. Which functions operate within groups (we could always implicitly sort within groups if we wanted). It seems that `@orderby` would be an exception in ignoring grouping, right? There may be good reasons for this but we should be clear about it.
2. Which functions ungroup `GroupedDataFrame` inputs. IIUC your position is that all functions should ungroup? Are there exceptions to this?

pdeffebach:
I think the rules are 

1. Any functions that take in columns, i.e. `@select` and `@orderby`, should take in grouped data frames and preform every operation by group. `@orderby` doesn't ignore groups, because the vector in `:b .- mean(:b)` is actually by group.
2. Everything ungroups. In the future, when we gain the ability to do keyword arguments, we can add the ability to *not* ungroup.  But by default, `@where` and `@orderby` produce `DataFrame`s just like `@select` and `@by`

src/DataFramesMeta.jl
```diff
@@ -397,33 +397,32 @@ end
...
 
-function orderby_helper(d, args...)
-    _D = gensym()
-    quote
-        let $_D = $d
-            $orderby($_D, $(with_anonymous(:($orderbyconstructor($_D)($(args...))))))
-        end
-    end
+    t = DataFrames.select(x, args...; copycols = false, keepkeys = false)
+    parent(x)[sortperm(t), :]
```
nalimilan, line 418: Is this correct if some groups have been dropped?

test/dataframes.jl
```diff
@@ -277,7 +277,24 @@ end
...
+@testset "orderby" begin
+    df = DataFrame(
+        g = [1, 1, 1, 2, 2],
+        i = 1:5,
+        t = ["a", "b", "c", "c", "e"],
+        y = [:v, :w, :x, :y, :z],
+        c = [:g, :quote, :body, :transform, missing]
+        )
+
+    gd = groupby(df, :g)
```
nalimilan, line 289: It's a bit weird to test `GroupedDataFrame` in dataframes.jl, right?

test/linqmacro.jl
```diff
@@ -38,6 +36,7 @@ xlinq2 = @linq df  |>
     transform(y = 10 * :x)  |>
     groupby(:b) |>
     orderby(-mean(:x))  |>
+    groupby(:b) |>
```
nalimilan, line 39: This sounds really suboptimal. Maybe change the tests to do something more logical? Same below.

nalimilan:
OK, sounds good. I'm still a bit concerned that we don't have the equivalent `orderby` in DataFrames but we can discuss that later. It sounds like we would need to duplicate all Base functions to have a more useful behavior for `GroupedDataFrame`... :-/

src/DataFramesMeta.jl
pdeffebach: Can you clarify circumstances in which groups can be dropped? That should never happen with `select`, right?

pdeffebach:
> OK, sounds good. I'm still a bit concerned that we don't have the equivalent `orderby` in DataFrames but we can discuss that later. It sounds like we would need to duplicate all Base functions to have a more useful behavior for `GroupedDataFrame`... :-/

Thankfully the implementation of this is super trivial and based off `select` should we decide to implement.

src/DataFramesMeta.jl
bkamins: in `select` groups are not allowed to be dropped, but can be reordered (but reordering is ignored).

nalimilan: But `x` could have been subsetted before calling `orderby`. Isn't that a problem since `sortperm(t)` will return a permutation of rows that have not been dropped? It's easy to check by adding `@assert nrow(t) == nrow(parent(x))`.

pdeffebach: `select` should fail in that case, but I will add the check to ensure.

pdeffebach:
Thanks for the comments. I have added a much more extensive docstring and responded to your requests for comments.

src/DataFramesMeta.jl
bkamins: yes - this is what I meant. `select` will fail if `x` is subsetted, but will not fail if `x` is just reordered

src/DataFramesMeta.jl
```diff
@@ -435,38 +451,64 @@ Sort by criteria. Normally used to sort groups in GroupedDataFrames.
...
+│ 3   │ 3     │ 3     │ 1.0     │
+│ 4   │ 2     │ 4     │ -2.0    │
+│ 5   │ 1     │ 5     │ -2.4    │
+│ 6   │ 1     │ 6     │ -1.4    │
+│ 7   │ 1     │ 7     │ -0.4    │
+│ 8   │ 2     │ 8     │ 2.0     │
+│ 9   │ 1     │ 9     │ 1.6     │
+│ 10  │ 1     │ 10    │ 2.6     │
+
+julia> @orderby(d, :x, -1 .* :n)
```
bkamins, line 493: why is the order of columns in the output changed here?

src/DataFramesMeta.jl
```diff
@@ -435,38 +451,64 @@ Sort by criteria. Normally used to sort groups in GroupedDataFrames.
...
-├─────┼────┼───┤
-│ 1   │ 8  │ 2 │
-│ 2   │ 11 │ 2 │
-│ 3   │ 14 │ 2 │
-│ 4   │ 15 │ 2 │
-│ 5   │ 16 │ 2 │
-│ 6   │ 20 │ 2 │
+julia> @linq g |>
+       transform(t = :n .- mean(:n)) |>
+       orderby(:n .- mean(:n))
```
bkamins, line 476: I do not understand this example. I would assume that the result of `transform` is a `DataFrame` and then you essentially sort it by `:n` (it was already sorted by `:n` actually, so here `@orderby` is no-op). But how does it give us an example how `@orderby` works on `GroupedDataFrame`?

src/DataFramesMeta.jl
pdeffebach: sorry, bad copy and pasting on my part. I thought it was nicer when the grouping column was the first column.

src/DataFramesMeta.jl
pdeffebach: Here is an extra sentence I added to the dosctring 

> Note that the column `:t` is arranged from
lowest to highest after the `@orderby` command. This shows that
`@orderby` is equivelent to a transformation by group followed
by ordering on the subsequent transformation.

Does this explanation make sense?

src/DataFramesMeta.jl
```diff
@@ -410,19 +410,39 @@ function orderby(x::AbstractDataFrame, @nospecialize(args...))
...
+the given `DataFrame` on the result, returning a new `DataFrame`.
+
+When given a `GroupedDataFrame`, `@orderby` applies the transformation
+given by its arguments *by group*. Then it returns the `parent` of the
+input `GroupedDataFrame` which has been sorted by the transformations.
+The second example below shows the logic of `@orderby` with a
+`GroupedDataFrame`. Note that the column `:t` is arranged from
+lowest to highest after the `@orderby` command. This shows that
+`@orderby` is equivelent to a transformation by group followed
+by ordering on the subsequent transformation.
```
nalimilan, line 440: Better put this with the corresponding example.

src/DataFramesMeta.jl
```diff
@@ -410,19 +410,39 @@ function orderby(x::AbstractDataFrame, @nospecialize(args...))
...
+`DataFrame` regardless of whether the input is a `DataFrame` or a
+`GroupedDataFrame`.
+
+When given a `DataFrame`, `@orderby` applies the transformation
+given by its arguments (but does not create new columns) and sorts
+the given `DataFrame` on the result, returning a new `DataFrame`.
+
+When given a `GroupedDataFrame`, `@orderby` applies the transformation
+given by its arguments *by group*. Then it returns the `parent` of the
+input `GroupedDataFrame` which has been sorted by the transformations.
```
nalimilan, line 435:
```suggestion
given by its arguments *by group*. Then it returns rows of the
input `GroupedDataFrame` sorted by the transformations.
```
(The parent is really an implementation detail.)

src/DataFramesMeta.jl
```diff
@@ -410,19 +410,39 @@ function orderby(x::AbstractDataFrame, @nospecialize(args...))
...
 
 """
     @orderby(d, i...)
 
-Sort by criteria. When using a grouped data frame, operation
-is performed at the group level.
+Sort by criteria. When given a `GroupedDataFrame`, operation
+is performed at the group level.`@orderby` always returns a fresh
+`DataFrame` regardless of whether the input is a `DataFrame` or a
+`GroupedDataFrame`.
```
nalimilan, line 427:
```suggestion
Sort rows by values in one of several columns or a transformation of columns.
Always returns a fresh `DataFrame` regardless of whether the input is a `DataFrame` or a
`GroupedDataFrame`.
```
(How things work with `GroupedDataFrame` is mentioned below. It also sounds weird that the second sentence would mention `GroupedDataFrame` before even mentioning `DataFrame`.)

test/grouping.jl
```diff
@@ -291,4 +291,22 @@ end
...
+        t = ["a", "b", "c", "c", "e"],
+        y = [:v, :w, :x, :y, :z],
+        c = [:g, :quote, :body, :transform, missing]
+        )
+
+    gd = groupby(df, :g)
+
+    @test @orderby(gd, mean(:i)).i == [1, 2, 3, 4, 5]
+    @test @orderby(df, std(:i) .- :i).i == [5, 4, 3, 2, 1]
+    @test @orderby(gd, :g, -1 .* (:i .- mean(:i))).i == [3, 2, 1, 5, 4]
```
nalimilan, line 308: Maybe check against `@orderby(@select(gd, ...), ...)`? That will also allow checking other columns.

Also `df` should be `gd` on the second line.

test/linqmacro.jl
```diff
@@ -70,6 +68,7 @@ xlinq3 = @linq df  |>
         transform(cols(y_str) = 10 * cols(x_sym))  |>
         DataFrames.groupby(b_str) |>
         orderby(-mean(cols(x_sym)))  |>
+        groupby(:b) |>
         based_on(cols("meanX") = mean(:x), meanY = mean(:y))
```
nalimilan, line 72: This example is really weird. Why isn't `orderby` called as the last step?

README.md
```diff
@@ -320,7 +331,7 @@ The following operations are now included:
   GroupedDataFrame.
 
 - `orderby(g, d -> mean(d[:a]))` and `@orderby(g, mean(:a))` -- Sort
-  groups based on the given criteria. Returns a GroupedDataFrame.
+  observations by a given criteria.
```
nalimilan, line 334:
```suggestion
  rows by a given criteria. Returns a `DataFrame`.
```

nalimilan:
Now let me play the devil's advocate. :-p

What are the use cases for having `@orderby` return a `DataFrame` rather than a `GroupedDataFrame`? That sounds useful only if sorting within groups was the goal of the operation and the last part of the piping sequence, right? Is the idea that people should either sort before grouping, or after ungrouping (using e.g. `select`)?

Also, functions like `@where` will keep returning a `GroupedDataFrame`, right?

bkamins:
@pdeffebach - given the discussion we have in this PR a more general question occurred to me. Here we discuss about `@orderby` macro, but it would be natural to define `@sort` macro (as in other cases - macro name is the same as function name). So the questions are:
* why at all we want `@orderby` macro (maybe we should remove it and just define `@sort` macro)
* does this macro have to work on `GroupedDataFrame` at all (maybe it is enough to allow sorting data frames) and then compose this sorting with `select` and `transform` that work on `GroupedDataFrame`

What do you think?

pdeffebach:
> What are the use cases for having `@orderby` return a `DataFrame` rather than a `GroupedDataFrame`? That sounds useful only if sorting within groups was the goal of the operation and the last part of the piping sequence, right? Is the idea that people should either sort before grouping, or after ungrouping (using e.g. `select`)?

I don't have a strong preference about returning a grouped data frame or a data frame. What I do have a preference for is not re-ordering groups, but rather performing an observation by group and re-ordering the result by rows. This seems like a more common need than sorting groups. Returning a `DataFrame` after this seems more about consistency. It's not good to have some operations on `GroupedDataFrame`s be sticky and others not. 



> * why at all we want `@orderby` macro (maybe we should remove it and just define `@sort` macro)
> * does this macro have to work on `GroupedDataFrame` at all (maybe it is enough to allow sorting data frames) and then compose this sorting with `select` and `transform` that work on `GroupedDataFrame`

I'm not super opposed to requiring a `@transform` and then a `@sort`, but no one likes having to create temporary columns, especially when dealing with data frames with many columns. In general we should avoid the kind of "calculations as columns" that Stata users have to do. 

In the interest of clarity I could re-name everything from scratch, here is what I would do 

* `@gencols`: `@transform` (both `DataFrame` and `GroupedDataFrame`)
* `@keepcols`: `@select` (both `DataFrame` and `GroupedDataFrame`)
* `@keeprows`: My proposed implementation of `@where`, which always returns a DataFrame and performs transformations by group
* `@keepgroups`: Current implementation of `@where` with a `GroupedDataFrame`
* `@sortrows`: My proposed implementation of `@orderby` in this PR, which always returns a DataFrame and performs transformations by group
* `@sortgroups`: Current implementation of `@orderby` on a `GroupedDataFrame`
* `@with`: Works only with a `DataFrame` (added to this list for completeness).

nalimilan:
> I'm not super opposed to requiring a `@transform` and then a `@sort`, but no one likes having to create temporary columns, especially when dealing with data frames with many columns. In general we should avoid the kind of "calculations as columns" that Stata users have to do.

I think the idea was that `@sort` would also allow sorting on transformations, without creating columns manually before sorting. Would that address your concern?

Following your terminology, I wonder whether the `@keepgroups` and `@sortgroups` are really needed. AFAICT dplyr doesn't support that, right?

pdeffebach:
Yes, I don't see a need for `@keepgroups` and `@sortgroups`. This is a main motivation for changing `@orderby` and `@where`, as this is their current functionality. 

Given that `GroupedDataFrame`s iterate through groups, I think there is value in the explicitness of `rows` vs `groups`.

nalimilan:
How about deprecating passing `GroupedDataFrame` to `@orderby` then and see how it goes?

pdeffebach:
What if I want to 

1. Sort by groups
2. Within a group, sort by a values deviation from the group mean

This is the kind of functionality this PR gives

```
@orderby(groupby(df, :g), :g, :v .- mean(:v))
```

bkamins:
>  sort by a values deviation from the group mean

```
@orderby(groupby(df, :g), :g, :v .- mean(:v))
```
is the same as:
sort(df, [:g, :v])

but I see the point. Still maybe (I will keep using DataFrames.jl syntax, not to fix ourselves on the solution for DataFramesMeta.jl):
```
combine(groupby(df, :g, sort=true), sdf -> sort(sdf, :v => v => v .- mean(v))
```
or
```
transform(groupby(sort(df, :g), :g), sdf -> sort(sdf, :v => v => v .- mean(v))
```
seems to be legible enough (and I have opened https://github.com/JuliaData/DataFrames.jl/issues/2489 to allow transformations in DataFrames.jl in the future; of course in DataFramesMeta.jl we can allow transformations already now)

pdeffebach:
After this discussion, in general, I am becoming more confident that this PR is the right move

1. As Milan mentioned, there isn't a big need to order groups, the `@sortgroups` macro above
2. The `combine` solution post by Bogumil is cumbersome and requires much more typing
3. As a matter of conisistency, it's important that all macros return `DataFrame`s. 

Though I do wish I could go ahead and implement all the re-naming macros given above...

test/grouping.jl
pdeffebach: Thanks!

test/linqmacro.jl
pdeffebach: Thanks, I have changed this pipe to match the other ones (it should be the same, just adding `cols` everywhere).

pdeffebach:
Thanks for the feedback. 

I will think on this more for a few days, but I am still leaning towards keeping this. I like the consistency that every macro which wither implicitely or explicitely performs a transformation allows a grouped data frame and does the transformation by group. 

For instance, I think we can all agree that `@where` should perform operations by group. For instance, taking the row with the highest value for each group is something people generally want to do. 

This would leave `@orderby` as lonely, the only macro which does not allow for a grouped operation returning a DataFrame.

bkamins:
As I have commented in other PR - if `@orderby` for `GroupedDataFrame` is a functionality that is needed very rarely we can leave it out for now and decide how to handle it later when we have very clear use cases. Thank you for working on this.

nalimilan:
To me the main advantage of not supporting `GroupedDataFrame` inputs would be that we wouldn't need `@orderby` but we could call it `@sort` instead, which is a natural name and consistent with Base. That's a similar decision to adding `where` in DataFames corresponding to `@where` in DataFramesMeta, while we already have `filter` but with semantics that don't match our needs for `GroupedDataFrame`.

pdeffebach:
Okay, I will "reserve" `@orderby` with a grouped data frame. 

However I still have a concern about `@sort`. I do think that the feature I propose for `@orderby` is useful overall. It's not a super high priority so I am fine not implementing it now. But once we use `@sort`, the feature I propose is off the table since it would break a contract with Base. 

So maybe I'm not sure what the solution is in the long run. It would be off for DataFrames to define both `sort` and `orderby`. But then again we will have both `filter` and `where`. 


Maybe `sort`'s transformations should operate on row-wide! That would be most similar to the contract with Base and open up an path for an alternative `orderby`. Similar to `filter` and `where`.

pdeffebach:
Okay ready for a review after disallowing a GroupedDataFrame

bkamins (approved):

bkamins:
OK - let's get rolling with it. Nightly fails due to unrelated reasons - right?

pdeffebach:
Yes, the errors are due to `MacroTools` and `Lazy` which are only imported during the `chaining` test set. This is ready to merge.

bkamins:
OK - go ahead and merge (if you do not have rights please let me know).

PR #192, opened 2020-10-16
Re-do the backend of `@where`

pdeffebach:
Consistent with the new backend of `@orderby` in #191 , I am writing a new backend for `@where`

1. Always returns a `DataFrame`
2. Makes a transformation by group and then filters

Need to add tests and write docs still.

### DIFF
--- a/src/DataFramesMeta.jl
+++ b/src/DataFramesMeta.jl
@@ -100,7 +100,7 @@ end
 # `@based_on(gd, fun(:x, :y))` where `fun` returns a `table` object.
 # We don't create the "new name" pair because new names are given
 # by the table.
-function fun_to_vec(kw::Expr; nolhs = false)
+function fun_to_vec(kw::Expr; nolhs = false, gensym_names = false)
     # nolhs: f(:x) where f returns a Table
     # !nolhs, y = g(:x)
     if kw.head === :(=) || kw.head === :kw || nolhs
@@ -116,10 +116,18 @@ function fun_to_vec(kw::Expr; nolhs = false)
         source = Expr(:vect, keys(membernames)...)
 
         if nolhs
-            # [:x] => _f
-            t = quote
-                DataFramesMeta.make_source_concrete($(source)) =>
-                ($(Expr(:tuple, values(membernames)...)) -> $body)
+            if gensym_names == false
+                # [:x] => _f
+                t = quote
+                    DataFramesMeta.make_source_concrete($(source)) =>
+                    ($(Expr(:tuple, values(membernames)...)) -> $body)
+                end
+            else
+                t = quote
+                    DataFramesMeta.make_source_concrete($(source)) =>
+                    ($(Expr(:tuple, values(membernames)...)) -> $body) =>
+                    $(QuoteNode(gensym()))
+                end
             end
          else
             if kw.args[1] isa Symbol
@@ -296,18 +304,34 @@ end
 ##############################################################################
 
 function where_helper(x, args...)
-    t = (fun_to_vec(arg; nolhs = true) for arg in args)
+    t = (fun_to_vec(arg; nolhs = true, gensym_names = true) for arg in args)
     quote
         $where($x, $(t...))
     end
 end
 
 and(x, y) = x .& y
+and(x) = x
 
-function where(x, args...)
-    res = DataFrames.select(x, args...; copycols = false, keepkeys = false)
-    tokeep = replace!(reduce(and, eachcol(res)), missing => false)
-    parent(x)[tokeep, :]
+function df_to_bool(res)
+    if any(t -> !(t isa Union{Vector{<:Union{Missing, Bool}}, BitArray{1}}), eachcol(res))
+        throw(ArgumentError("All arguments in @where must return a " *
+                            "Vector{<:Union{Missing, Bool} or a BitArray{1}"))
+    end
+
+    return reduce(and, eachcol(res)) .=== true
+end
+
+function where(df::AbstractDataFrame, @nospecialize(args...))
+    res = DataFrames.select(df, args...; copycols = false)
+    tokeep = df_to_bool(res)
+    df[tokeep, :]
+end
+
+function where(gd::GroupedDataFrame, @nospecialize(args...))
+    res = DataFrames.select(gd, args...; copycols = false, keepkeys = false)
+    tokeep = df_to_bool(res)
+    parent(gd)[tokeep, :]
 end
 
 """
--- a/src/DataFramesMeta.jl
+++ b/src/DataFramesMeta.jl
@@ -306,15 +306,13 @@ function where_helper(x, args...)
     end
 end
 
-and(x, y) = x .& y
-
 function df_to_bool(res::AbstractDataFrame)
     if any(t -> !(t isa AbstractVector{<:Union{Missing, Bool}}), eachcol(res))
         throw(ArgumentError("All arguments in @where must return an " *
                             "AbstractVector{<:Union{Missing, Bool}}"))
     end
 
-    return reduce(and, eachcol(res)) .=== true
+    return reduce((x, y) -> x .& y, eachcol(res)) .=== true
 end
 
 function where(df::AbstractDataFrame, @nospecialize(args...))
--- a/src/DataFramesMeta.jl
+++ b/src/DataFramesMeta.jl
@@ -346,6 +346,17 @@ Select row subsets in AbstractDataFrames and GroupedDataFrames.
 
 Multiple `i` expressions are "and-ed" together.
 
+If given a `GroupedDataFrame`, `@where` applies transformations by
+group, but does not store the result. `@where` returns a fresh
+`DataFrame` where the rows generated from the transformations are all
+`true`.
+
+!!! note
+    `@where` treats `missing` values as `false` when filtering rows.
+    Unlike `DataFrames.filter` and other boolean operations with
+    `missing`, `@where` will *not* error on missing values, and
+    will only keep `true` values.
+
 ### Examples
 
 ```jldoctest
@@ -400,6 +411,14 @@ julia> @where(g, :n .== first(:n))
 │ 2   │ 5     │ 1     │
 │ 3   │ 8     │ 2     │
 
+julia> d = DataFrame(a = [1, 2, missing], b = ["x", "y", missing]);
+
+julia> @where(d, :a .== 1)
+1×2 DataFrame
+│ Row │ a      │ b       │
+│     │ Int64? │ String? │
+├─────┼────────┼─────────┤
+│ 1   │ 1      │ x       │
 ```
 """
 macro where(x, args...)
--- a/src/DataFramesMeta.jl
+++ b/src/DataFramesMeta.jl
@@ -414,7 +414,7 @@ end
 ##############################################################################
 
 function orderby_helper(x, args...)
-    t = (fun_to_vec(arg; nolhs = true) for arg in args)
+    t = (fun_to_vec(arg; nolhs = true, gensym_names = true) for arg in args)
     quote
         $DataFramesMeta.orderby($x, $(t...))
     end
--- a/src/DataFramesMeta.jl
+++ b/src/DataFramesMeta.jl
@@ -100,7 +100,7 @@ end
 # `@based_on(gd, fun(:x, :y))` where `fun` returns a `table` object.
 # We don't create the "new name" pair because new names are given
 # by the table.
-function fun_to_vec(kw::Expr; nolhs = false, gensym_names = false)
+function fun_to_vec(kw::Expr; nolhs::Bool = false, gensym_names::Bool = false)
     # nolhs: f(:x) where f returns a Table
     # !nolhs, y = g(:x)
     if kw.head === :(=) || kw.head === :kw || nolhs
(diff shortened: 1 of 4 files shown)


pdeffebach:
@bkamins, @nalimilan 

Just to confirm before moving forward with this, despite the problems with `@orderby` with a GroupedDataFrame, we *do* want `@where` to perform tranformations by group and then return a DataFrame, right? 

Unlike `@orderby`, this feature will be widely used. 

```
@where(df, :x .> quantile(:x, .9)
```

for example, keep the rows with the top 10% `:x` value in each group.

bkamins:
In my words `@where` would get whole columns on input and do "within group" filtering and return a `DataFrame`. Right?

Would it drop `missing`? I assume yes.

What would be the order of rows? I assume the same as in the source (so it would not order rows by group but just filter the original data frame).

We have not discussed it yet (and I have not thought on it much), but I understand that this is also what `where` in DataFrames.jl would do. Then we would document `where` as working in the same way as `select` and `transform` with the exception that it drops rows.

pdeffebach:
Yes, I think it would treat `missing` as `false`. 

Great. I'm glad you agree. I will move forward with this, then.

bkamins:
There is one detail here. `@where` will not work on subsetted `GroupedDataFrame` (and obviously it will ignore group order). I would be good to document both things.

pdeffebach:
> There is one detail here. `@where` will not work on subsetted `GroupedDataFrame`

Can you clarify what `subsetted` grouped data frame means?

bkamins:
you can subset a `GroupedDataFrame` like `gdf[[1,3,5]]` (where you keep only groups 1,3, and 5 in this example)

pdeffebach:
Ah okay. I will make sure to document that!

pdeffebach:
Ready for a review!

src/DataFramesMeta.jl
```diff
@@ -116,10 +116,19 @@ function fun_to_vec(kw::Expr; nolhs = false)
         source = Expr(:vect, keys(membernames)...)
 
         if nolhs
-            # [:x] => _f
-            t = quote
-                DataFramesMeta.make_source_concrete($(source)) =>
-                ($(Expr(:tuple, values(membernames)...)) -> $body)
+            if gensym_names == false
```
bkamins, line 119: maybe change order and just use `gensym_names` as a condition?

src/DataFramesMeta.jl
```diff
@@ -100,7 +100,7 @@ end
 # `@based_on(gd, fun(:x, :y))` where `fun` returns a `table` object.
 # We don't create the "new name" pair because new names are given
 # by the table.
-function fun_to_vec(kw::Expr; nolhs = false)
+function fun_to_vec(kw::Expr; nolhs = false, gensym_names = false)
```
bkamins, line 103:
```suggestion
function fun_to_vec(kw::Expr; nolhs::Bool=false, gensym_names::Bool=false)
```

src/DataFramesMeta.jl
```diff
@@ -141,7 +150,7 @@ function fun_to_vec(kw::Expr; nolhs = false)
     end
 end
 
-fun_to_vec(kw::QuoteNode; nolhs = false) = kw
+fun_to_vec(kw::QuoteNode; nolhs = false, gensym_names = false) = kw
```
bkamins, line 153: same here

src/DataFramesMeta.jl
```diff
@@ -295,20 +304,40 @@ end
...
+    quote
+        $where($x, $(t...))
+    end
+end
+
+and(x, y) = x .& y
 
-and(x, y) = :($x .& $y)
+function df_to_bool(res)
+    if any(t -> !(t isa Union{Vector{<:Union{Missing, Bool}}, BitArray{1}}), eachcol(res))
```
bkamins, line 317: why `AbstractVector` is not allowed? E.g. from a `SubDataFrame`? Is this tested?

src/DataFramesMeta.jl
```diff
@@ -295,20 +304,40 @@ end
...
+
+function where(df::AbstractDataFrame, @nospecialize(args...))
+    res = DataFrames.select(df, args...; copycols = false)
+    tokeep = df_to_bool(res)
+    df[tokeep, :]
+end
 
-function where_helper(d, args...)
-    :($where($d, $(with_anonymous(reduce(and, args)))))
+function where(gd::GroupedDataFrame, @nospecialize(args...))
```
bkamins, line 331: why is this defined?

src/DataFramesMeta.jl
```diff
@@ -295,20 +304,40 @@ end
...
+    res = DataFrames.select(gd, args...; copycols = false, keepkeys = false)
+    tokeep = df_to_bool(res)
+    parent(gd)[tokeep, :]
 end
 
 """
     @where(d, i...)
 
-Select row subsets in AbstractDataFrames or groups in GroupedDataFrames.
+Select row subsets in AbstractDataFrames and GroupedDataFrames.
```
bkamins, line 340: only data frames

src/DataFramesMeta.jl
```diff
@@ -346,48 +386,43 @@ julia> @where(df, :x .> x, :y .== 3)
...
-6×2 SubDataFrame{Array{Int64,1}}
-│ Row │ n  │ x │
-├─────┼────┼───┤
-│ 1   │ 8  │ 2 │
-│ 2   │ 11 │ 2 │
-│ 3   │ 14 │ 2 │
-│ 4   │ 15 │ 2 │
-│ 5   │ 16 │ 2 │
-│ 6   │ 20 │ 2 │
+julia> @where(g, :n .> mean(:n))
```
bkamins, line 391: you still have examples with `gdf`.

src/DataFramesMeta.jl
pdeffebach: I thought we decided that we wanted `where` to work on a GroupedDataFrame.

bkamins: yes - I mixed it up with `orderby`

src/DataFramesMeta.jl
pdeffebach: Thanks for this. I added a method for `SubDataFrame` for both `@where` and `@orderby` because of this. (needs a new method because `select(subdf, ..., copycols = false)` fails.

pdeffebach:
Thanks! All your commends should be addressed.

src/DataFramesMeta.jl
bkamins: it is expected to fail if `copycols=false` only column reusing and then `SubDataFrame` is produced (no transformations) are allowed.

pdeffebach:
Another note, 

```
@where(gd[[1:2]], args...)
```

fails at the stage of `DataFrames.select`, which I think is fine.

bkamins:
> at the stage of `DataFrames.select`

As you know this is disallowed and intended :) if `gdf` has more than 2 groups. 😄.

Do you think we want to go back to the discussion if `select` on `gdf[[1:2]]` should be allowed?

pdeffebach:
No, I am happy with the current scenario. Maybe people will request that feature in the future, but I'm fine with whatever DataFrames does.

pdeffebach:
Tests pass on 1.0, fail on nightly for an unrelated reason. 

We should probably start testing 1.0 and 1.6-dev rather than nightly.

bkamins (approved):

src/DataFramesMeta.jl
```diff
@@ -295,20 +299,46 @@ end
...
+        $where($x, $(t...))
+    end
+end
+
+and(x, y) = x .& y
+
+function df_to_bool(res)
+    if any(t -> !(t isa Union{AbstractVector{<:Union{Missing, Bool}}, BitArray{1}}), eachcol(res))
+        throw(ArgumentError("All arguments in @where must return a " *
+                            "AbstractVector{<:Union{Missing, Bool} or a BitArray{1}"))
```
nalimilan, line 314: `BitArray{1} <: AbstractVector{<:Union{Missing, Bool}` already:
```suggestion
    if any(t -> !(t isa AbstractVector{<:Union{Missing, Bool}}), eachcol(res))
        throw(ArgumentError("All arguments in @where must return an " *
                            "AbstractVector{<:Union{Missing, Bool}}"))
```

src/DataFramesMeta.jl
```diff
@@ -295,20 +299,46 @@ end
...
+function where_helper(x, args...)
+    t = (fun_to_vec(arg; nolhs = true, gensym_names = true) for arg in args)
+    quote
+        $where($x, $(t...))
+    end
+end
+
+and(x, y) = x .& y
+
+function df_to_bool(res)
```
nalimilan, line 311:
```suggestion
function df_to_bool(res::AbstractDataFrame)
```

src/DataFramesMeta.jl
```diff
@@ -295,20 +299,46 @@ end
...
+    if any(t -> !(t isa Union{AbstractVector{<:Union{Missing, Bool}}, BitArray{1}}), eachcol(res))
+        throw(ArgumentError("All arguments in @where must return a " *
+                            "AbstractVector{<:Union{Missing, Bool} or a BitArray{1}"))
+    end
+
+    return reduce(and, eachcol(res)) .=== true
```
nalimilan, line 317: Why not this?
```suggestion
    return reduce((x, y) -> x .& y, eachcol(res)) .=== true
```

src/DataFramesMeta.jl
```diff
@@ -295,20 +299,46 @@ end
...
+    res = DataFrames.select(df, args...)
+    tokeep = df_to_bool(res)
+    df[tokeep, :]
 end
 
 """
     @where(d, i...)
 
-Select row subsets in AbstractDataFrames or groups in GroupedDataFrames.
+Select row subsets in AbstractDataFrames and GroupedDataFrames.
```
nalimilan, line 341:
```suggestion
Select row subsets in `AbstractDataFrame`s and `GroupedDataFrame`s.
```

src/DataFramesMeta.jl
```diff
@@ -317,6 +347,17 @@ Select row subsets in AbstractDataFrames or groups in GroupedDataFrames.
 
 Multiple `i` expressions are "and-ed" together.
 
+If given a `GroupedDataFrame`, `@where` applies transformations by
+group, but does not store the result. `@where` returns a fresh
+`DataFrame` where the rows generated from the transformations are all
+`true`.
```
nalimilan, line 353:
```suggestion
If given a `GroupedDataFrame`, `@where` applies transformations by
group, and returns a fresh `DataFrame` containing the rows
for which the generated values are all `true`.
```

test/grouping.jl
```diff
@@ -291,4 +287,19 @@ end
...
+        t = ["a", "b", "c", "c", "e"],
+        y = [:v, :w, :x, :y, :z],
+        c = [:g, :quote, :body, :transform, missing]
+    )
+
+    gd = groupby(df, :g)
+
+    @test @where(gd, :i .== first(:i)).t == ["a", "c"]
+    @test @where(gd, cols(:i) .> mean(cols(:i)), :t .== "c").t == ["c"]
+    @test @where(gd, :c == :g).i == Int[]
```
nalimilan, line 303: As usual I'd recommend testing equality against a full data frame, to ensure everything is correct and not just one column.

src/DataFramesMeta.jl
```diff
@@ -413,6 +449,11 @@ function orderby(x::GroupedDataFrame, @nospecialize(args...))
     throw(ArgumentError("@orderby with a GroupedDataFrame is reserved"))
 end
 
+function orderby(x::SubDataFrame, @nospecialize(args...))
+    t = DataFrames.select(x, args...)
+    x[sortperm(t), :]
+end
```
nalimilan, line 455: Why is this in this PR?

test/dataframes.jl
```diff
@@ -286,11 +300,15 @@ end
...
-
     @test @orderby(df, :c).i == [3, 1, 2, 4, 5]
     @test @orderby(df, -:g).i == [4, 5, 1, 2, 3]
     @test @orderby(df, :t).i == [1, 2, 3, 4, 5]
+
+    @test @orderby(df, identity(:g), :g.^2).i == [1, 2, 3, 4, 5]
+
+    subdf = @view df[1:3, :]
+
+    @test @orderby(subdf, -:i).i == [3, 2, 1]
```
nalimilan, line 311: Same remark here.

src/DataFramesMeta.jl
pdeffebach: I am still getting a grip on how anonymous functions change compilation time. You are right that this doesn't matter.

src/DataFramesMeta.jl
pdeffebach: Because its a bug that Bogumil noticed in the implementation of `@where` that we hadn't caught in the implementation of `@orderby`. I can delete it if you want to keep things separate.

pdeffebach:
Thanks for the review. I believe I have addressed all your comments. 

Sorry about tacking on the `@orderby` fix onto this PR. It seemed natural since it's the same implementation. I can split those changes into another PR if needed.

nalimilan (approved):

pdeffebach:
Thanks!

PR #195, opened 2020-10-22
Nospecialize

pdeffebach:
Before this PR, *not* on DataFrames master, commands take on average `.3` seconds. 

After this PR and on DataFrames master, `.08` seconds. 

Still a bit of a frustrating latency, but I think this is lots of progress and an acceptable latency!


pdeffebach:
This needs benchmarking. I will revive the benchmark script in the `test` folder.

bkamins:
nice. Please let me know when it will be good to have a look at.

pdeffebach:
Just wrote a quick performance-testing script. Things look good!

```
DataFrames benchmark timings
  193.431 ms (2001075 allocations: 240.56 MiB)
DataFramesMeta benchmark, timings
  185.151 ms (2001358 allocations: 248.20 MiB)

DataFrames raw timing
  0.210350 seconds (2.00 M allocations: 240.558 MiB, 14.87% gc time)
DataFramesMeta raw timing
  0.185518 seconds (2.00 M allocations: 240.558 MiB, 6.17% gc time)
```

pdeffebach:
@bkamins @nalimilan this is ready for a review

test/performance.jl
```diff
@@ -0,0 +1,103 @@
...
+@btime DataFrames_timings($df, $gd);
+println("DataFramesMeta benchmark, timings")
+@btime DataFramesMeta_timings($df, $gd);
+
+println()
+
+println("DataFrames raw timing")
+@time DataFrames_timings(df, gd);
+println("DataFramesMeta raw timing")
+@time DataFrames_timings(df, gd);
```
bkamins, line 101: you probably have a typo here.

test/performance.jl
```diff
@@ -0,0 +1,103 @@
...
+		:v2 => mean => :res8,
+		[:v1, :v2] => ((t, s) -> std(t) + std(s)) => :res9,
+		[:v1, :v2, :v3] => complicated_scalar => :res10,
+		[:v1, :v2, :v3] => ((a, b, c)  -> first(a) + mean(b) * std(a) + last(c)) => :res11
+	)
+
+	return(df_res, gd_res)
+end
+
+function DataFramesMeta_timings(df, gd)
```
bkamins, line 70: why this should be faster than the `DataFrames_timings`?

pdeffebach: They aren't different on average, the `@btime` estimates are somewhat noisy.

test/performance.jl
pdeffebach: Sorry. Just re-ran and the `@time` timings are nonetheless the same.

src/DataFramesMeta.jl
```diff
@@ -104,44 +104,67 @@ function fun_to_vec(kw::Expr; nolhs::Bool = false, gensym_names::Bool = false)
...
+            else
+                # act on g(:x)
+                body = replace_syms!(kw.args[2], membernames)
+            end
 
-        source = Expr(:vect, keys(membernames)...)
+            source = Expr(:vect, keys(membernames)...)
+            inputargs = Expr(:tuple, values(membernames)...)
+            fun = quote
+                (@nospecialize args...) -> begin
```
bkamins, line 133: but if this is the case and we removed the `@nospecialize` here how would the benchmarks be affected?

pdeffebach: They are completely unchanged. 

I think it would be useful to try and find scenarios where they *are* affected. I will do that this weekend.

bkamins: That is what I was wanting to understand 😄

pdeffebach:
I have uncovered how we get a performance penalty. You get problems with 

```

function DataFrames_timings(df, gd)
	df_res = transform(df,
		:v1 => (t -> t .- mean(t)) => :res1,
		:v2 => demean => :res2,
		[:v1, :v2] => (+) => :res3,
		:id4 => string => :res4,
		[:v1, :v2, :v3] => complicated_vec => :res5,
		[:v1, :v2, :v3] => ((a, b, c) -> @. a + b * c * c + a) => :res6a,
		[:v1, :v2, :v3] =>
		(
			(a, b, c) -> begin
				d = Vector{Float64}(undef, length(a))
				for i in eachindex(d)
					d[i] = a[i] + b[i] * c[i] * c[i] + a[i]
				end
				d
			end
		) => :res6b
	)

	gd_res = combine(gd,
		:v1 => (t -> mean(t)) => :res7,
		:v2 => mean => :res8,
		[:v1, :v2] => ((t, s) -> std(t) + std(s)) => :res9,
		[:v1, :v2, :v3] => complicated_scalar => :res10,
		[:v1, :v2, :v3] => ((a, b, c)  -> first(a) + mean(b) * std(a) + last(c)) => :res11,
	)

	return(df_res, gd_res)
end

function DataFramesMeta_timings(df, gd)
	df_res = @transform(df,
		res1 = :v1 .- mean(:v1),
		res2 = demean(:v2),
		res3 = :v1 + :v2,
		res4 = string(:id4),
		res5 = complicated_vec(:v1, :v2, :v3),
		res6a = @.(:v1 + :v2 + :v3 * :v3 + :v1),
		res6b = begin
			d = Vector{Float64}(undef, length(:v1))
			for i in eachindex(d)
				d[i] = :v1[i] + :v2[i] * :v3[i] * :v3[i] + :v1[i]
			end
			return d
		end
	)


	gd_res = @based_on(gd,
		res7 = mean(:v1),
		res8 = (t -> mean(t))(:v2),
		res9 = std(:v1) + std(:v2),
		res10 = complicated_scalar(:v1, :v2, :v3),
		res11 = first(:v1) + mean(:v2) * std(:v1) + last(:v3)
	)

	return(df_res, gd_res)
end
```

`:res6b` is the problem. In DataFramesMeta, we never have a function barrier, so julia can't compile efficient code in the `for` loop. With `@nospecialize`. The options are

1. `@nospecialize`
```
            fun = quote
                (@nospecialize args...) -> begin
                    $inputargs = args
                    $body
                end
            end
```

Results: 

```
DataFrames benchmark timings
  189.276 ms (2001104 allocations: 248.19 MiB)
DataFramesMeta benchmark, timings
  419.528 ms (10999330 allocations: 393.12 MiB)
DataFrames raw timing
  0.260386 seconds (2.00 M allocations: 248.189 MiB)
DataFramesMeta raw timing
  0.592254 seconds (11.00 M allocations: 393.124 MiB, 30.86% gc time)
```

2. Just get rid of the macro 
```
            fun = quote
                (args...) -> begin
                    $inputargs = args
                    $body
                end
            end
```

Results (oddly bad)
```
DataFrames benchmark timings
  205.529 ms (2001135 allocations: 248.19 MiB)
DataFramesMeta benchmark, timings
  391.286 ms (10999330 allocations: 393.12 MiB)
DataFrames raw timing
  0.334949 seconds (2.00 M allocations: 248.189 MiB, 43.00% gc time)
DataFramesMeta raw timing
  0.476538 seconds (11.00 M allocations: 393.124 MiB, 17.27% gc time)
```

3. Define function without `vararg`
```
            fun = quote
                 $inputargs -> begin
                    $body
                end
            end
```

Results:

```
DataFrames benchmark timings
  203.278 ms (2001135 allocations: 248.19 MiB)
DataFramesMeta benchmark, timings
  203.322 ms (2001250 allocations: 255.82 MiB)
DataFrames raw timing
  0.194473 seconds (2.00 M allocations: 248.190 MiB, 7.60% gc time)
DataFramesMeta raw timing
  0.240795 seconds (2.00 M allocations: 255.822 MiB, 16.85% gc time)
```

All results are on nightly. On 1.5 there is a performance hit with `@nospecialize` that doesn't appear in 1.6. 

Is this a worthwhile tradeoff? Can we encourage users not to write loops? 

In particular, I don't think there is a way to, inside the `begin` block, specialize on the type of arguments again. Maybe one of you knows if there is an "easy fix" for users who want to opt in to performance.

bkamins:
> Can we encourage users not to write loops?

I assume that with DataFramesMeta.jl we can assume users will not write loops (and can put a warning about it).
If someone wants to write a loop we should recommend - just define a function with the loop and later call it.

pdeffebach:
The loop is just one example, of course. I will have to think  of other non-function operations people do. 

If there isn't anything easy to do in the body of the expression but we still want easy use, we can also set flags, like how `@byrow` will be implemented, for users to easily opt-in to faster code.

nalimilan:
Yes I think we can consider that DataFramesMeta users shouldn't write loops (or they should put them in functions). As @bkamins said we can detect this and print warnings anyway. So I would drop these from the benchmarks are they probably make a big difference.

pdeffebach:
I've added tests for the `:x .* :y` case, which we optimize in 1.6 but not before due to https://github.com/JuliaLang/julia/pull/37583

I also changed performance.jl to include a reasonable "fix" for writing a loop inside a `@transform` call. 

This is ready for a review / merging.

pdeffebach:
I filed an issue with julia here: https://github.com/JuliaLang/julia/issues/38176

It seems like `.*` etc. is still not working. This optimization might have to wait.

pdeffebach:
Okay I have purged everything related to the function re-use optimization. 

I have a strategy now to make it work but it's a lot of LOC and I would like to get a release out soon. Since the optimization relies on 1.6 behavior it doesn't make sense to push for it now. 

Ready for a review.

pdeffebach:
Tests pass! On both 1.0 and nightly.

bkamins (approved):

nalimilan:
What benchmark results should I look at to see what they give with `res6b` using an anonymous function?

pdeffebach:
If you mean how to configure

1. DataFrames `master`
2. Julia nightly
3. The results using `@btime`, not `@time`. 

I don't do a great job showing the improvements in compilation time but it's a bit tough to measure.

nalimilan:
Sorry, I wondered whether the results you posted above were still relevant, but I got my reply. Have you run these on master on the same machine for the comparison?

pdeffebach:
Okay I have updated the script to better estimate the changes in raw timings. 

Rather than estimate vs `master` I am estimating with and without `@nospecialize`. We have 

1. On nightly
2. On DataFrames master

Without `@nospecialize`

```
julia> include("test/performance.jl")
DataFrames benchmark timings
  180.904 ms (2001291 allocations: 248.20 MiB)
DataFramesMeta benchmark, timings
  196.559 ms (2001447 allocations: 255.83 MiB)
DataFramesMeta raw timing
  0.597592 seconds (523.12 k allocations: 32.467 MiB, 99.20% compilation time)
Anonymous function without `@nospecialize`
  0.056371 seconds (32.12 k allocations: 9.080 MiB, 90.52% compilation time)
Anonymous function with `@nospecialize`
  0.012096 seconds (7.74 k allocations: 8.117 MiB, 55.92% compilation time)
```

*with* `@nospecialize`

```
julia> include("test/performance.jl")
DataFrames benchmark timings
  176.683 ms (2001291 allocations: 248.20 MiB)
DataFramesMeta benchmark, timings
  185.638 ms (2001585 allocations: 255.84 MiB)
DataFramesMeta raw timing
  0.538601 seconds (523.57 k allocations: 33.295 MiB, 99.24% compilation time)
Anonymous function without `@nospecialize`
  0.055212 seconds (32.12 k allocations: 9.080 MiB, 90.50% compilation time)
Anonymous function with `@nospecialize`
  0.011967 seconds (7.75 k allocations: 8.117 MiB, 55.84% compilation time)
```

1. Looking at the difference between the "benchmark" timings --- the difference with specialize and the difference without it ---, there is a small performance hit for DataFramesMeta compared to DataFrames. Given that this decreases due to `@nospecialize`, this is probably due to improvements in compilation time, which I don't fully understand. 
2. Comparing the `DataFramesMeta raw timing` with and without `@nospecialize` we see that `@nospecialize` gets us about `.06` seconds improvement in latency. 

**However** It's really hard to pin down this number. According to the anonymous function timings we should be seeing about `.03` seconds improvement per call. But that's not quite what we have. That's actually what I get when I compare a single `@select` call to another.

Honestly it's very hard to benchmark this. I initially thought there was a bigger improvement than there seems to be after benchmarking.

pdeffebach:
Here is a non-reproducible benchmark, but it shows why I think this change is worth it:

Screenshot from 2020-10-27 19-55-40

pdeffebach:
I just benchmarked some more .

`@nospecialize` *only* gives a benefit for the `@select` command in the screenshot above. 

For other, simpler, commands, i.e. `complicated_vec(:v1, :v2, :v3)`, the compilation time is the same. 

If someone could try out this benchmark script, I would really appreciate it. I posted on slack for users. If the modal DataFramesmeta call is very complicated, then this is worth it. But if not, there isn;t a huge benefit. 

We don't need to have this improvement immediately, especially since it only works well on DataFrames `master`. It's up to you two if you think it ought to be merged.

nalimilan:
OK. Maybe better hold this off until we have a solution with a more general impact?

pdeffebach:
Closed in favor of #196

PR #203, opened 2020-11-11
Add real docs

pdeffebach:
This is a major PR (it felt like the last one was, too!)

I am getting us set up with "real" documentation. 

1. Transferred the `README` to `index.md`
1. Added an `api.md` with the `@doc` macros 
1. Fixed all the `jldoctests` so they run

I still need to

1. Write a new `README`
2. Figure out how to get set up so it's a website

Also, I can't get the link to the `api` page to work from the `index` page. If anyone knows how to fix this, please let me know!

cc @bkamins, @nalimilan


bkamins:
Great! Let me know when you are done and I will have a look (though @nalimilan is much better than me with documentation).

pdeffebach:
Okay I've taken this PR as far as I'm comfortable with for now. 

The changes to the docs are minimal. But lots of examples have been changed to no longer rely on random number generation. 

Next step is hosting. If this were my own package I would fumble through the hosting configuration on my own. But given this is owned by JuliaData I would like someone, probably @quinnj  or @nalimilan to help take this over the finish line. in particular, the part about adding an ssh key (which I have, I think) and making sure a website actually *exists* confuses me. 

Thank you!

quinnj:
Usually it's not too bad to setup; I just follow the Documenter.jl guide for hosting; in travis, you put the DOCUMENTER_KEY as an environment variable (I can do this if you dm me the key), then make sure your travis job has a docs build job.

pdeffebach:
@quinnj or @nalimilan 

I've worked through the documenter docs on hosting, and i need to add a ssh key for the repository. I don't have ownership permissions to do that for DataFramesMeta (can't click on settings at the repo page on github). I need one of you to work through [this section](https://juliadocs.github.io/Documenter.jl/stable/man/hosting/#travis-ssh) to set up the SSH deploy keys.

pdeffebach:
Okay I think this is ready for another review. I have 

* Added the ssh key to the Travis settings
* Added the ssh key to the DataFramesMeta settings
* Added the deployment ssh command to the tagbot yml
* Deleted an extraneous appveyer badge
* Added badges for the documentation
* Written a little more on the README

I think that if we merge this the docs should deploy correctly. 

Please let me know!

docs/src/api/api.md
```diff
@@ -0,0 +1,34 @@
+# API
+
+```@docs
+@transform(x, args...)
```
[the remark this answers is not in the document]
pdeffebach: Thanks! didn't know about `@autodocs`. It's probably good to document deprecated things anyways so people can see them in a search.

src/DataFramesMeta.jl
```diff
@@ -596,19 +599,19 @@ function combine_helper(x, args...; deprecation_warning = false)
...
 
 """
-    @combine(g, i...)
+    @combine(x, args...)
 
 Summarize a grouping operation
 
 ### Arguments
 
 * `g` : a GroupedDataFrame
```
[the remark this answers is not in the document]
pdeffebach: thanks. added.

src/DataFramesMeta.jl
```diff
@@ -618,34 +621,13 @@ julia> g = groupby(d, :x);
 
 julia> @combine(g, nsum = sum(:n))
 3×2 DataFrame
-│ Row │ x │ nsum │
-├─────┼───┼──────┤
-│ 1   │ 1 │ 99   │
-│ 2   │ 2 │ 84   │
-│ 3   │ 3 │ 27   │
-
-julia> @combine(g, x2 = 2 * :x, nsum = sum(:n))
```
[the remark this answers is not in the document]
pdeffebach: Yes, it shows the "spreading". re-added.

pdeffebach:
Tests pass!

docs/src/index.md
```diff
@@ -0,0 +1,457 @@
...
+These macros improve performance and provide more convenient syntax.
+
+DataFrames.jl has the functions `select`, `transform`, and `combine` 
+for manipulating data frames. DataFramesMeta provides the macros 
+`@select`, `@transform`, and `@combine` to mirror these functions with 
+more convenient syntax. Inspired by [dplyr](https://dplyr.tidyverse.org/) in R 
+and [LINQ](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/linq/)
+in C#. 
+
+In addition, DataFramesMeta provides
```
bkamins, line 13:
```suggestion
In addition, DataFramesMeta.jl provides 
```

bkamins: same in other places

docs/src/index.md
```diff
@@ -0,0 +1,457 @@
+# Introduction
+
+Metaprogramming tools for DataFrames.jl objects.
+These macros improve performance and provide more convenient syntax.
```
bkamins, line 4: maybe add 1-2 examples of DataFrames.jl code and equivalent DataFramesMeta.jl code and say that they show this convenience.

Also maybe comment in what cases the macros improve performance somewhere (probably not here)

docs/src/index.md
```diff
@@ -0,0 +1,457 @@
...
+    The above function will return `[4, 5, 6]` because the `return` inside the `@with`
+    applies to the anonymous function created by `@with`. 
+
+    Given that `@eachrow` (below) is implemented with `@with`, the same caveat applies to 
+    `@eachrow` blocks. 
+
+
+## `@eachrow`
+
+Act on each row of a data frame. Includes support for control flow and `begin end` blocks. Since the "environment" induced by `@eachrow df` is implicitly a single row of `df`, one uses regular operators and comparisons instead of their elementwise counterparts as in `@with`. Does not change the input data frame argument.
```
bkamins, line 197: add line breaks

docs/src/index.md
```diff
@@ -0,0 +1,457 @@
...
+            x += :B * :C
+        end
+    end
+    x
+end
+```
+
+`@eachrow` also supports special syntax for allocating new columns to make
+`@eachrow` more useful for data transformations. The syntax `@newcol
+x::Array{Int}` allocates a new column `:x` with an `Array` container with eltype
```
bkamins, line 223: This is the first time I notice it. Why `Array{Int}` not `Vector{Int}`? Also can you allocate other types than vectors from Julia Base (e.g. `PooledVector`)?

docs/src/index.md
```diff
@@ -0,0 +1,457 @@
...
+    orderby(:meanX) |>
+    select(:meanX, :meanY, var = :b)
+```
+
+Relative to the use of individual macros, chaining looks cleaner and
+more obvious with less noise from `@` symbols. This approach also
+avoids filling up the limited macro name space. The main downside is
+that more magic happens under the hood.
+
+This method is extensible. Here is a comparison of the macro and
```
bkamins, line 389: is this comparison or an example how one could write ones own extension?

docs/src/index.md
```diff
@@ -0,0 +1,457 @@
...
+```
+
+The `linq` method above registers the expression-replacement method
+defined for all `with()` calls. It should return an expression like a
+macro.
+
+Again, this is experimental. Based on feedback, we may decide to only
+use `@linq` or only support the set of linq-like macros.
+
+Alternatively you can use Lazy.jl `@>` macro like this:
```
bkamins, line 410: do we really need to show Lazy.jl here? What is the value of this?

docs/src/index.md
```diff
@@ -0,0 +1,457 @@
...
+    @select(:meanX, :meanY, var = :b)
+end
+```
+
+Please note that Lazy.jl exports the function `groupby` which would clash
+with `DataFrames.groupby`. Hence, it is recommended that you only import a
+select number of functions into the namespace by only importing `@>` e.g. 
+`using Lazy: @>` instead of `using Lazy`.
+
+Another alternative is Pipe.jl which exports the `@pipe` macro for piping.
```
bkamins, line 434: same with Pipe.jl. I like and use it with DataFrames.jl but what is the benefit here? Do we recommend it?

bkamins:
Looks good. Thank you. I have left some minor comments.

docs/src/index.md
pdeffebach: You can allocate any type that would allow this pattern 

```
T(undef, N)
```

So it would probably be pretty hard to construct other types of arrays, I think. 

This is something I would really like to fix but haven't thought of a way to do so without type instabilities.

bkamins: OK - clear. We can leave it for later. Now we document that `Array` is only allowed, so we have flexibility for the future.

docs/src/index.md
pdeffebach: I agree. I'm fine to delete.

docs/src/index.md
pdeffebach: I don't think I want to expose this API formally yet. Its untested, and I don't know if there is a need to someone to write their own extension. I think we should cut this.

docs/src/index.md
pdeffebach: Actually, I think I will keep the piping sections. `@>` and `@pipe` have different behaviors, so it makes sense for users to choose what they are most comfortable with, for now. 

Eventually I would like DataFramesMeta to provide one really good macro and not recommend others, but that hasn't happened yet.

docs/src/index.md
pdeffebach: Added an example of the convenience of DataFramesMeta. 

I dont think any of them improve performance over DataFrames. I will defer writing this section and delete the claim at the top of the document.

docs/src/index.md
pdeffebach: I have changed all examples to array.

docs/src/index.md
pdeffebach: Trying to only use the `.jl` in introductions, and just the package name after.

pdeffebach:
Thanks!

I deleted the example about extending `@linq` but am keeping other piping. `@linq` isn't perfect so users might want to know about other options. 

Ready to merge if people approve!

docs/src/index.md
bkamins: OK - we can go this way if you prefer

docs/src/index.md
bkamins: OK - then I would cut this.

bkamins (approved):

docs/make.jl
```diff
@@ -3,6 +3,10 @@ using Documenter, DataFramesMeta
 makedocs(
 	modules = [DataFramesMeta],
 	sitename = "DataFramesMeta Documentation",
+	format = Documenter.HTML(
+		canonical = "https://juliadata.github.io/DataFramesMeta.jl/stable/",
+		assets = ["assets/favicon.ico"]
```
nalimilan, line 8: I think you need to add the icon to docs/src/assets/ (you can copy it from DataFrames). Or just drop this line.

pdeffebach:
Thanks for your feedback, everyone! I am going to merge!

PR #205, opened 2020-11-15
up DataFrames.jl dependency

bkamins:
@pdeffebach - I guess DataFramesMeta.jl should work on 0.22 release of DataFrames.jl equally well. Right? How fare are you with the release of DataFramesMeta.jl? (it is holding back the update of the tutorials)

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -7,7 +7,7 @@ DataFrames = "a93c6f00-e57d-5684-b7b6-d8193f3e46c0"
 Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
 
 [compat]
-DataFrames = "0.21"
+DataFrames = "0.21, 0.22"
 Reexport = "0.2"
 julia = "1"
 


bkamins:
It seems that `@combine` still passes a `Pair` as a first argument to DataFrames.jl `combine`.

pdeffebach:
This is tough. To make the smallest change for this I would need to add `=> AsTable`. but this will fail for `0.21`. 

Would it be possible to add an explicit version check for this one bit of behavior? Add a `=> asTable` if DataFrames is on `0.22` or above, and leave it as is otherwise?

bkamins:
> Would it be possible to add an explicit version check for this one bit of behaviour?

Sure - I think it is a standard practice. As you know `if` does not introduce its own scope so you can even to in top level a check of version and define two different methods/macros depending on the version (and then you would support both 0.21 and 0.22).

pdeffebach:
Do you know the syntax for this? It should be 

```
if version(DataFrames) < v"0.22" 
   ...
else 
   ...
end
```

bkamins:
I think the canonical way to do it now is:
```
Pkg.dependencies()[Base.UUID("a93c6f00-e57d-5684-b7b6-d8193f3e46c0")].version < v"0.22"
```
but maye @KristofferC can propose something better (I have not checked in which Julia version `Pkg.dependencies` was introduced)

You can alternatively check if `DataFrames.rownumber` is defined as we added it in 0.22 if I remember correctly.

pdeffebach:
Lots of broken tests in the "undefined behavior" test sets. Thankfully since it's in those test sets I can add if-else without worrying about breakin.

bkamins:
:smile: - hopefully in the new release the "undefined behavior" is just an error - right?

pdeffebach:
Yes more and more things are becoming errors

PR #206, opened 2020-11-24
Zero 22 support

pdeffebach:
Support for DataFrames 0.22

Supersedes #205


bkamins (approved):

pdeffebach:
Tests pass with 0.22 on Travis (and locally), tests pass on 0.21 locally. 

If someone confirms this hacky fix is performant, I will merge. The fix is to define 

```
global const DATAFRAMES_GEQ_22 = isdefined(DataFrames, :pretty_table) ? true : false
```

and then branch based on that value in the necessary functions.

bkamins:
Yes - I have checked this fix when I approved and it is OK IMO.

PR #208, opened 2020-11-25
Host documentation on GH Actions

pdeffebach:


pdeffebach:
Given that we already have a `DOCUMENTER_KEY` set up, I think this should work when merged into master? 

I'm not sure, to be honest.

pdeffebach:
cc @nalimilan or @quinnj, or anyone who knows how github actions work. 

This PR is a bit time-sensitive, I guess since people are already asking why the docs are down.

nalimilan (approved):
Look reasonable. I don't know for sure, but if you merge you'll find out. :-)

nalimilan:
Also, I was suggesting to move all CI to GitHub since Travis doesn't work at all now.

pdeffebach:
Yes, that is definitely the next step.

Issue #218, opened 2021-01-07
@transform! does not exist?


dmetivie:
I have DataFramesMeta v0.6.0 and I cannot use the macro `@transform!` (I get: `ERROR: LoadError: UndefVarError: @transform! not defined`). Is it a problem with my version? The macro `@transform` works well.

pdeffebach:
Sorry! This is my bad. 

I just merged the PR for `@transform!` two days ago (#216). This PR included changes to the readme, which always shows the state of the `master` branch. So the readme contains things which do not exist. 

This was a mistake on my part. I should have not merged changes to the readme until we actually make a new release. 

To use `@select!` and `@transform!` you can add the `master` branch.

pdeffebach:
Fixed via #219

PR #219, opened 2021-01-07
delete @select! reference

pdeffebach:
In reference to #218. We shouldn't update the readme to discuss `@select!` and `@transform!` until we make a new release. 

I would rather wait for a new release until we have `@subset` and `@subset!` and maybe and `@orderby!`. 

cc @nalimilan

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,9 +10,7 @@ Metaprogramming tools for DataFrames.jl objects.
 # Macros 
 
 * `@transform`, for adding new columns to a data frame
-* `@transform!`, for adding new columns to a data frame *in-place*
 * `@select`, for selecting columns in a data frame
-* `@select!`, for selecting columns in a data frame *in-place*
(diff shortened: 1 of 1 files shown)


bkamins (approved):
Though I would remove the documentation from README.md entirely

pdeffebach:
I think the best solution is to take the current version of the readme and make it the docstring for the module. Then we can add to that as needed. Then before releases we just copy over the docstring as is to the actual readme.

nalimilan:
There's already a list of provided function in the homepage of the Documenter manual. So why not remove the list from the README completely? I don't think it's standard to have a docstring for a module listing all functions it exports: Documenter does that automatically already so adding a manual list which is likely to get outdated isn't great.

pdeffebach:
Okay, updated to delete the list of macros entirely. With an approval I will merge.

nalimilan (approved):

PR #221, opened 2021-02-13
avoid anonymous functions for simple function calls and broadcasts

jkrumbiegel:
This avoids the construction of anonymous functions when:
- there is a simple call like `f(args...)` where all args are QuoteNodes
- there is a broadcast call like `f.(args...)` where all args are QuoteNodes (from 1.6.0-beta onwards we use BroadcastFunction, before we use ByRow)

There is a workaround for calls like `:x .+ :y` which gives `:.+` as the call symbol, which we can't use directly. It is assumed that all such special operators, and only those, start with a dot, and we convert to `BroadcastFunction` by removing the leading dot from the symbol. That seemed sensible to me, but I don't know if there could be edge cases.


src/DataFramesMeta.jl
```diff
@@ -103,74 +103,134 @@ macro col(kw)
...
+        # and then use broadcast wrappers
+        if startswith(string(fun), ".")
+            f_sym_without_dot = Symbol(chop(string(fun), head = 1, tail = 0))
+            fun = broadcast_f_expr(f_sym_without_dot)
         end
 
+    # f.(:x, ...) into [:x, ...] => BroadcastFunction(f)
+    elseif is_simple_broadcast_call(function_expr)
+        # extract source symbols from quotenodes
+        source = [q.value for q in function_expr.args[2].args]
```
pdeffebach, line 171: I'm not sure I'm following this. In the other branches, `source` is an expression. Here it's an actual vector?

pdeffebach (commented):
Thanks a ton for this!

Can you add lots of tests? 

Additionally, I think te diff is kind of odd. It looks cleaner in my sublime text than it does in github for whatever reason.

src/DataFramesMeta.jl
jkrumbiegel: the diff is weird because I un-nested the whole if expression and put the guard at the top. I find it hard to read otherwise

jkrumbiegel: hm maybe I misread the other code with `source`, could be that you're right and it should also be an expression, I didn't check what `make_source_concrete` does. Although it's interesting that it still worked ;)

jkrumbiegel:
about the tests, I'm not sure what the best way to test is. Is the important part to check that second-time use doesn't allocate that much anymore, because it doesn't have to recompile? Or check transformation of expressions directly, by testing the internals.

pdeffebach:
For tests I would do two things. 

First, just add a battery of  `@transform` etc. calls to catch all the optimizations, `@transform(y = f(:x))` etc. 

Second, I would be satisfied with adding things at the end of `test/performance.jl` with `@time` commands, with and without anonymous functions. Then I can run the script myself and be satisfied that we are hitting the branches.

src/DataFramesMeta.jl
```diff
@@ -103,74 +103,134 @@ macro col(kw)
...
+function broadcast_f_expr(f)
+    # theoretically we could use ByRow everywhere, but maybe we can save some compilation
+    # here and there by using the Base functionality where it's available, as other code
+    # might have already called that
+    if VERSION >= v"1.6.0-beta"
+        :(Base.BroadcastFunction($f))
+    else
+        :(DataFrames.ByRow($f))
+    end
+end
```
nalimilan, line 134: @bkamins Do you think there would be any advantage to using `ByRow`? Maybe we will automatically use multithreading with `ByRow` in the future?

src/DataFramesMeta.jl
```diff
@@ -103,74 +103,134 @@ macro col(kw)
...
+    if is_simple_function_call(function_expr)
+        # extract source symbols from quotenodes
+        source = [q.value for q in function_expr.args[2:end]]
+        fun = function_expr.args[1]
+        # some normal-looking calls are actually broadcasts, like
+        # .+ .- etc., if we just pass on those symbols as function calls
+        # we get UndefVarErrors
+        # instead we transform the symbols to their non-broadcast versions
+        # and then use broadcast wrappers
+        if startswith(string(fun), ".")
```
nalimilan, line 163:
```suggestion
        if startswith(string(fun), '.')
```

src/DataFramesMeta.jl
bkamins: `ByRow` is and I think should be just a wrapper for broadcasting. If broadcasting uses multithreading in the future it will automatically make `ByRow` use it.

Though I think that compilation time savings are minimal, as `ByRow` is very cheap to compile.

pdeffebach:
This PR is looking good. No review yet, I need to re-organize the code some more.

pdeffebach:
This ended up being an enormous re-write of the entire `fun_to_vec` infrastructure. 

If people feel like the diff is too big (I created new files), then I can split this into another PR. 

Overall there might be a bit of redundancy with how things are handled, but tests pass and it's much more readable.

bkamins:
I understand that these changes are only internal and should not affect end-user experience (except performance). Is this correct?
If this is the case the question is if you believe we have an adequate test coverage of the functionality (as the answer to this question affects how I would approach to the review of this PR).

Thank you very much for working on this PR - this is a massive volume of changes indeed.

src/parsing.jl
```diff
@@ -0,0 +1,292 @@
...
+end
+
+function args_to_selectors(v)
+    t = map(v) do arg
+        if arg isa QuoteNode
+            arg
+        elseif onearg(arg, :cols)
+            arg.args[2]
+        else
+            Throw(ArgumentError("This path should not be reached, arg: $(arg)"))
```
pdeffebach, line 78: @bkamins @nalimilan 

I would appreciate some feedback on this line. Basically, this function should only get called if all the arguments are either QuoteNodes or `cols` (see `is_simple_broadcast_call`, above). 

But I feel like I'm repeating myself and this error seems heavy-handed. Let me know if there are existing design patterns I should use.

pdeffebach:
Tests pass! I would appreciate a review from @nalimilan, @bkamins, and @jkrumbiegel. 

I think the logic for everything should be a lot clearer now.

src/DataFramesMeta.jl
pdeffebach: It's all `ByRow` now.

src/DataFramesMeta.jl
pdeffebach: This was fixed.

src/parsing.jl
```diff
@@ -145,6 +145,8 @@ function fun_to_vec(ex::Expr; nolhs::Bool = false, gensym_names::Bool = false)
     # cols(:y) = f(cols(:x)) # re-write as simple call, RHS is block, use cols
     # cols(y) = :x + 1 # re-write as complicated col, but RHS is :block
     # cols(:y) = cols(:x) + 1 # re-write as complicated call, RHS is block, use cols
+    dump(ex)
+    asdf
```
bkamins: ?

pdeffebach: fixed. sorry.

pdeffebach: This is still ready for a review, though.

src/parsing.jl
bkamins: can you please give an example, so that I am sure what you want to disallow? Thank you!

nalimilan:
Could you split this into a commit that changes code and a commit that moves code to new files? Otherwise it's almost impossible to review, and in the future it will be super hard to know when lines have been introduced or modified.

pdeffebach:
@nalimilan, @bkamins 

Sorry about the crummy diff. I got ahead of myself. 

I have fixed the diff. It now only reflects the minimum changes necessary for the implementation of this PR. We can save the re-factoring for another PR.

 
An overview of the logic of `get_source_fun`

* If we see `@transform(df, z = f(:x, cols(:y)))`, we search through the RHS of the expression. If all the arguments are `QuoteNodes` or `cols(...)`, then
  * We collect all the arguments into a `Vector` `[:x, :y]`
  * We return `[:x, :y] => f => z`
* If we see `@transform(df, z = f(:x) .+ :y)` this is *not* a simple function call and we need to handle it via the traditional DataFramesMeta code. 


Edge cases: 

* `@transform(df, cols(:x))` This is valid `transform` syntax. But `cols(:x)` is an `Expr` and thus we must handle the case in our big `fun_to_vec` function. 
* Broadcasting, `.+` and `.*` are special-cased by the parser. We have to inspect the string of the function name. 
* Broadcasting, `f.(:x, cols(:y))` also looks different, but is handled. 
* In `@transform(df, cols(z) = f(:x))`, the RHS is actually an expression in a `block`. This also requires special handling. 


All these cases are listed in a long comment, which should maybe be a docstring. 

Very thankful for the review!

pdeffebach:
And tests pass. This is definitely ready for a review.

pdeffebach:
Bumping this. I would like to move forward with `@byrow` and can't until this is done.

bkamins (approved):

src/DataFramesMeta.jl
```diff
@@ -52,6 +53,82 @@ replace_syms!(e::Expr, membernames) =
...
+    :(DataFramesMeta.make_source_concrete($(Expr(:vect, t...))))
+end
+
+function get_source_fun(function_expr)
+
+    if is_simple_non_broadcast_call(function_expr)
+        source = args_to_selectors(function_expr.args[2:end])
+        fun_t = function_expr.args[1]
+
+        # .+ to +
```
bkamins, line 98: the only question is what would happen if you passed e.g. `".sum"` here?

pdeffebach: Thankfully it's a syntax error!

bkamins:
Why does CI fail?

bkamins:
Also you might consider updating CI to test latest Julia release not 1.5.

pdeffebach:
Tests didn't pass because we check the type of error thrown when `@eachrow` errors. 

However the fix was to change source code and not the tests. So it was good that we caught it.

bkamins:
OK. Nice.

pdeffebach:
@nalimilan if want to review, let me know. Otherwise I will merge.

nalimilan:
Sorry for the delay. I'll try to have a look soon.

src/DataFramesMeta.jl
```diff
@@ -103,78 +186,160 @@ macro col(kw)
...
+        source, fun = get_source_fun(rhs)
+        dest = lhs.args[2]
+
+        return quote
+            $source => $fun => $dest
+        end
     end
+
+    throw(ArgumentError("This path should not be reached"))
+    return nothing
```
nalimilan, line 340:
```suggestion
```

src/DataFramesMeta.jl
```diff
@@ -202,23 +367,17 @@ function replace_dotted!(e, membernames)
     Expr(:., x_new, y_new)
 end
 
+function exec(df, v, fun)
+    fun(map(c -> DataFramesMeta.getsinglecolumn(df, c), v)...)
+end
```
nalimilan, line 372: Any reason to put this in a separate function? AFAICT it's used only from `with_helper`.

Also:
```suggestion
exec(df, v, fun) = fun(map(c -> DataFramesMeta.getsinglecolumn(df, c), v)...)
```

src/DataFramesMeta.jl
```diff
@@ -52,6 +53,88 @@ replace_syms!(e::Expr, membernames) =
...
+            Throw(ArgumentError("This path should not be reached, arg: $(arg)"))
+        end
+    end
+
+    :(DataFramesMeta.make_source_concrete($(Expr(:vect, t...))))
+end
+
+function get_source_fun(function_expr)
+    # recursive step for begin :a + :b
+    # end
```
nalimilan, line 94:
```suggestion
    # recursive step for begin :a + :b end
```

src/DataFramesMeta.jl
```diff
@@ -52,6 +53,88 @@ replace_syms!(e::Expr, membernames) =
...
+            arg.args[2]
+        else
+            Throw(ArgumentError("This path should not be reached, arg: $(arg)"))
+        end
+    end
+
+    :(DataFramesMeta.make_source_concrete($(Expr(:vect, t...))))
+end
+
+function get_source_fun(function_expr)
```
nalimilan, line 92: Maybe add a comment explaining what this function does?

src/DataFramesMeta.jl
```diff
@@ -52,6 +53,88 @@ replace_syms!(e::Expr, membernames) =
...
+        && length(expr.args) == 2
+        && expr.args[1] isa Symbol
+        && expr.args[2] isa Expr
+        && expr.args[2].head == :tuple
+        && all(x -> x isa QuoteNode || onearg(x, :cols), expr.args[2].args))
+end
+
+function is_simple_function_call(x)
+    is_simple_non_broadcast_call(x) || is_simple_broadcast_call(x)
+end
```
nalimilan, line 76:
```suggestion
is_simple_function_call(x) = is_simple_non_broadcast_call(x) || is_simple_broadcast_call(x)
```

src/DataFramesMeta.jl
```diff
@@ -103,78 +186,160 @@ macro col(kw)
...
+            rhs = rhs_t.args[2]
+        else
+            rhs = rhs_t
         end
-        return t
-    elseif onearg(kw, :cols)
-        return kw.args[2]
     else
-        throw(ArgumentError("Expressions not of the form `y = f(:x)` currently disallowed."))
+        Throw(ArgumentError("This path should not be reached"))
```
nalimilan, line 262:
```suggestion
        throw(ArgumentError("This path should not be reached"))
```

src/DataFramesMeta.jl
```diff
@@ -103,78 +186,160 @@ macro col(kw)
...
+    end
+
+    # y = f(cols(:x))
+    # handled above
+
+    # y = :x + 1
+    # handled above
+
+    # y = cols(:x) + 1
+    # handled above
```
nalimilan, line 327: Why not mention it at the place where it's handled?

test/function_compilation.jl
```diff
@@ -0,0 +1,191 @@
+module TestFunctionCompilation
+
+using Test
+using DataFramesMeta
+
+@testset "function_compilation" begin
+	# Lazy way of making sure all
+	# functions are pre-compiled. @eval
+	# prevents julia from caching the
+	# intermediate anonymous functions.
```
nalimilan, line 10:
```suggestion
	# Lazy way of making sure all functions are pre-compiled.
	# @eval prevents Julia from caching the intermediate anonymous functions.
```

test/function_compilation.jl
```diff
@@ -0,0 +1,191 @@
...
+	for _ in 1:2
```
nalimilan, line 11: Why run this twice?

src/DataFramesMeta.jl
```diff
@@ -103,78 +186,160 @@ macro col(kw)
...
+    # :x
+    # handled above via dispatch
+
+    # cols(:x)
+    if onearg(ex, :cols)
+        return ex.args[2]
+    end
+
+    # The above cases are the only ones allowed
+    # if you don't have nolhs etc.
```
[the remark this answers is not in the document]
pdeffebach: clarified.

test/function_compilation.jl
pdeffebach: `transform(df, :x => (t -> some_stuff(t)) => :y)` should re-compiled every time, so the 2nd time should also be slow, and I wanted to ensure that.

test/function_compilation.jl
```diff
@@ -0,0 +1,191 @@
...
+
+			testfun(x, y) = x .* y
+			testdotfun(x, y) = x * y
+			testnt(x) = (c = x,)
+
+			@test @select(df, c = :a + :b) == DataFrame(c = [3])
+
+				fasttime = @timed @select(df, c = :a + :b)
+				slowtime = @timed select(df, [:a, :b] => ((a, b) -> a + b) => :c)
+				@test slowtime[2] > fasttime[2]
```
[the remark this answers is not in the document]
pdeffebach: I fixed the indentation and the tabs. 

However looking at `bytes` also errors sometimes in unpredictable ways. 

I wish there were a way to do a `@test` where it logs the result of the test but doesn't complain on Github if it doesn't pass. 

I also had to add an extra `@with` call because somehow it wasn't picking up the compilation. 

So not sure what the solution is.

src/DataFramesMeta.jl
pdeffebach: Having it in a separate function seems be the only way to get compilation to work. As in, I couldn't just construct an anonymous function and call it in the same scope. So I looked into DataFrames and saw how it does `_transform`.

src/DataFramesMeta.jl
```diff
@@ -103,78 +186,160 @@ macro col(kw)
...
+    end
 
-        source = Expr(:vect, keys(membernames)...)
-        inputargs = Expr(:tuple, values(membernames)...)
-        fun = quote
-            $inputargs -> begin
-                $body
-            end
+    # (; a = :x, ) # named tuple
+    if nokw && !(is_simple_function_call(ex))
```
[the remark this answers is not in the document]
pdeffebach: this was actually redundant since we do that check inside `get_source_fun`. 

I've taken care of all the redundant `if` statements.

src/DataFramesMeta.jl
pdeffebach: added a docstring.

pdeffebach:
Thank you for the detailed review!

Only outstanding issue is how to best do these tests.

test/function_compilation.jl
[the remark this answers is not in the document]
nalimilan: Well you can just do `slowtime[2] > fasttime[2] || @warn(...)`, right? :-)

src/DataFramesMeta.jl
nalimilan: OK. You mean because of the metaprogramming? Anyway I guess that's not a big deal.

src/DataFramesMeta.jl
```diff
@@ -103,78 +186,160 @@ macro col(kw)
...
+
+    # cols(:x)
+    if onearg(ex, :cols)
+        return ex.args[2]
+    end
+
+    # The above cases are the only ones allowed
+    # if you don't have nolhs etc.
+    if !(ex.head === :(=) || ex.head === :kw || nolhs)
+        throw(ArgumentError("Expressions not of the form `y = f(:x)` currently disallowed."))
```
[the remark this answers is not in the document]
pdeffebach: For `@select` and `@transform`, yes. But not for `@where` and `@combine` (where we return a Table). 

But I'm picky enough with calling `gensym_name=true` to make a `y = f(:x)` call in `@where`. and having `nokw=true` in `@combine` that the user should only see that error in `@select` and `@transform`.

src/DataFramesMeta.jl
pdeffebach: done.

test/function_compilation.jl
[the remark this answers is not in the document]
pdeffebach: added.

pdeffebach:
Okay I've fixed the test script. 

We can deal with the `"expressions not of the form y = f(:x) not allowed"` error later on, since it's tangential to this PR. 

I also want to require an explicit `AsTable` flag at some point that would make that error unnecessary. 

So this is ready for another review or merge.

src/DataFramesMeta.jl
```diff
@@ -82,19 +78,59 @@ function args_to_selectors(v)
...
     end
 
     :(DataFramesMeta.make_source_concrete($(Expr(:vect, t...))))
 end
 
+
+"""
+    get_source_fun(function_expr)
+
+Given an expression that may `QuoteNode`s (`:x`)
```
nalimilan, line 92:
```suggestion
Given an expression that may contain `QuoteNode`s (`:x`)
```

src/DataFramesMeta.jl
```diff
@@ -82,19 +78,59 @@ function args_to_selectors(v)
...
+
+julia> using MacroTools
+
+julia> ex = :(:x + :y)
+
+julia> DataFramesMeta.get_source_fun(ex)
+(:(DataFramesMeta.make_source_concrete([:x, :y])), :+)
+
+julia> ex = quote
+       :x .+1 .* :y
```
nalimilan, line 120:
```suggestion
           :x .+ 1 .* :y
```

src/DataFramesMeta.jl
```diff
@@ -218,38 +252,31 @@ function fun_to_vec(ex::Expr; nolhs::Bool = false, gensym_names::Bool = false)
     nokw = (ex.head !== :(=)) && (ex.head !== :kw) && nolhs
 
     # :x
-    # handled above via dispatch
+    # handled belos via dispatch on ::QuoteNode
```
nalimilan, line 255:
```suggestion
    # handled below via dispatch on ::QuoteNode
```

nalimilan:
> Only outstanding issue is how to best do these tests.

What's the plan then? Just print warnings in case of failure? How reliably do these tests pass?

pdeffebach:
> What's the plan then? Just print warnings in case of failure? How reliably do these tests pass?

Just print the warnings in case of failure. 

The pass very reliably on my local machine. But you are right that we don't know what kind of setting they will be run in on the cloud. So I guess we just have to be vigilant. I run tests locally a lot, so I will see those warnings if they happen.

pdeffebach:
All checks passed! Any failures were a small typo.

nalimilan (approved):

Issue #125, opened 2019-03-06
Rename `^` to `syms`


pdeffebach:
`^()`  is currently used to refer to a symbol on its own. for example, 

`df @> transform(x = :b)` will make a new column `:x` which looks like the column `:b`, while

`df @> transform(x = ^(:b))` makes a new column `:x` and fills it with the symbol `:b`. There is nothing about `^()` that indicates this behavior. I propose renaming it to `syms`, because it escapes `Symbol`s and has the same number of letters as `cols`.

nalimilan:
Doesn't JuliaDBMeta use `cols` for this? Cc: @piever

piever:
`cols` is used to recognize things that evaluate to a symbol, say::

```julia
a = :SepalLength
@with iris cols(a)
```

Whereas `^` is to escape something to be used as a symbol, for example `@map iris ^(:a)` will return a column full of the symbol `:a`.

I disagree with the renaming. `^` can escape a whole expression and not only a symbol (it basically takes it out of the macro and evaluates it as is, so the name `syms` is not completely correct. If anything, for consistency I'd be in favor of adding a infix operator version of `cols`, say `@with iris $a`, which is in some sense interpolation, in that `a` is evaluated before replacing symbols with columns.

nalimilan:
Got it. Yes, `$` for cols kind of makes sense. Maybe we could use `esc` for `^`?

(I still kind of wish we treated all names as columns by default, so that we wouldn't need to escape symbols...)

pdeffebach:
> Got it. Yes, `$` for cols kind of makes sense. Maybe we could use `esc` for `^`?

This is okay, but `cols` also "escapes" the input by evaluating what's inside. 

```
a = :x 
@transform(df, newcol = cols(a)
```
 which is the same as 

```
@transform(df, newcol = :x) # the column `:x` in df. 
```

Basically we have two divergent ways to escape the variable `a`. 

> (I still kind of wish we treated all names as columns by default, so that we wouldn't need to escape symbols...)

Don't we do this already? Or do you mean full tidy evaluation?

nalimilan:
So what's the conclusion here?

pdeffebach:
It's weird that we have `cols` instead of `col`... But we should rename this to `syms` for consistency.

pdeffebach:
Closing this, `^()` is fine, and now that we use `$` intead of `cols` it doesn't make sense to have `syms`.

PR #228, opened 2021-03-07
use `syms` instead of `^()`

pdeffebach:
Addresses #125

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -179,7 +179,7 @@ x = @with df begin
     res
 end
 
-@with(df, df[:x .> 1, ^(:y)]) # The ^ means leave the :y alone
+@with(df, df[:x .> 1, syms(:y)]) # The `syms` means leave the :y alone
 
 ```
 
@@ -363,11 +363,11 @@ outside of DataFramesMeta macros.
 
 # Working with `Symbol`s without referring to columns
 
-To refer to `Symbol`s without aliasing the column in a data frame, use `^`. 
+To refer to `Symbol`s without aliasing the column in a data frame, use `syms`. 
 
 ```
 df = DataFrame(x = [1, 1, 2, 2], y = [1, 2, 101, 102]);
-@select(df, :x2 = :x, :x3 = ^(:x))
+@select(df, :x2 = :x, :x3 = syms(:x))
 ```
 
 This rule applies to all DataFramesMeta macros.
--- a/src/DataFramesMeta.jl
+++ b/src/DataFramesMeta.jl
@@ -37,12 +37,12 @@ replace_syms!(x, membernames) = x
 replace_syms!(q::QuoteNode, membernames) =
     replace_syms!(Meta.quot(q.value), membernames)
 replace_syms!(e::Expr, membernames) =
-    if onearg(e, :^)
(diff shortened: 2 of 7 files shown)


bkamins (approved):
Looks good. Can you just leave comments in the code in places that have to be cleaned up after deprecation?

pdeffebach:
Thanks! 

Despite being a small PR, this is a big decision. I will ask on Slack how people feel.

jkrumbiegel:
My first thought is, why `syms` if it's used to designate one Symbol? Maybe `sym` is better. Another idea would be `@sym` because macro symbols always signal nicely that something non-standard is going on, while a normal looking function could fly under the radar. My last idea now would be to use `esc` because I think that's a function usually only used within macros, therefore not likely to clash with anything, and I think it's very closely related to its macro use. Escape the symbol into the outer expression, instead of mangling it inside the macro

jkrumbiegel:
I have seen you compared `syms` to the existing `cols` and I'd argue the same there. Why not `col` if we're talking about a single column specifier here?

nalimilan:
I agree a function is clearer than `^`. But why use the plural here? `esc` is also an interesting possibility.

> I have seen you compared `syms` to the existing `cols` and I'd argue the same there. Why not `col` if we're talking about a single column specifier here?

Actually `syms` is almost the contrary of `cols` here since it means that the symbol should be used literally.

pdeffebach:
Thanks for the feedback. I think we need to iterate on the design more. 

1. I called it `syms` because `cols` is plural. This would be more consistent. Given that `cols` only accepts a single column at the moment, arguably they should both be singular. But we should keep in mind #229, since we might eventually want `cols` to accept multiple columns eventually.  
2. `syms` is actually a bad name for this particular functionality. What `syms` *actually* does is protect the entire expression contained, so you can do 

```
@transform(df, x = syms([:a, :b])
```

or
 
```
@transform(df, x = syms(1 + 2 * 100))
```

in which case `syms` does nothing. 

I think this is a valid thing to be able to do, but I also think that, for the sake of the user, we could keep "protecting a `Symbol`" and "protecting an entire expression" separate.

nalimilan:
Right. So maybe `esc` is better?

pdeffebach:
Sure `esc` is good. 

It's not the most intuitive name but if you think it's consistent with it's usage in Base, it sounds good.

EconometricsBySimulation:
I am not sure if this would cause a problem with macro coding but what about `x = $(:a)`. I think `$` is used consistently to tell the user that the block of code is looking outside of itself.

knuesel:
`$` makes sense! It's very convenient (`x = $:a` should also work) and it's consistent with Base: the `@.` macro uses `$` to escape functions. From the documentation:

> If you want to avoid adding dots for selected function calls in `expr`, splice those function calls in with `$`. For example, `@. sqrt(abs($sort(x)))` is equivalent to `sqrt.(abs.(sort(x)))` (no dot for `sort`).

This is similar to how `$` would protect symbols from being "columnized".

pdeffebach:
Hmmm... I dont know, `$` also seems like it could be a good replacement for `cols`. Like in a quoted expression

```
x = :hello
t = :($x = 4)
```

I'm partial to replacing `cols` with `$` because it will be used more frequently.

Issue #166, opened 2020-09-14
@orderby in reverse


xiaodaigh:
Is it possible to achieve? Looking through the code, it doesn't seem like it's possible.

pdeffebach:
Thanks for the issue report. Yes this does seem feasible. 

DataFrames has a cool struct called `order` for specifying whether or not to go in a different order. 

```
  julia> sort(df, order(:x, rev=true))
```

it also has a `rev = true` keyword argument. 

Looking at `@orderby` it seems to have a lot of weird behavior. 

1. It applies to both `DataFrame`s and `GroupedDataFrame`s, but has different behavior. The first functions like `DataFrames.sort` while the second sorts the groups in a grouped data frame, which I don't think has a function in DataFrames currently. These should be split into two functions. 

2. Because in Julia `<` uses lexicographic comparisons on vectors, `@orderby(df, :b)` will actually do something. The thing to compare doesn't need to be a scalar. This is pretty unintuitive behavior for me at least. 

3. The `@orderby` code was written 6(!!!) years ago. I hardly understand it and I think it could be re-written in a more maintainable way. 

4. You will notice that none of the macros in DataFramesMeta actually allow keyword arguments, even if they look like they do. (try doing `transform(df; y = :x)`). This is because Julia puts keyword arguments in the *front* of the syntax tree for a call, rather than at the end, so things like `x.args[1]` break. 

In general I would be against keyword arguments in things like `@transform` because that would confuse the user. Is it a new column or a keyword argument? But in `@orderby` this seems fine because there is no pseudo-keyword argument syntax. 

Consequently I propose the following

1. Develop an infrastructure for adding parsing keyword arguments. What I would like is a struct with two fields, one with main arguments and one with the keyword arguments. That way we can apply existing `replace_syms!` recursive stuff on the first one and change behavior using the second one. 
1. Deprecate `@orderby` for an `AbstractDataFrame`. 
2. Implement a `@sort` macro which takes over the same behavior for an `AbstractDataFrame`
3. re-write `@orderby` to keep up with modern Julia idioms.

pdeffebach:
Paging @bkamins on this issue. 

Are there any recent developments in DataFrames that would make this easy to implement in DataFramesMeta? 

I had forgotten about this but want to get it sorted out between 1.0.

bkamins:
[...]

PR #231, opened 2021-03-07
Add `sortperm` example in `@orderby` docstring

pdeffebach:
In reference to #166

### DIFF
--- a/src/DataFramesMeta.jl
+++ b/src/DataFramesMeta.jl
@@ -531,7 +531,7 @@ julia> @orderby(d, sortperm(:c, rev = true))
    9 │     3      2  b
   10 │     3      1  a
 
-julia> @orderby(d, :x, :n .- mean(:n))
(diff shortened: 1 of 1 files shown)


src/DataFramesMeta.jl
```diff
@@ -496,40 +496,57 @@ the given `DataFrame` on the result, returning a new `DataFrame`.
...
+   3 │     2      8  h
+   4 │     1      7  g
+   5 │     1      6  f
+   6 │     1      5  e
+   7 │     2      4  d
+   8 │     3      3  c
+   9 │     3      2  b
+  10 │     3      1  a
 
 julia> @orderby(d, :x, :n .- mean(:n))
```
nalimilan, line 534: While we're at it, this is equivalent to just `@orderby(d, :x, :n)`, right? Can we find a better example?

pdeffebach: Added a slightly more exciting example.

bkamins (approved):

PR #241, opened 2021-04-20
Bump version to allow Reexport 1+

raphaelsaavedra:
Closes https://github.com/JuliaData/DataFramesMeta.jl/issues/240.

I did a minor version bump – let me know if there have been any breaking changes in master so I can fix it.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "DataFramesMeta"
 uuid = "1313f7d8-7da2-5740-9ea0-a2ca25f37964"
-version = "0.6.0"
+version = "0.6.1"
 
 [deps]
 DataFrames = "a93c6f00-e57d-5684-b7b6-d8193f3e46c0"


pdeffebach:
Looking through PRs, the only changes to the API are 

1. allowing `@select(df, cols(:a))`
2. adding `@select!` and `@transform!`

These were all errors before, so I think this means we can do `0.6.1`? cc @bkamins

bkamins:
Yes, but in line 10 (https://github.com/JuliaData/DataFramesMeta.jl/pull/241/files#diff-72ed386c2a0cd1d23c0968297e70023ed98c22490d146dd89fc91f48369bad4dR10) of `[compat]` section we should also allow for DataFrames.jl 1 version. Can this change please be added? Than you!

bkamins (approved):

pdeffebach:
Thanks for this! 

I was going to wait a bit longer to make a release, but if we want `reexport` to be consistent, it's better to make a patch release now. I will work on a small write up of what's changed and post that somewhere.

bkamins:
@pdeffebach - OK to make 0.6.1 release, because I see you have merged the PR but not made the release.

PR #245, opened 2021-04-22
Make block

pdeffebach:
With this PR we have 

```
julia> df = DataFrame(a = 1, b = 2);

julia> @transform df begin 
           x = 1
           y = 2
           z = 5
       end
1×5 DataFrame
 Row │ a      b      x      y      z     
     │ Int64  Int64  Int64  Int64  Int64 
─────┼───────────────────────────────────
   1 │     1      2      1      2      5
```


bkamins:
If I understand this correctly you want to bring a similar syntax to what Chain.jl offers?

pdeffebach:
The implementation is similar, yeah. Turn a collection of args into a `:block`. 

I need to think harder about what I do and do not want to allow, as I think being too flexible adds a lot of complication. I'm surprised that tests fail, they passed locally.

pdeffebach:
Chain.jl seems to have figured this out, actually. cc @jkrumbiegel

pdeffebach:
This roadblock has been solved. It was an optimization that was introduced in the anonymous functions PR. I have fixed it by removing `LineNumberNode`s in more places. Will have to figure out how bad this is.

src/macros.jl
```diff
@@ -674,11 +677,14 @@ function combine_helper(x, args...; deprecation_warning = false)
     deprecation_warning && @warn "`@based_on` is deprecated. Use `@combine` instead."
 
     # Only allow one argument when returning a Table object
-    if length(args) == 1 &&
-        !(first(args) isa QuoteNode) &&
-        !(first(args).head == :(=) || first(args).head == :kw)
+    exprs = fix_args(args...)
```
pdeffebach, line 680: This is complicated stuff to preserve some the 0.21.* rules about only having one `AsTable` argument. 

It's clunky and only exists to ensure tests pass. I am fine deprecating in a later PR.

src/parsing.jl
```diff
@@ -106,9 +106,9 @@ function get_source_fun(function_expr)
     # recursive step for begin :a + :b end
     if function_expr isa Expr &&
         function_expr.head == :block &&
-        length(function_expr.args) == 2 # omitting the line number node
```
pdeffebach, line 109: I just call `Base.remove_linenums!` wherever possible, which changed the way we index expressions in a few places.

pdeffebach (commented):
Ready for a review!

test/dataframes.jl
```diff
@@ -146,9 +183,13 @@ end
     @test @transform!(df, n = 1).n == fill(1, nrow(df))
     @test @transform!(df, n = :i .* :g).n == [1, 2, 3, 8, 10]
 
+
+
+
+
```
bkamins, line 189: this is probably not needed - right?
```suggestion
```

pdeffebach:
Okay I've cleaned up the `diff` a bit. 

One thing in particular to look at is the `transform` tests. We should nail those down before I copy and paste for every macro.

pdeffebach:
I've simplified the implementation and changed the docstring for `@transform`. Requesting a review from @nalimilan

src/macros.jl
```diff
@@ -436,14 +436,37 @@ Add additional columns or keys based on keyword arguments.
 
 * `::AbstractDataFrame` or `::GroupedDataFrame`
 
+Inputs to `@transform` can come in two formats. One
+can create new columns in a single block, where each
+line is of the form `y = f(:x)`, or as a series of
+keyword arguments. For example, the following are
+equivelant
```
nalimilan, line 443:
```suggestion
equivalent.
```

src/macros.jl
```diff
@@ -436,14 +436,37 @@ Add additional columns or keys based on keyword arguments.
...
+
+```julia
+@transform df begin
+  a = :x
+  b = :y
```
nalimilan, line 448:
```suggestion
    a = :x
    b = :y
```

src/macros.jl
```diff
@@ -436,14 +436,37 @@ Add additional columns or keys based on keyword arguments.
 
 * `::AbstractDataFrame` or `::GroupedDataFrame`
 
+Inputs to `@transform` can come in two formats. One
+can create new columns in a single block, where each
```
nalimilan, line 440:
```suggestion
Inputs to `@transform` can come in two formats.
New columns can be created either in a single block, where each
```
(I first read "one" as meaning "one way is to".)

src/macros.jl
```diff
@@ -436,14 +436,37 @@ Add additional columns or keys based on keyword arguments.
...
 
 ```jldoctest
 julia> using DataFramesMeta
 
-julia> @transform(df, a = 2 * :A, x = :A .+ :B)
+julia> @transform df begin
+  a = 2 * :A
+  x = :A .+ :B
```
nalimilan, line 467:
```suggestion
    a = 2 * :A
    x = :A .+ :B
```

src/parsing.jl
```diff
@@ -305,4 +305,31 @@ function replace_dotted!(e, membernames)
...
     y_new = protect_replace_syms!(e.args[2], membernames)
     Expr(:., x_new, y_new)
+end
+
+"""
+    fix_args(args...; )
+
+Given multiple arguments which can be any type
+of expression-like object (`Expr`, `QuoteNode`, etc.),
+appends them
```
nalimilan, line 315: "appends them"? AFAICT this collects them in a vector after removing line numbers?

Also, couldn't the function use a more specific name than "fix"?

test/dataframes.jl
```diff
@@ -147,8 +178,8 @@ end
     @test @transform!(df, n = :i .* :g).n == [1, 2, 3, 8, 10]
 
     # non-copying
-    @test @transform!(df, n = :i).g === df.g
-    @test @transform!(df, n = :i).n === df.i
+    @test @transform!(df, n = :i).g == df.g
+    @test @transform!(df, n = :i).n == df.i
```
nalimilan, line 182: This change defeats the purpose of this test, right? :-)

test/function_compilation.jl
```diff
@@ -18,172 +18,172 @@ using DataFramesMeta
 
             fasttime = @timed @select(df, c = :a + :b)
             slowtime = @timed select(df, [:a, :b] => ((a, b) -> a + b) => :c)
-            (slowtime[2] > fasttime[2]) || @warn("Slow compilation")
+            (slowtime[2] > fasttime[2]) || @error("Slow compilation")
```
nalimilan, line 21: Unrelated?

nalimilan:
Interesting! Have you thought about the possibility to support advanced syntax allowed by `@eachrow` in the future, like `@newcol x::Vector{Float64}` and `if .... x = 1; else; x = 2; end`? That would allow simplifying the API by deprecating `@eachrow` in favor of `@transform`.

> This roadblock has been solved. It was an optimization that was introduced in the anonymous functions PR. I have fixed it by removing `LineNumberNode`s in more places. Will have to figure out how bad this is.

Isn't it a problem to remove line numbers? Aren't they necessary to print accurate stacktraces in case of errors? (That's really a strong point compared to dplyr where I remember having a hard time finding out which lines generate errors.)

Finally, I guess the idea is to extend this to `@transform!`, `@select`, `@select!`, etc.?

test/dataframes.jl
pdeffebach: It's an upstream change in DataFrames. I should do it in it's own PR.

test/function_compilation.jl
pdeffebach: Wow sorry about that. From debugging.

src/parsing.jl
pdeffebach: fixed.

pdeffebach:
> Interesting! Have you thought about the possibility to support advanced syntax allowed by `@eachrow` in the future, like `@newcol x::Vector{Float64}` and `if .... x = 1; else; x = 2; end`? That would allow simplifying the API by deprecating `@eachrow` in favor of `@transform`.

I don't think so. `@eachrow` is a for-loop and so you can do all sorts of things inside it, not strictly create columns. 

This leads me to a design question: Is it weird to have `@transform begin ... end` with a `:block` but require that expressions *only* start with `y = ...`? Is this against the philosophy of using a `:block`? 

In `@eachrow` for example, you can do anything you want inside an `@eachrow` block. 

> Finally, I guess the idea is to extend this to `@transform!`, `@select`, `@select!`, etc.?

Yes, If you think the tests and docstring are good for `@transform` I will copy them over to everything else.

pdeffebach:
> Isn't it a problem to remove line numbers? Aren't they necessary to print accurate stacktraces in case of errors? (That's really a strong point compared to dplyr where I remember having a hard time finding out which lines generate errors.)

You would think so, but we are no worse than in 0.6.0, before the parsing re-write where I dropped line numbers. 

I think part of this might be due to Chain. We seem to get the line number of the start of a `@chain` block, but not commands inside of it. Which is really bad! But no worse than before. We can detect errors okay outside of `@chain`. 

This needs a separate investigation later on. If we find that these line number removals are causing problems, I can fix them later. But it's likely upstream.

pdeffebach:
w.r.t. line numbers. I think we are in the clear. 

With this script 

```
using Chain

foo(x, y) = x * y

x = [1, 2]

@chain x begin
    identity
    identity
    identity
    foo([3, 4])
end
```

We get the error

```
julia> include("linenumbers_test.jl")
ERROR: LoadError: MethodError: no method matching *(::Vector{Int64}, ::Vector{Int64})
Closest candidates are:
  *(::Any, ::Any, ::Any, ::Any...) at operators.jl:560
  *(::StridedMatrix{T}, ::StridedVector{S}) where {T<:Union{Float32, Float64, ComplexF32, ComplexF64}, S<:Real} at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/matmul.jl:44
  *(::StridedVecOrMat{T} where T, ::LinearAlgebra.Adjoint{var"#s832", var"#s831"} where {var"#s832", var"#s831"<:LinearAlgebra.LQPackedQ}) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/lq.jl:254
  ...
Stacktrace:
 [1] foo(x::Vector{Int64}, y::Vector{Int64})
   @ Main ~/Documents/Development/DataFramesMeta/linenumbers_test.jl:3
 [2] top-level scope
   @ ~/Documents/Development/DataFramesMeta/linenumbers_test.jl:11
 [3] include(fname::String)
   @ Base.MainInclude ./client.jl:444
 [4] top-level scope
   @ REPL[58]:1
in expression starting at /home/peterwd/Documents/Development/DataFramesMeta/linenumbers_test.jl:7
```

In `[2]` is correctly identifies the line number as line `11`. Unfortuantely it points to the block beginning on line `7` as the most highlighted error. But oh well, at least it's catching line `11`.

pdeffebach:
@nalimilan this PR should be ready for a review, if not merging. 

I added docstrings and tests for everything. 

I didn't change `index.md` in the docs because there aren't many examples of multi-argument transforms. Though eventually I want to get rid of all commas and parentheses in the docs. But I will wait until after `@byrow` is added to do that re-write.

pdeffebach:
Fixed conflicts with master branch after #247 

Still ready for a review.

bkamins (approved):
Looks good.

src/parsing.jl
```diff
@@ -305,4 +305,32 @@ function replace_dotted!(e, membernames)
     x_new = replace_syms!(e.args[1], membernames)
     y_new = protect_replace_syms!(e.args[2], membernames)
     Expr(:., x_new, y_new)
+end
+
+"""
+    create_args_vector(args...; )
```
nalimilan, line 311:
```suggestion
    create_args_vector(args...)
```

src/macros.jl
```diff
@@ -225,6 +226,23 @@ If given a `GroupedDataFrame`, `@where` applies transformations by
 group, and returns a fresh `DataFrame` containing the rows
 for which the generated values are all `true`.
 
+Inputs to `@where` can come in two formats: a `block`, in which case each
```
nalimilan, line 229: "Block" isn't a language keyword, right? Also isn't super easy to understand. Maybe "as a `begin... end` block" and say "line" instead of "argument"?

(Same for other functions.)

src/macros.jl
```diff
@@ -225,6 +226,23 @@ If given a `GroupedDataFrame`, `@where` applies transformations by
...
+argument in the block is a separate selector, or as multiple arguments.
+For example the following two statements are equivalent.
```
nalimilan, line 231:
```suggestion
For example the following two statements are equivalent:
```

nalimilan (commented):
Sorry for the delay. Just small details about the docs.

src/macros.jl
pdeffebach: Fixed all of these.

pdeffebach:
Thanks for the review. I have fixed all the issues. Once tests finish this should be ready for merging.

nalimilan (approved):

pdeffebach:
Thanks!

PR #246, opened 2021-05-02
Fix typo in manual

nalimilan:


pdeffebach:
Thanks, test failure are fixed in #247 so we can merge this.

PR #247, opened 2021-05-15
Fix of === after DataFrames PR

pdeffebach:
After https://github.com/JuliaData/DataFrames.jl/pull/2721, `transform!(df, :a => identity => :b)` will perform a copy. This is causing breaking tests.


bkamins (approved):

pdeffebach:
Thanks!

PR #250, opened 2021-05-28
Add @byrow attempt 2

pdeffebach:
This supercedes #239 

Here is a docstring indicating how `@byrow` works

----
`@transform` uses the syntax `@byrow` to wrap transformations in
the `ByRow` function wrapper from DataFrames, enabling broadcasting
and more. For example, the call

```
@transform(df, y = @byrow :x == 1 ? "true" : "false)
```

becomes

```
transform(df, :x => ByRow(x -> x == 1 ? "true", "false") => :y)
```

a transformation which  cannot be conveniently expressed
using broadcasting.

To avoid writing `@byrow` multiple times when performing multiple
transformations by row, `@transform` allows `@byrow` at the
beginning of a block of transformations. All transformations
in the block will operate by row.

```
julia> using DataFramesMeta

julia> @transform df z = @byrow :A * :B
3×3 DataFrame
 Row │ A      B      z
     │ Int64  Int64  Int64
─────┼─────────────────────
   1 │     1      2      2
   2 │     2      1      2
   3 │     3      2      6

julia> @transform df @byrow begin
           x = :A * :B
           y = :A == 1 ? 100 : 200
       end

3×4 DataFrame
 Row │ A      B      x      y
     │ Int64  Int64  Int64  Int64
   1 │     1      2      2    100
   2 │     2      1      2    200
   3 │     3      2      6    200
```
----

The implementation is a bit clunky with `create_args_vector` returning a `tuple` of the vector of arguments and whether to wrap in `ByRow`. This should be less clunky after we clean up the `@combine` code, which I will do in a later PR. So I accumulate a little bit of technical debt I promise I will pay off in the future. 

I have added tests for `@transform`, but not others, until we find all the edge cases we need. Overall I am pleased with this PR. I think the work in #245 made this really easy. It also preserves the snappy function parsing in #221, though I still have to add those tests. 

All things considered, this is ready for a review @nalimilan and @bkamins.


bkamins:
Thank you for working on it. Note that CI fails.

src/macros.jl
```diff
@@ -503,6 +503,28 @@ and
 @transform(df, a = :x, b = :y)
 ```
 
+`@transform` uses the syntax `@byrow` to wrap transformations in
```
bkamins, line 506: Is `@byrow` only `@transform` specific (like `@aside` in `@chain`) or it can be used stand-alone?

src/macros.jl
```diff
@@ -503,6 +503,28 @@ and
...
+and more. For example, the call
+
+```
+@transform(df, y = @byrow :x == 1 ? "true" : "false)
+```
+
+becomes
+
+```
+transform(df, :x => ByRow(x -> x == 1 ? "true", "false") => :y)
```
bkamins, line 517: I think the point is that only `:x` will be broadcasted. Right? So if instead of e.g. `"true"` we would have `[1,2,3]` it would not be broadcasted.

src/macros.jl
```diff
@@ -511,17 +533,38 @@ julia> using DataFramesMeta
...
+       end
+3×4 DataFrame
+ Row │ A      B      a      x
+     │ Int64  Int64  Int64  Int64
+─────┼────────────────────────────
+   1 │     1      2      2      3
+   2 │     2      1      4      3
+   3 │     3      2      6      5
+
+julia> @transform df z = @byrow :A * :B
```
bkamins, line 547: do you think it is didactic to show `@transform` without `(` and `)` when you do not use `begin`-`end`?

test/dataframes.jl
```diff
@@ -123,6 +123,56 @@ end
...
 
+@testset "@transform with @byrow" begin
+    df = DataFrame(
+        g = [1, 1, 1, 2, 2],
+        i = 1:5,
+        t = ["a", "b", "c", "c", "e"],
+        y = [:v, :w, :x, :y, :z],
+        c = [:g, :quote, :body, :transform, missing])
+
+    @test @transform(df, n = @byrow :i).n == df.i
```
bkamins, line 134: Maybe add some tests checking the whole output `DataFrame` not only its specific columns.
Also do we support `GroupedDataFrame` here?

src/macros.jl
pdeffebach: It is no a real macro, so it's like `@aside`. 

This is partially a good thing. We've had a deprecation warning for the old version of `@byrow` for a year or so, but never actually made it an error. 

Even if we did make it a real macro, it's not obvious what it's behavior should be. One could imagine it would be like `@with` but operating by row, but this can easily be accomplished with

```julia
julia> using DataFramesMeta;

julia> @with df @byrow begin 
           :a * :b
       end
2-element Vector{Int64}:
 3
 8
```

src/macros.jl
pdeffebach: Yes. This also avoids the need for `Ref` with broadcasting. I will add a section in the docs (not the docstring) discussing this along with a few other gotchas.

src/macros.jl
pdeffebach: I would generally like this to be the default way of writing in the future. I think it's cleaner and more declarative. So maybe I can have a future PR change some of these en-masse if people agree with me.

test/dataframes.jl
pdeffebach: I have updated tests to be more reflective of the types of cases we are concerned with. 

I will updated the `GroupedDataFrame` tests once we think these are good. They are in a different file.

pdeffebach:
Okay I've added a very lengthy documentation section in `index.md`. 

Could use an approval of the tests, any comments on the API as discussed in `index.md` and the `@transform` docstring. and then I will get to work adding tests everywhere and cleaning up the documentation.

docs/src/index.md
```diff
@@ -268,6 +268,178 @@ df2 = @eachrow df begin
 end
 ```
 
+## Row-wise transformations with `@byrow`
+
+DataFrames provides the function-wrapper `ByRow`. `ByRow(f)(x, y)`
```
bkamins, line 273: I am not sure what style you follow in DataFramesMeta.jl, but in DataFrames.jl we always add .jl to package names

docs/src/index.md
```diff
@@ -268,6 +268,178 @@ df2 = @eachrow df begin
...
+is roughly equivalent to `f.(x, y)`, with a few exceptions discussed below. 
+DataFramesMeta allows for users to construct expressions using `ByRow` 
+function wrapper with the syntax `@byrow`.
```
bkamins, line 276: maybe highlight that it is not a macro?

bkamins: ah - you did below :)

docs/src/index.md
```diff
@@ -268,6 +268,178 @@ df2 = @eachrow df begin
...
+```
+
+becomes
+
+```
+transform(df, :x => ByRow(x -> x == 1 ? "true", "false") => :y)
+```
+
+!!! note 
+    Unlike `@.`, `@byrow` is not a "real" macro and cannot be used outside of
```
bkamins, line 289: the question is what are the parsing rules for `@byrow`? does it take exactly one expression that follows it?

docs/src/index.md
```diff
@@ -268,6 +268,178 @@ df2 = @eachrow df begin
...
+
+!!! note 
+    Unlike `@.`, `@byrow` is not a "real" macro and cannot be used outside of 
+    DataFramesMeta macros. However it's behavior within DataFramesMeta 
+    macros should be indistinguishable from externally defined macros. 
+
+### Comparison with `@eachrow`
+
+In previous versions of DataFramesMeta, `@eachrow` was named `@byrow`. 
+This version of `@byrow` is deprecated, but the syntax can be used
```
bkamins, line 296: "the syntax" - which syntax?

docs/src/index.md
```diff
@@ -268,6 +268,178 @@ df2 = @eachrow df begin
...
+to for similar, but not identical, behavior.
```
bkamins, line 297:
```suggestion
too for similar, but not identical, behavior.
```

docs/src/index.md
```diff
@@ -268,6 +268,178 @@ df2 = @eachrow df begin
...
+            a[i] * b[i]
+        end
+    end
+    tempfun(df.a, df.b)
+    df
+end
+```
+
+The function `*` is applied by-row. But the result of those operations
+is not stored in a new vector. Additionally, `@eachrow` and `@eachrow!`
```
bkamins, line 322: where is it stored then? (or not stored unless stored explicitly?)

docs/src/index.md
```diff
@@ -268,6 +268,178 @@ df2 = @eachrow df begin
...
+return data frames. 
+
+By contrast,
+
+```julia
+@with df @byrow begin
```
bkamins, line 328: this is confusing. I understand you refer to the deprecated `@byrow`? If this is the case then I would skip it - no need to document (or even mention deprecated functionality except for writing that it is deprecated).

If you refer to new `@byrow` then it is strange, as earlier you have written that `@byrow` is restricted to `@transform`.

bkamins: Looking at your examples below it seems `@byrow` is accepted not only in `@transform` - I would suggest explicitly listing places where it is allowed.

docs/src/index.md
```diff
@@ -268,6 +268,178 @@ df2 = @eachrow df begin
...
+           end
+       end # will error
+```
+
+* Broadcasting objects that are not columns. `@byrow` constructs an 
+anonymous function *which accepts only the columns of the dataframe*
+and broadcasts that function. Consequently, it does not broadcast
+objects that are referenced which are not columns. 
+```julia
+@with df @byrow :x + [5, 6]
```
bkamins, line 407: will this always error? What if column `:x` would be a vector of two element vectors?

docs/src/index.md
```diff
@@ -268,6 +268,178 @@ df2 = @eachrow df begin
...
+```
+will error. On the other hand
+```julia
+@with df @. :x + [5, 6]
```
bkamins, line 411: It will only work if `:x` has 2 elements - right?

docs/src/index.md
```diff
@@ -268,6 +268,178 @@ df2 = @eachrow df begin
...
+```
+will not. 
+
+* Broadcasting expensive calls. In Base Julia, broadcastsing 
+evaluates calls first and then broadcasts the result. Because
+`@byrow` constructs an anonymous function and evaluates
+that function for every row in the DataFrame, expensive functions
```
bkamins, line 418:
```suggestion
that function for every row in the data frame, expensive functions
```

docs/src/index.md
```diff
@@ -268,6 +268,178 @@ df2 = @eachrow df begin
...
+This problem comes up when using the `@.` macro as well, but can easily be fixed with `$`.
+```julia
+julia> @time @with df @. :a + expensive();
+  1.036888 seconds (97.55 k allocations: 5.617 MiB, 3.20% compilation time)
+
+julia> @time @with df @. :a + $expensive();
+  0.537961 seconds (110.68 k allocations: 6.525 MiB, 6.73% compilation time)
+```
+No such solution currently exists with `@byrow`. 
+
```
bkamins, line 442: as commented earlier - from the description it is not clear how `GroupedDataFrame` is handled with `@byrow` and I did not see any tests of `GroupedDataFrame` below.

docs/src/index.md
pdeffebach: clarified.

docs/src/index.md
pdeffebach: Added.

docs/src/index.md
pdeffebach: Added in the re-write. In general it's a bit pointless to use `@byrow` with a GroupedDataFrame as with `ByRow`. Though maybe if one wants to combine operations which take advantage of the grouping and operations which do not it could be useful.

docs/src/index.md
pdeffebach: yes. I have added more detail.

docs/src/index.md
pdeffebach: Yes, I have clarified in the example.

docs/src/index.md
pdeffebach: it's not stored. It's literally just a `for`-loop. Hopefully the re-write is clear.

docs/src/index.md
pdeffebach: added `.jl` but will make a point to do this before 1.0.

docs/src/index.md
pdeffebach: I included a more detailed discussion in updated docs. Hopefully it's clear.

pdeffebach:
Thanks a ton for reading all this. 

I re-wrote a lot of this documentation for clarity based on your comments. Hopefully it's more complete now.

pdeffebach:
It's worth noting that the special behavior related to `NamedTuple`s and `AsTable` doesn't get referenced here because DataFramesMeta doesn't support `AsTable `yet.

docs/src/index.md
bkamins: Agreed - but the documentation should precisely state "how things work" (this is a definite reference for a contract of the functionalities you provide), and then also recommend "what makes sense" if needed I think.

pdeffebach:
I added the tests for all the macros and ensured tests pass. Will add docstrings for everything next, but could still use a review.

docs/src/index.md
```diff
@@ -268,6 +268,222 @@ df2 = @eachrow df begin
 end
 ```
 
+## Row-wise transformations with `@byrow`
+
+DataFrames.jl provides the function-wrapper `ByRow`. `ByRow(f)(x, y)`
+is roughly equivalent to `f.(x, y)`, with a few exceptions discussed below.
```
bkamins, line 274: do you discuss these exceptions indeed? The exceptions are:
* `AsTable` case - which I think we do not cover yet
* potentially another type of container can be allocated as we use `map` which is not guaranteed to be 100% consistent with broadcasting in corner cases

(I am not sure it is worth to go into such details in the documentation but I felt it is good to leave this comment)

docs/src/index.md
```diff
@@ -268,6 +268,222 @@ df2 = @eachrow df begin
...
 
+## Row-wise transformations with `@byrow`
+
+DataFrames.jl provides the function-wrapper `ByRow`. `ByRow(f)(x, y)`
+is roughly equivalent to `f.(x, y)`, with a few exceptions discussed below. 
+DataFramesMeta.jl allows users to construct expressions using `ByRow` 
+function wrapper with the syntax `@byrow`. 
+
+`@byrow` is not a "real" macro and cannot be used outside of 
+DataFramesMeta macros. However it's behavior within DataFramesMeta.jl
```
bkamins, line 279:
```suggestion
DataFramesMeta.jl macros. However, it's behavior within DataFramesMeta.jl
```

docs/src/index.md
```diff
@@ -268,6 +268,222 @@ df2 = @eachrow df begin
...
+
+```julia
+transform(df, :x => ByRow(x -> x == 1 ? "true", "false") => :y)
+```
+
+Macros that accept `@byrow`:
+
+* `@transform` and `@transform!`, `@select`, `@select!`, and `@combine`. `
+  @byrow` appears in the right hand side of expressions, of the form `@select(
+  df, z = @byrow :x * :y)`.
```
bkamins, line 298:
```suggestion
* `@transform` and `@transform!`, `@select`, `@select!`, and `@combine`.
  `@byrow` appears in the right hand side of expressions, of the form
  `@select(df, z = @byrow :x * :y)`. 
```

docs/src/index.md
```diff
@@ -268,6 +268,222 @@ df2 = @eachrow df begin
...
+```
+
+Macros that accept `@byrow`:
+
+* `@transform` and `@transform!`, `@select`, `@select!`, and `@combine`. `
+  @byrow` appears in the right hand side of expressions, of the form `@select(
+  df, z = @byrow :x * :y)`. 
+* `@where` and `@orderby`, with syntax of the form `@where(df, @byrow :x > :y)`
+* `@with`, where the anonymous function created by `@with` is wrapped in
+  `ByRow`. `@with df @byrow :x * :y`, which is conceptually similar to
```
bkamins, line 301: for consistency with other examples:
```suggestion
  `ByRow`. `@with(df, @byrow :x * :y)`, which is conceptually similar to 
```

docs/src/index.md
```diff
@@ -268,6 +268,222 @@ df2 = @eachrow df begin
...
+  @byrow` appears in the right hand side of expressions, of the form `@select(
+  df, z = @byrow :x * :y)`. 
+* `@where` and `@orderby`, with syntax of the form `@where(df, @byrow :x > :y)`
+* `@with`, where the anonymous function created by `@with` is wrapped in
+  `ByRow`. `@with df @byrow :x * :y`, which is conceptually similar to 
+  `map((x, y) -> x * y, df.x, df.y)`.
+
+To avoid writing `@byrow` multiple times when performing multiple
+operations, it is allowed `@byrow` at the beginning of a block of 
+operations. All transformations in the block will operate by row.
```
bkamins, line 306: one transformation is one expression?

docs/src/index.md
```diff
@@ -268,6 +268,222 @@ df2 = @eachrow df begin
...
+ Row │ a      b     
+     │ Int64  Int64 
+─────┼──────────────
+   1 │     2      4
+```
+
+`@byrow` can be used on macros which accept `GroupedDataFrame`s,
+however, like with `ByRow` in DataFrames.jl, when `@byrow` is
+used, functions do not take advantage of the grouping, so the 
+behavior of `@transform(df, y = @byrow f(:x))` and
```
bkamins, line 323: I add this comment as with `@combine` the result would be different:
```suggestion
used, functions do not take advantage of the grouping, so 
for example the result of `@transform(df, y = @byrow f(:x))` and 
```

docs/src/index.md
```diff
@@ -268,6 +268,222 @@ df2 = @eachrow df begin
...
+
+`@byrow` can be used on macros which accept `GroupedDataFrame`s,
+however, like with `ByRow` in DataFrames.jl, when `@byrow` is
+used, functions do not take advantage of the grouping, so the 
+behavior of `@transform(df, y = @byrow f(:x))` and 
+`@transform(groupby(df, :g), y = @byrow f(:x))` is the same.
+
+### Comparison with `@eachrow`
+
+In previous versions of DataFramesMeta, `@eachrow` was named `@byrow`.
```
bkamins, line 328:
```suggestion
In previous versions of DataFramesMeta.jl, `@eachrow` was named `@byrow`. 
```

docs/src/index.md
```diff
@@ -268,6 +268,222 @@ df2 = @eachrow df begin
...
+`@byrow` can be used on macros which accept `GroupedDataFrame`s,
+however, like with `ByRow` in DataFrames.jl, when `@byrow` is
+used, functions do not take advantage of the grouping, so the 
+behavior of `@transform(df, y = @byrow f(:x))` and 
+`@transform(groupby(df, :g), y = @byrow f(:x))` is the same.
+
+### Comparison with `@eachrow`
+
+In previous versions of DataFramesMeta, `@eachrow` was named `@byrow`. 
+The old macro `@byrow` is deprecated, but the `@byrow` syntac can
```
bkamins, line 329:
```suggestion
The old macro `@byrow` is deprecated, but the `@byrow` syntax can 
```

docs/src/index.md
pdeffebach: I added a short note about the broadcasting issue. 

I will avoid discussing `AsTable` until we implement that.

docs/src/index.md
pdeffebach: Yes, that's the new rule after #245. Hopefully it's not too confusing to people.

docs/src/index.md
pdeffebach: No. By "`@byrow` syntax" I mean `@with df @byrow ...` is very similar to `@eachrow df ...`. In that it's looking through rows.

pdeffebach:
Thanks for the detailed review!

bkamins:
Let us wait for @nalimilan to have a look at it now.

docs/src/index.md
```diff
@@ -268,6 +268,228 @@ df2 = @eachrow df begin
...
+  will error, because the `:x` in the above expression refers
+  to a scalar `Int`, and you cannot do `1 + [5, 6]`.
+
+  On the other hand
+
+  ```julia
+  @with df @. :x + [5, 6]
+  ```
+  will not, as `df.x` is a 2-element vector as is `[5, 6]`, and 
+  `1 .+ [5, 6]` is allowed.
```
nalimilan, line 453: `1 .+ [5, 6]` is never called, right? The two arguments have the same length.
```suggestion
  will succeed, as `df.x` is a 2-element vector as is `[5, 6]`.
```

docs/src/index.md
```diff
@@ -268,6 +268,228 @@ df2 = @eachrow df begin
...
+
+  Because `ByRow` inside `transform` blocks does not internally 
+  use broadcasting in all circumstances, in the rare instance
+  that a column in a data frame is a custom vector type that
+  implements custom broadcasting, this custom behavior will 
+  be called with `@byrow`.
```
nalimilan, line 459: Is this what you mean?
```suggestion
  not be called with `@byrow`.
```

docs/src/index.md
```diff
@@ -268,6 +268,228 @@ df2 = @eachrow df begin
...
+
+* Broadcasting expensive calls. In Base Julia, broadcastsing
```
nalimilan, line 461:
```suggestion
* Broadcasting expensive calls. In Base Julia, broadcasting 
```

docs/src/index.md
```diff
@@ -268,6 +268,228 @@ df2 = @eachrow df begin
...
+tempfun.(df.a, df.b)
+```
+
+In contrast to `@eachroe`, `@with` combined with `@byrow` will return a vector of the broadcasted multiplication and not a data frame.
+
+Additionally, `@eachrow` and `@eachrow!` allow modifying a data
+data frame. Just as with Base Julia broadcasting, `@byrow` will
+not update columns. 
+
+```julia
```
nalimilan, line 379: How about making this a doctest?

docs/src/index.md
```diff
@@ -268,6 +268,228 @@ df2 = @eachrow df begin
...
+  
+  julia> @time @with df @byrow :a + expensive();
+    1.037073 seconds (51.67 k allocations: 3.035 MiB, 3.19% compilation time)
+  
+  julia> @time @with df :a .+ expensive();
+    0.539900 seconds (110.67 k allocations: 6.525 MiB, 7.05% compilation time)
+  
+  ```
+  
+  This problem comes up when using the `@.` macro as well, but can easily be   fixed with `$`.
```
nalimilan, line 481:
```suggestion
This problem comes up when using the `@.` macro as well, but can easily be fixed with `$`.
```

src/eachrow.jl
```diff
@@ -70,27 +70,6 @@ function eachrow_helper(df, body, deprecation_warning)
...
-"""
-    @byrow(d, expr)
-
-Deprecated version of `@eachrow`, see: [`@eachrow`](@ref)
-
-Acts the exact same way.
-"""
-macro byrow(d, body)
-    esc(eachrow_helper(d, body, true))
-end
```
nalimilan, line 73:
```suggestion
```

src/macros.jl
```diff
@@ -176,8 +176,8 @@ end
 ##############################################################################
 
 function where_helper(x, args...)
-    exprs = create_args_vector(args...)
-    t = (fun_to_vec(ex; gensym_names = true, nolhs = true) for ex in exprs)
+    exprs, wrap_ByRow = create_args_vector(args...)
```
nalimilan, line 179: Usually, variable names are all lower case.

src/macros.jl
```diff
@@ -463,8 +463,8 @@ end
 
 
 function transform_helper(x, args...)
-    exprs = create_args_vector(args...)
-    t = (fun_to_vec(ex; gensym_names = false, nolhs = false) for ex in exprs)
+    exprs, wrap_ByRow = create_args_vector(args...)
+    t = [fun_to_vec(ex; gensym_names = false, nolhs = false, wrap_ByRow = wrap_ByRow) for ex in exprs]
```
nalimilan, line 467: Why use `[` here but not in other functions?

src/macros.jl
nalimilan, line 506:
```suggestion
`@transform` supports the `@byrow` syntax to wrap transformations in
```

src/macros.jl
```diff
@@ -503,6 +503,28 @@ and
 @transform(df, a = :x, b = :y)
 ```
 
+`@transform` uses the syntax `@byrow` to wrap transformations in
+the `ByRow` function wrapper from DataFrames, enabling broadcasting
+and more. For example, the call
```
nalimilan, line 508: "and more" isn't super explicit. Maybe keep exporting `@byrow` so that you can attach a docstring to it with more details (copying what appears in the manual)? Then I wouldn't mention `ByRow` here to keep things simple.

src/macros.jl
```diff
@@ -503,6 +503,28 @@ and
...
+@transform(df, y = @byrow :x == 1 ? "true" : "false)
+```
+
+becomes
+
+```
+transform(df, :x => ByRow(x -> x == 1 ? "true", "false") => :y)
+```
+
+a transformation which  cannot be conveniently expressed
```
nalimilan, line 520:
```suggestion
a transformation which cannot be conveniently expressed
```
BTW, is this still the case on Julia 1.7? Otherwise I wouldn't mention it.

src/parsing.jl
```diff
@@ -83,11 +85,16 @@ representing the vector of inputs that will be
 used as the `src` in the `src => fun => dest`
 call later on.
 
+If `wrap_ByRow=true` then the function gets wrapped
+in `ByRow`. If the expression begins with `@byrow`,
+then `get_source_fun` is recurively called on the
```
nalimilan, line 90:
```suggestion
then `get_source_fun` is recursively called on the
```

src/parsing.jl
```diff
@@ -315,22 +332,35 @@ of expression-like object (`Expr`, `QuoteNode`, etc.),
 puts them into a single array, removing line numbers.
 """
 function create_args_vector(args...)
-    Any[Base.remove_linenums!(arg) for arg in args]
+    Any[Base.remove_linenums!(arg) for arg in args], false
 end
 
 """
    create_args_vector(arg)
```
nalimilan, line 339:
```suggestion
   create_args_vector(arg) -> vec, wrap_ByRow
```

test/dataframes.jl
```diff
@@ -123,6 +123,54 @@ end
     @test d ≅ @transform(df, n1 = :i .* :g, n2 = :i .* :g)
 end
 
+@testset "@transform with @byrow" begin
+    df = DataFrame(
+        g = [1, 1, 1, 2, 2],
```
nalimilan, line 128: Keep on previous line? Same below.

test/dataframes.jl
```diff
@@ -123,6 +123,54 @@ end
     @test d ≅ @transform(df, n1 = :i .* :g, n2 = :i .* :g)
 end
 
+@testset "@transform with @byrow" begin
```
nalimilan, line 126: Maybe worth putting the new tests in a separate byrow.jl file? This one is getting pretty long.

src/macros.jl
```diff
@@ -503,6 +503,28 @@ and
...
+using broadcasting.
+
+To avoid writing `@byrow` multiple times when performing multiple
+transformations by row, `@transform` allows `@byrow` at the
+beginning of a block of transformations. All transformations
+in the block will operate by row.
```
nalimilan, line 526:
```suggestion
beginning of a block of transformations (i.e. `@byrow begin... end`).
All transformations in the block will operate by row.
```

nalimilan (commented):
Cool! I've made comments mainly about the docs. Overall I think it would make sense to put less stress on the technical details to target "basic" users.

I also wonder whether `@byrow z = :x * :y` shouldn't be allowed as it seems a bit simpler not to put `@byrow` in the middle of the transformation. It's also more consistent with `@byrow begin... end`. What do you think?

docs/src/index.md
pdeffebach: I will procrastinate this for a "jldoctest" PR.

docs/src/index.md
```diff
@@ -268,6 +268,228 @@ df2 = @eachrow df begin
...
+
+```julia
+df = DataFrame(a = [1, 2], b = [3, 4])
+```
+
+* Control flow. In all versions of Julia, expressions of the form 
+  `if...else`, `a ? b : c` cannot be broadcasted. In versions below
+  1.7-dev, expressions of the form `a && b` and `a || b` cannot be 
+  broadcasted. Consequently, the `@.` macro will fail when encountering such
+  control flow while `@byrow` will not.
```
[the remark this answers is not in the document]
pdeffebach: re-written.

src/macros.jl
pdeffebach: debugging, sorry.

docs/src/index.md
```diff
@@ -268,6 +268,228 @@ df2 = @eachrow df begin
...
+```julia
+tempfun(a, b) = a * b
+tempfun.(df.a, df.b)
+```
+
+In contrast to `@eachroe`, `@with` combined with `@byrow` will return a vector of the broadcasted multiplication and not a data frame.
+
+Additionally, `@eachrow` and `@eachrow!` allow modifying a data
+data frame. Just as with Base Julia broadcasting, `@byrow` will
+not update columns.
```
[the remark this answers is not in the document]
pdeffebach: This seems right.

src/macros.jl
pdeffebach: fixed everywhere.

test/dataframes.jl
pdeffebach: done.

test/dataframes.jl
pdeffebach: Will clean up along with other tests after this.

pdeffebach:
> I also wonder whether `@byrow z = :x * :y` shouldn't be allowed as it seems a bit simpler not to put `@byrow` in the middle of the transformation. It's also more consistent with `@byrow begin... end`. What do you think?

No, I don't think we should do this. It breaks the mapping to `src => fun => dest` logic. The thing on the RHS becomes the `fun`, which is where the `ByRow` goes. Plus we have to support `@byrow` on the RHS because the RHS gets sent to the same processing that stuff in `@where` goes (without no RHS). 

w.r.t. detailed technical docs vs introductory docs, if we were to make a docstring for `@byrow`, how would that work? We could make `@byrow` a macro the same as `@with df @byrow begin ... end`, but that seems a little redundant.

nalimilan:
> No, I don't think we should do this. It breaks the mapping to `src => fun => dest` logic. The thing on the RHS becomes the `fun`, which is where the `ByRow` goes. Plus we have to support `@byrow` on the RHS because the RHS gets sent to the same processing that stuff in `@where` goes (without no RHS).

Well the mapping isn't 1-to-1 already: while the LHS corresponds to `src`, the RHS completely mixes `fun` and `dest`. I get that in terms of implementation, `@byrow z = x + y` may be more complex to support, but in terms of API I'm not sure it's really inferior. (That said, it could be added later.)



> w.r.t. detailed technical docs vs introductory docs, if we were to make a docstring for `@byrow`, how would that work? We could make `@byrow` a macro the same as `@with df @byrow begin ... end`, but that seems a little redundant.
> 
> EDIT: Unfortunately `@byrow` doesn't error on the current release. So we have to have a few months where it errors before we can add `@byrow` back in as a real macro.

Can't you define a dummy macro that throws an error?

pdeffebach:
> `@byrow z = x + y` may be more complex to support, but in terms of API I'm not sure it's really inferior. (That said, it could be added later.)

Yeah. I guess a simpler argument is that we kind of need to support `z = @byrow :x + :y`, so while `@byrow z = :x + :z` is useful, we can't substitute out that syntax for the current one. 

> Can't you define a dummy macro that throws an error?

Yeah that sounds good. I will do that and switch the docstrings around.

pdeffebach:
Okay I'm going to allow 

```
@transform df begin 
    @byrow y = f(:x)
end
```

because with the `@transform df @byrow begin ... end` implemenation we allow 

```
@transform df @byrow y = f(:x)
```

so it would be weird to allow that and not the same for multiple arguments.

docs/Project.toml
```diff
@@ -1,6 +1,5 @@
 [deps]
-DataFramesMeta = "1313f7d8-7da2-5740-9ea0-a2ca25f37964"
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 
 [compat]
-Documenter = "0.25"
\ No newline at end of file
+Documenter = "0.25"
```
bkamins, line 5: maybe bump to 0.27?

pdeffebach: added.

pdeffebach:
Okay two substantive changes to the API

* `@transform df y = @byrow :x` now errors, in favor of `@transform df @byrow y = :x`
* Because a lot of this is implemented with recursion, I add a few checks to stop people from doing `@transform @byrow @byrow @byrow y = :x`

I added docstrings to all the relevant macros (except `@combine` and `@by`, since `@byrow` is useless there.) So in theory this PR is ready to be merged. All the work is done. 

Could use a review to go over the docstrings, I guess. But I'm getting anxious to merge this into master.

bkamins (approved):
Fantastic work.

src/parsing.jl
```diff
@@ -101,10 +108,23 @@ julia> src, fun = DataFramesMeta.get_source_fun(ex);
...
+:(ByRow(*))
+```
+
 """
-function get_source_fun(function_expr)
+function get_source_fun(function_expr; wrap_byrow=false)
     # recursive step for begin :a + :b end
-    if function_expr isa Expr &&
+    if is_macro_head(function_expr, "@byrow")
+        if wrap_byrow == true
```
nalimilan, line 123:
```suggestion
        if wrap_byrow
```

src/parsing.jl
```diff
@@ -101,10 +108,23 @@ julia> src, fun = DataFramesMeta.get_source_fun(ex);
...
+
+julia> src, fun = DataFramesMeta.get_source_fun(ex);
+
+julia> MacroTools.prettify(fun)
+:(ByRow(*))
+```
+
 """
-function get_source_fun(function_expr)
+function get_source_fun(function_expr; wrap_byrow=false)
```
nalimilan, line 120:
```suggestion
function get_source_fun(function_expr; wrap_byrow::Bool=false)
```

src/parsing.jl
```diff
@@ -168,6 +189,13 @@ function fun_to_vec(ex::Expr; nolhs::Bool = false, gensym_names::Bool = false)
     # cols(:y) = f(cols(:x)) # re-write as simple call, RHS is block, use cols
     # cols(y) = :x + 1 # re-write as complicated col, but RHS is :block
     # cols(:y) = cols(:x) + 1 # re-write as complicated call, RHS is block, use cols
+    # `@byrow` before any of the above
+    if is_macro_head(ex, "@byrow")
+        if wrap_byrow == true
```
nalimilan, line 194:
```suggestion
        if wrap_byrow
```

src/parsing.jl
```diff
@@ -142,15 +158,20 @@ function get_source_fun(function_expr)
...
+
+    return source, fun
 end
 
 # `nolhs` needs to be `true` when we have syntax of the form
 # `@combine(gd, fun(:x, :y))` where `fun` returns a `table` object.
 # We don't create the "new name" pair because new names are
 # given by the table.
-function fun_to_vec(ex::Expr; nolhs::Bool = false, gensym_names::Bool = false)
+function fun_to_vec(ex::Expr; nolhs::Bool = false, gensym_names::Bool = false, wrap_byrow=false)
```
nalimilan, line 174:
```suggestion
function fun_to_vec(ex::Expr; nolhs::Bool = false, gensym_names::Bool = false, wrap_byrow::Bool=false)
```

test/grouping.jl
```diff
@@ -378,6 +378,9 @@ end
 
 	@test @transform(g, t = :c).a ≅ d.a
 	@test @select(g, :a, t = :c).a ≅ d.a
+
+    @test @transform(g, @byrow t = :a ^ 2).t ≅ d.a .^ 2
+    @test @select(g, :a, @byrow t = :a ^ 2).t ≅ d.a .^ 2
```
nalimilan, line 383:
```suggestion
	@test @transform(g, @byrow t = :a ^ 2).t ≅ d.a .^ 2
	@test @select(g, :a, @byrow t = :a ^ 2).t ≅ d.a .^ 2
```

docs/src/index.md
```diff
@@ -18,6 +18,7 @@ In addition, DataFramesMeta provides
   convenient syntax
 * `@eachrow` and `@eachrow!` for looping through rows in data frame, again with high performance and 
   convenient syntax. 
+* `@byrow` for applying functions by-row of a data frame.
```
nalimilan, line 21:
```suggestion
* `@byrow` for applying functions to each row of a data frame (only supported inside other macros).
```

docs/src/index.md
```diff
@@ -268,6 +268,228 @@ df2 = @eachrow df begin
 end
 ```
 
+## Row-wise transformations with `@byrow`
```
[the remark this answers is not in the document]
nalimilan: I think the second paragraph still applies: it would be nice to start with a sentence or two saying that `@byrow` allows writing code that is applied to each row instead of having to vectorize it.

src/macros.jl
```diff
@@ -55,6 +55,231 @@ macro col(kw)
     esc(fun_to_vec(kw))
 end
 
+##############################################################################
+##
+## @byrow
+##
+##############################################################################
+"""
```
nalimilan, line 63:
```suggestion
"""
    @byrow

```

src/macros.jl
```diff
@@ -55,6 +55,231 @@ macro col(kw)
...
+Broadcast operations within DataFramesMeta.jl macros.
+
+`@byrow` is not a "real" Julia macro but rather serves as a "flag"
+to indicate that the anonymous function created by DataFramesMeta
+to represent an operation should be applied "by-row".
+
+If an expression starts with `@byrow`, either of the form `@byrow = f(:x)`
+in transformations or `@byrow f(:x)` in `@orderby`, `@where`, and `@with`,
+then the anonymous function created by DataFramesMeta is wrapped in the
+function-wrapped `DataFrames.ByRow`.
```
nalimilan, line 73:
```suggestion
`DataFrames.ByRow` function wrapper, which broadcasts the function so that it run on each row.
```

src/macros.jl
```diff
@@ -55,6 +55,231 @@ macro col(kw)
...
+1×2 DataFrame
+ Row │ a      b
+     │ Int64  Int64
+─────┼──────────────
+   1 │     2      4
+```
+
+### Comparison with `@eachrow`
+
+To re-cap, the `@eachrow` rougly transforms
```
nalimilan, line 116:
```suggestion
To re-cap, the `@eachrow` macro roughly transforms
```

src/macros.jl
```diff
@@ -55,6 +55,231 @@ macro col(kw)
...
+tempfun(a, b) = a * b
+tempfun.(df.a, df.b)
+```
+
+In contrast to `@eachrow`, `@with` combined with `@byrow` returns a vector of the
+broadcasted multiplication and not a data frame.
+
+Additionally, `@eachrow` and `@eachrow!` allow modifying a data
+data frame. Just as with Base Julia broadcasting, `@byrow` will
+not update columns.
```
nalimilan, line 162:
```suggestion
```suggestion 
Additionally, transformations applied using `@eachrow!` modify the input
data frame. On the contrary, `@byrow` does not update columns.
```
```

File: src/macros.jl [hunk 53]
```diff
@@ -55,6 +55,231 @@ macro col(kw)
...
+
+julia> @time @with df @byrow :a + expensive();
+  1.037073 seconds (51.67 k allocations: 3.035 MiB, 3.19% compilation time)
+
+julia> @time @with df :a .+ expensive();
+  0.539900 seconds (110.67 k allocations: 6.525 MiB, 7.05% compilation time)
+
+```
+
+This problem comes up when using the `@.` macro as well, but can easily be fixed with `\$`.
```
nalimilan, line 267:
```suggestion
  This problem comes up when using the `@.` macro as well,
  but can easily be fixed with `\$`.
```
(Have you checked that this looks OK after generating the manual? Code blocks inside lists are risky...)

pdeffebach (commented):
Thanks for the review!

I have incorporated the requested changes.

src/macros.jl
```diff
@@ -157,9 +159,9 @@ tempfun.(df.a, df.b)
 In contrast to `@eachrow`, `@with` combined with `@byrow` returns a vector of the
 broadcasted multiplication and not a data frame.
 
-Additionally, `@eachrow` and `@eachrow!` allow modifying a data
-data frame. Just as with Base Julia broadcasting, `@byrow` will
-not update columns.
+```suggestion
```
nalimilan, line 162:
```suggestion
```

nalimilan (approved):
Cool!

PR #251, opened 2021-06-04
Update index.md

tutlane:
Added good c# reference to learn. Please check it, if it is good, add it to the article.


bkamins:
This is a C# tutorial reference, while we are referring to LINQ in the sentence.

pdeffebach:
Yeah i don't think we need this. We have a link to something describing LINQ. We don't need to worry about C# at large.

PR #257, opened 2021-06-17
Make `@astable` flag for retuning a table

pdeffebach:
Instead of having `cols(AsTable) = ...`, I am implementing an `@astable` flag to indicate that an expression returns a `table`

```
@transform df @astable (a = :x, b = :y)
```

I initially started on implementing `@subset` but there was too much junk in the `make_source_fun` implementation to make that a clean PR. So I figured I would implement this feature first so make `@subset` easier.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -7,7 +7,7 @@ DataFrames = "a93c6f00-e57d-5684-b7b6-d8193f3e46c0"
 Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
 
 [compat]
-DataFrames = "0.21, 0.22, 1"
+DataFrames = "0.22, 1"
 Reexport = "0.2, 1"
 julia = "1"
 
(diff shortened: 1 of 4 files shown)


bkamins:
Bravo for the continued work. I have one question (related to your recent question in DataFrames.jl 😄). Do you think we can keep the design of DataFramesMeta.jl modular enough when adding the features to ensure we do not end up with spaghetti code like we have in DataFrames.jl?

pdeffebach:
I think so? This PR helps get rid of a decent amount of spaghetti code. 

More generally, I think we need a story for how these macro-flags should work. And modularize their use. My recursive trick I use a few places seems promising. But I think we can implement this PR without find the "right" way of doing things. 

Also, the end is in sight for DataFramesMeta, there are only a few more features to be added before 1.0, I think.

bkamins:
> I think we need a story for how these macro-flags should work.

This was exactly the reason why I have asked the question.

pdeffebach:
Some really annoying behavior here w.r.t. line number nodes and macros. I will file a more detailed issue soon.

pdeffebach:
Superceded by #261

PR #258, opened 2021-06-20
Add dependency on MacroTools to avoid `Base.remove_linenums!`

pdeffebach:
After #245, we have some `Base.remove_linenums!` scattered around. Initially I thought this was harmless, but it was not. See the following

```
julia> macro linenums_macro(arg)
           if arg isa Expr && arg.head == :block && length(arg.args) == 1 && arg.args[1] isa LineNumberNode
               esc(:(true))
           else
               esc(:(false))
           end
       end
@linenums_macro (macro with 1 method)
```

On master:

```
julia> df = DataFrame(a = [1], b = [2]);

julia> @transform(df, y = @linenums_macro begin end)
1×3 DataFrame
 Row │ a      b      y     
     │ Int64  Int64  Bool  
─────┼─────────────────────
   1 │     1      2  false
```

On this branch

```
julia> @transform(df, y = @linenums_macro begin end)
1×3 DataFrame
 Row │ a      b      y    
     │ Int64  Int64  Bool 
─────┼────────────────────
   1 │     1      2  true
```

Basically, macros inside `@transform` blocks might make assumptions about line number nodes, and we don't want to mess with that. I apologize for using `Base.remove_linenums!` earlier. 

To implement this, I finally bit the bullet and added a dependency on MacroTools.jl. It's just not worth re-inventing the wheel anymore when it comes to these sorts of issues. Every other macro package today has MacroTools as a dependency, so I might do so as well. 

We do remove line numbers still, but we only do it when `unblock`-ing an expression or working with `:block`-based expression, where as `Base.remove_linenums!` acts recursively. 

I have added a testset so this should be good to merge.

### DIFF
--- a/src/parsing.jl
+++ b/src/parsing.jl
@@ -126,7 +126,6 @@ function get_source_fun(function_expr; wrap_byrow::Bool=false)
             throw(ArgumentError("Redundant `@byrow` calls."))
         end
         return get_source_fun(function_expr.args[3], wrap_byrow=true)
-
     elseif is_simple_non_broadcast_call(function_expr)
         source = args_to_selectors(function_expr.args[2:end])
         fun_t = function_expr.args[1]
--- a/src/parsing.jl
+++ b/src/parsing.jl
@@ -118,17 +118,15 @@ julia> MacroTools.prettify(fun)
 
 """
 function get_source_fun(function_expr; wrap_byrow::Bool=false)
+    function_expr = MacroTools.unblock(function_expr)
+
     # recursive step for begin :a + :b end
     if is_macro_head(function_expr, "@byrow")
         if wrap_byrow
             throw(ArgumentError("Redundant `@byrow` calls."))
         end
         return get_source_fun(function_expr.args[3], wrap_byrow=true)
-    elseif function_expr isa Expr &&
-        function_expr.head == :block &&
-        length(function_expr.args) == 1
 
-        return get_source_fun(function_expr.args[1])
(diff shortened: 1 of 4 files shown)


pdeffebach:
I should ping @nalimilan and @bkamins here to clarify that this is ready for a review.

bkamins (approved):
I assume all is OK. I am not an expert in MacroTools.jl, but I agree with the direction of changes.

test/dataframes.jl
```diff
@@ -728,4 +728,47 @@ end
...
+
+    @test nrow(d) == 1
+
+    d = @where df begin
+        @byrow @linenums_macro begin end
+    end
+
+    @test nrow(d) == 1
+
+     d = @where df @byrow begin
```
nalimilan, line 767:
```suggestion
    d = @where df @byrow begin
```

src/parsing.jl
```diff
@@ -118,17 +118,15 @@ julia> MacroTools.prettify(fun)
...
+
     # recursive step for begin :a + :b end
     if is_macro_head(function_expr, "@byrow")
         if wrap_byrow
             throw(ArgumentError("Redundant `@byrow` calls."))
         end
         return get_source_fun(function_expr.args[3], wrap_byrow=true)
-    elseif function_expr isa Expr &&
-        function_expr.head == :block &&
-        length(function_expr.args) == 1
```
nalimilan, line 129:
```suggestion
```

nalimilan (approved):
Review challenge: only make suggestions on whitespace. :-D

PR #261, opened 2021-06-22
Re-factor macro-flag handling, fix Table-returning paths, only allow DataFrames 0.22 and above

pdeffebach:
This PR implements the following syntax

```
julia> df = DataFrame(a = [1, 2, 3], b = [4, 5, 6]);

julia> @transform df @astable (x = :a, y = :b)
3×4 DataFrame
 Row │ a      b      x      y     
     │ Int64  Int64  Int64  Int64 
   1 │     1      4      1      4
   2 │     2      5      2      5
   3 │     3      6      3      6
```

It also re-configures the way we treat macro flags so it will be easier to generalize flags in the future. 

I still need to add docs and tests.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -9,9 +9,9 @@ Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
 
 [compat]
 DataFrames = "0.21, 0.22, 1"
+MacroTools = "0.5"
 Reexport = "0.2, 1"
 julia = "1"
-MacroTools = "0.5"
 
 [extras]
 CategoricalArrays = "324d7699-5711-5eae-9e2f-1d82baa6b597"
--- a/Project.toml
+++ b/Project.toml
@@ -8,7 +8,7 @@ MacroTools = "1914dd2f-81c6-5fcd-8719-6d5c9610ff09"
 Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
 
 [compat]
-DataFrames = "0.21, 0.22, 1"
+DataFrames = "0.22, 1"
 MacroTools = "0.5"
 Reexport = "0.2, 1"
 julia = "1"
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -333,7 +333,7 @@ DataFrames.jl exports the data type `AsTable` which indicates that a transformat
 returns a Tables.jl-compatible object. DataFramesMeta.jl emulates this behavior
 with the macro-flag `@astable`. 
 
-For more information, see the [DataFramesMeta.@astable](@ref). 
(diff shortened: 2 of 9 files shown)


bkamins:
Thank you. Please let me know when it is good for a review.

pdeffebach:
Okay this is ready for a review. 

It's not obvious how useful this is, I guess. It doesn't do the automatic NamedTuple creation with `:x = ...` as in [DFMacros](https://github.com/jkrumbiegel/DataFrameMacros.jl#t-flag-macro-for-automatic-astable). 

But I guess it's less clunky than `cols(AsTable) = ...`.

bkamins:
Is the intended use case of this is something like:
```
df = DataFrame(a=[[1,2], [2,3]])
@select(df, @astable(:a))
```
?
(I am asking because most of the examples it tests seemed a bit artificial to me.

pdeffebach:
That example would work. 

To be clear, there is no benefit from `AsTable` that you can't get with `cols(AsTable) = ...` currently. 

I ended up adding this because I knew I needed to re-factor the table-producing paths before I implement `@select`. If we feel like this isn't a priority, I can change this PR to get rid of `@astable` but keep the good refactoring and the new macro flag logic.

pdeffebach:
I'm going to keep this as a re-factoring PR to keep moving forward on the more essential parts.

bkamins:
I understand that in the future it would be `$(AsTable)` right? In such case do you think there would be a big benefit of `@astable` over it? (and we could save introducing a new pseudo-macro)

CC @nalimilan

pdeffebach:
Okay this PR is ready for a review. 

There are some minor changes to how we handle `@byrow` and `@with` internally, but neither of these affect any user-facing APIs. 

The only thing I need a review for is the function `extract_macro_flags`, and whether the implementation isn't too complex. 

I am going to *not* add docs about `cols(AsTable)` usage right now, in order to move forward with `@subset`.

bkamins:
CI fails

pdeffebach:
Okay tests pass now. 

Basically, before, `@byrow` stuff was handled in two places, the function that makes the `src => fun` part and the more general expression handling function. Now the `@byrow` stuff is handled exclusively in the "expression handling" function.

pdeffebach:
Given this is just about internals and not the API, I will merge after just an approval from @bkamins

bkamins (approved):
Thank you! (I have not investigated every individual line, but I assume you are pretty sure about this internals refactor?)

pdeffebach:
Yes. At the very least, something similar was required before `@subset` so better be a separate PR for simplicity.

PR #262, opened 2021-06-23
Deprecate `@linq` in favor of re-exporting `@chain`. 

pdeffebach:


Project.toml
```diff
@@ -3,15 +3,17 @@ uuid = "1313f7d8-7da2-5740-9ea0-a2ca25f37964"
...
 MacroTools = "1914dd2f-81c6-5fcd-8719-6d5c9610ff09"
 Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
 
 [compat]
 DataFrames = "0.21, 0.22, 1"
+MacroTools = "0.5"
 Reexport = "0.2, 1"
 julia = "1"
-MacroTools = "0.5"
+Chain = "0.4"
```
bkamins, line 16: do we have CompatHelper installed for this repo?

bkamins (approved):

docs/src/index.md
```diff
@@ -496,26 +473,33 @@ x_thread = @chain df begin
 end
 ```
 
-Another alternative is Pipe.jl which exports the `@pipe` macro for piping. 
-The piping mechanism in Pipe requires explicit specification of the piped
-object via `_` instead of assuming it is the first argument to the next function.
-The Pipe.jl equivalent of the above is:
+By default, `@chain` places the argument returned from the
```
bkamins, line 476:
```suggestion
By default, `@chain` places the value of the 
```
(expression is by definition something that has a value; I expecially like this explanation https://fsharpforfunandprofit.com/posts/expressions-vs-statements/#:~:text=In%20programming%20language%20terminology%2C%20an,and%20doesn't%20return%20anything.)

pdeffebach:
@nalimilan would like your feedback on this. I know we've discussed it earlier and I think this is the direction you thought was good.

docs/src/index.md
```diff
@@ -496,26 +474,33 @@ x_thread = @chain df begin
...
+    map(_.a, _.x) do a, x
+        a + x
+    end
+end
+```
 
-df = DataFrame(a = repeat(1:5, outer = 20),
-               b = repeat(["a", "b", "c", "d"], inner = 25),
-               x = repeat(1:20, inner = 5))
+`@chain` also provides the `@aside` macro-flag to perform opertaions
```
nalimilan, line 491:
```suggestion
`@chain` also provides the `@aside` macro-flag to perform operations
```

docs/src/index.md
```diff
@@ -496,26 +474,33 @@ x_thread = @chain df begin
...
+and refer to the argument returned from the previous 
+expression.
 
 ```julia
-using Pipe, Statistics
+@chain df begin 
+    map(_.a, _.x) do a, x
+        a + x
+    end
+end
```
nalimilan, line 488: This example isn't super useful in practice. Maybe show a more realistic one? E.g. one where `_` is used after a few transformations?

src/linqmacro.jl
```diff
@@ -62,6 +62,9 @@ julia> @linq df |>
 ```
 """
 macro linq(arg)
+
+    @warn "@linq is deprecated. Use @chain instead. See ? @chain for details"
```
nalimilan, line 66: `depwarn` doesn't work in macros?
Also:
```suggestion
    @warn "@linq is deprecated. Use @chain instead. See ?@chain for details"
```

src/DataFramesMeta.jl
```diff
@@ -6,6 +6,8 @@ using MacroTools
 
 @reexport using DataFrames
 
+@reexport using Chain
```
nalimilan, line 9: Probably better only reexport `@chain` and `@aside` explicitly, so that we don't end up exporting more symbols than anticipated if Chain.jl changes.

docs/src/index.md
```diff
@@ -496,26 +474,33 @@ x_thread = @chain df begin
...
-x_thread = @pipe df |>
-    @transform(_, y = 10 * :x) |>
-    @where(_, :a .> 2) |>
-    @by(_, :b, meanX = mean(:x), meanY = mean(:y)) |>
-    @orderby(_, :meanX) |>
-    @select(_, :meanX, :meanY, var = :b)
+```julia
+@chain df begin 
+    @transform y = 10 .* :x
+    @aside y_mean = mean(_.y)
```
nalimilan, line 497: Mention `@aside` somewhere? Otherwise one might think it's a DataFramesMeta macro.

nalimilan (commented):
Yes the direction looks good!

src/DataFramesMeta.jl
pdeffebach: Thanks. Only adding `@chain` since `@aside` is an unexported macro-flag.

docs/src/index.md
pdeffebach: Added a better one.

docs/src/index.md
pdeffebach: Added a clarifying comment.

pdeffebach:
@nalimilan ready to merge?

src/linqmacro.jl
nalimilan: Do you confirm that `depwarn` cannot be used here?

pdeffebach: It can. But it says that the warning stems from `eval` at `boot.jl`, which I guess is fine.

Project.toml
pdeffebach: Yeah I will work on that.

nalimilan (approved):

pdeffebach:
Thank you!

PR #263, opened 2021-06-25
Add `@subset`

pdeffebach:
This is an initial attempt to add `@subset`. 

I was able to entirely replace `@where` by calling `skipmissing=true` as a keyword argument, which is great. 

We are in a real bind with regards to keyword arguments. With the move to using `:block` we can't just support keyword argument handling like `@subset(df, ...; skipmissing = true)`. So I added the flag `@skipmissing` and re-factored the macro-flags a little bit. 

```
julia> df = DataFrame(a = [1, missing], b = [3, 4]);

julia> @subset df @skipmissing begin 
           :a .== 1
           :b .== 3
       end
1×2 DataFrame
 Row │ a       b     
     │ Int64?  Int64 
─────┼───────────────
   1 │      1      3

```

But adding tests is still a pain, because the tests have `missing` so we would have to add `@skipmissing` everywhere. Is this a time when we should break with the DataFrames API and make `skipmissing=true` the default? That would make people's lives a bit easier when upgrading from `@where` to `@subset`. 

@nalimilan This is a design decision, so I would appreciate your input.


pdeffebach:
Okay, I've made `skipmissing=true` the default with subset and removed the `@skipmissing` flag. This seems like the easiest way forward. Unlike DataFrames.jl, we will treat `missing` as `false` in `@subset`. 


All docs are added and tests added and pass. So this could be merged.

src/macros.jl
```diff
@@ -415,52 +415,191 @@ macro with(d, body)
...
+
+and
+
+```
+@subset(df, :x .> 1, :y .< 2)
+```
+
+!!! note
+    `@subset` treats `missing` values as `false` when filtering rows.
+    Unlike `DataFrames.subset` and other boolean operations with
```
nalimilan, line 475:
```suggestion
    Unlike `DataFrames.subset` and other Boolean operations with
```

src/macros.jl
```diff
@@ -487,22 +626,22 @@ end
...
-@where(df, :x .> 1, :y .< 2)
+@subset!(df, :x .> 1, :y .< 2)
 ```
 
 !!! note
-    `@where` treats `missing` values as `false` when filtering rows.
-    Unlike `DataFrames.filter` and other boolean operations with
-    `missing`, `@where` will *not* error on missing values, and
+    `@subset!` treats `missing` values as `false` when filtering rows.
+    Unlike `DataFrames.subset!` and other boolean operations with
```
nalimilan, line 634:
```suggestion
    Unlike `DataFrames.subset!` and other Boolean operations with
```

src/parsing.jl
```diff
@@ -183,7 +183,7 @@ end
 # We need wrap_byrow as a keyword argument here in case someone
 # uses `@transform df @byrow begin ... end`, which we
 # deal with outside of this function.
-function fun_to_vec(ex::Expr; gensym_names::Bool=false, no_dest::Bool=false, wrap_byrow::Bool=false)
+function fun_to_vec(ex::Expr; gensym_names::Bool=false, outer_flags=nothing, no_dest::Bool=false)
```
nalimilan, line 186: Is this correct? Same below.
```suggestion
function fun_to_vec(ex::Expr;
                    gensym_names::Bool=false,
                    outer_flags::Union{NamedTuple, Nothing}=nothing,
                    no_dest::Bool=false)
```

src/parsing.jl
```diff
@@ -344,42 +345,16 @@ function create_args_vector(args...)
...
-with a `wrap_byrow` flag indicating that the
-expressions should operate by row.
-
-If `arg` is a single `:block`, it is unnested.
-Otherwise, return a single-element array.
-Also removes line numbers.
-
-If `arg` is of the form `@byrow ...`, then
-`wrap_byrow` is returned as `true`.
+   create_args_vector(arg) -> vec, outer_flags
```
nalimilan, line 348: Why remove the contents of the docstring?

test/dataframes.jl
```diff
@@ -763,13 +705,13 @@ end
     @test nrow(d) == 1
 
     d = @where df begin
```
nalimilan, line 707: Move this to deprecated.jl?

test/subset.jl
```diff
@@ -0,0 +1,143 @@
+module TestSubset
```
nalimilan, line 1: Can you add tests with `GroupedDataFrame`?

test/subset.jl
```diff
@@ -0,0 +1,143 @@
...
+    @test d ≅ @subset(df, :A .> 1, :B .> 1)
+end
+
+
+@testset "subset!" begin
+    df = DataFrame(A = [1, 2, 3, missing], B = [2, 1, 2, 1])
+
+    x = [2, 1, 0, 0]
+
+    @test @subset!(copy(df), :A .> 1) == df[(df.A .> 1) .=== true,:]
```
nalimilan, line 82: Better do this to really check the expected behavior:
```suggestion
    df2 = copy(df)
    @test @subset!(df2, :A .> 1) === df2
    @test df2 == df[(df.A .> 1) .=== true,:]
```

nalimilan (commented):
Thanks! That makes the API much more consistent with DataFrames.

Regarding `skipmissing`, I think it would be good to outline a general plan. Should DataFramesMeta automatically skip/propagate missing values everywhere? We discussed adding a keyword argument to do that in DataFrames at https://github.com/JuliaData/DataFrames.jl/issues/2314. It hasn't been implemented at this point, but it would make sense to decide whether we would like to enable it by default eventually in DataFramesMeta.

src/parsing.jl
pdeffebach: I will add a correct docstring.

test/subset.jl
pdeffebach: added, ported from `@where`.

test/dataframes.jl
pdeffebach: moved.

test/subset.jl
pdeffebach: Okay. Not changing all of them because the rest are just checking parsing.

pdeffebach:
w.r.t. `missing`s. 

I think that adding `transform!` with `SubDataFrame` goes along way to emulating Stata's `if` syntax. But you are right it doesn't help with `missing`s. 

I think something along the lines of [This PR](https://github.com/JuliaData/Missings.jl/pull/122) in Missings.jl is the solution. Since we are constructing anonymous functions we can just add a `spreadmissing(anon)` when we need to. I hope we can make it performant. I don't know if it should be default in case people compare the speed to `data.table`, but I am open to the idea. It may also help people who don't like row-wise since a lot of the benefit of row-wise functions is dealing with `missing`s. 

That's a long term strategy. Maybe in the meantime we should just continue to treat `missing`s as false since it's the default behavior with `@where` currently.

src/parsing.jl
```diff
@@ -304,7 +308,7 @@ function fun_to_vec(ex::Expr; gensym_names::Bool=false, no_dest::Bool=false, wra
 
     throw(ArgumentError("This path should not be reached"))
 end
-fun_to_vec(ex::QuoteNode; no_dest::Bool=false, gensym_names::Bool=false, wrap_byrow::Bool=false) = ex
+fun_to_vec(ex::QuoteNode; no_dest::Bool=false, gensym_names::Bool=false, outer_flags=nothing) = ex
```
nalimilan, line 311:
```suggestion
fun_to_vec(ex::QuoteNode;
           no_dest::Bool=false,
           gensym_names::Bool=false,
           outer_flags::Union{NamedTuple, Nothing}=nothing) = ex
```

src/parsing.jl
```diff
@@ -333,19 +336,20 @@ function replace_dotted!(e, membernames)
...
 end
 
 """
    create_args_vector(arg) -> vec, outer_flags
+
+Given an expression return a vector of operations
+and a `NamedTuple` of the macro-flags that appear
+in the expression.
+
+If a `:block` expression, returns the `args` of
```
nalimilan, line 350:
```suggestion
If a `:block` expression, return the `args` of
```

test/subset.jl
```diff
@@ -140,4 +142,20 @@ end
     @test d ≅ @subset!(copy(df), :A .> 1, :B .> 1)
 end
 
-end # module
\ No newline at end of file
+@testset "@subset with a grouped data frame" begin
```
nalimilan, line 145: Also test `@subset!` with `GroupedDataFrame`?

pdeffebach: Added!

pdeffebach:
Okay ready for merging.

nalimilan (approved):

pdeffebach:
Thank you!

PR #264, opened 2021-06-30
`:y = f(:x)` instead of `y = f(:x)`

pdeffebach:
In light of DataFrameMacros, this seems to be what people want. 

Benefits of `:y` on LHS

* More consistency, `QuoteNode`s *always* mean columns
* It's very hard to achieve that consistency without quoting everything, i.e. `y = f(x)` is hard to implement, and confuses scope. 

This will be a *major* PR in terms of updating tests and documentation. The actually change is very easy, so I will ping you when it's ready for a review.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -31,7 +31,7 @@ df = DataFrame(a = [1, 2], b = [3, 4]);
 transform(df, [:a, :b] => ((a, b) -> a .* b .+ first(a) .- sum(b)) => :c);
 
 # With DataFramesMeta
-@transform(df, c = :a .* :b .+ first(:a) .- sum(:b))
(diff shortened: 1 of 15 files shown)


bkamins:
As discussed - I understand it is not just `:y` on LHS but also strings, right?

pdeffebach:
I think the simplest rule for now is to have complete symmetry between LHS and RHS. So to use strings, you would have to escape the string, i.e. `$"A long column name" = f($"Another long column name")`. 

Where `$` replaces `cols` in a future PR>

pdeffebach:
Ready for a merging! As in, ready for a review, but with an approval I will merge.

bkamins (approved):

PR #266, opened 2021-07-08
Cols to dollar

pdeffebach:
With this PR, we no longer have `@transform df :y = f(cols("x"))`, but rather `@transform df :y = f($"x")`.

### DIFF
--- a/src/eachrow.jl
+++ b/src/eachrow.jl
@@ -9,8 +9,11 @@
 function eachrow_replace(e::Expr)
     # Traverse the syntax tree of e
     if onearg(e, :cols)
+        @warn "cols is deprecated, use \$ instead"
         # cols(:x) becomes cols(:x)[row]
         return Expr(:ref, Expr(:call, :cols, e.args[2]), :row)
+    elseif is_column_expr(e)
+        return Expr(:ref, Expr(:$, e.args[1]), :row)
     end
 
     if e.head == :.
--- a/src/eachrow.jl
+++ b/src/eachrow.jl
@@ -9,7 +9,7 @@
 function eachrow_replace(e::Expr)
     # Traverse the syntax tree of e
     if onearg(e, :cols)
-        @warn "cols is deprecated, use \$ instead"
+        @warn "cols(x) is deprecated, use \$x instead"
         # cols(:x) becomes cols(:x)[row]
         return Expr(:ref, Expr(:call, :cols, e.args[2]), :row)
     elseif is_column_expr(e)
--- a/src/eachrow.jl
+++ b/src/eachrow.jl
@@ -161,7 +161,7 @@ julia> varA = :A; varB = :B;
 
 julia> df2 = @eachrow df begin
            @newcol :colX::Vector{Float64}
-           :colX = cols(varB) == 2 ? pi * cols(varA) : cols(varB)
+           :colX = \$varB == 2 ? pi * \$varA : \$varB
        end
 3×3 DataFrame
  Row │ A      B      colX
@@ -310,7 +310,7 @@ julia> df2 = copy(df);
 
 julia> @eachrow! df2 begin
            @newcol :colX::Vector{Float64}
-           :colX = cols(varB) == 2 ? pi * cols(varA) : cols(varB)
+           :colX = \$varB == 2 ? pi * \$varA : \$varB
        end
 3×3 DataFrame
  Row │ A      B      colX
--- a/src/eachrow.jl
+++ b/src/eachrow.jl
@@ -102,8 +102,8 @@ by `@eachrow`. Also note that the returned data frame does not share columns
 with `df`. See [`@eachrow!`](@ref) which employs the same syntax but modifies
 the data frame in-place.
 
-Like with `@transform`, `@eachrow` supports the use of `cols` to work with column names
-stored as variables. Using `cols` with a multi-column selector, such as a `Vector` of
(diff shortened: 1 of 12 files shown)


pdeffebach:
Okay tests pass, so this is ready for a review. 

There are two problems that I am fine resolving later, after this is merged. 

1. For the purposes of tests, we still need `cols` occasionally, particularly in the `function_compilation` tests. Also, the `@linq` tests seem to error with `$` inside a `@testset`. Overall, this is strong evidence we need to keep `cols` around permanently. If users want to use DataFramesMeta.jl macros inside other macros, `cols` is a great way to get around the mine-field that is interpolation with `$` and it's many meanings. 
2. Broadcasting. In general, you can write 

```
x = rand(10); 
@. x - $mean(x)
```

But you can't do the same in DataFramesMeta.jl. `@. x - ^($)mean(:x)` does not even work. 

This is not a big deal, I guess. You can't have `$` work for every instance. But it's a bit of a bummer. 

We should still merge, though.

bkamins:
Do you know if the same issues with `$` are in DataFrameMacros.jl?

pdeffebach:
The same issues exist with DataFrameMacros.jl. I've spoken to @jkrumbiegel and he's fine with not working with every use-case of `$` out there. We should probably accept that, too, given how many different meanings it has throughout the ecosystem. 

It's probably something we can iterate on at a later date, possibly through nesting `$$`.

src/eachrow.jl
```diff
@@ -9,8 +9,11 @@
 function eachrow_replace(e::Expr)
     # Traverse the syntax tree of e
     if onearg(e, :cols)
+        @warn "cols is deprecated, use \$ instead"
```
nalimilan, line 12: Maybe more explicit?
```suggestion
        @warn "cols(x) is deprecated, use \$x instead"
```

src/parsing.jl
```diff
@@ -8,33 +8,40 @@ end
...
     replace_syms!(Meta.quot(q.value), membernames)
-replace_syms!(e::Expr, membernames) =
+function replace_syms!(e::Expr, membernames)
     if onearg(e, :^)
         e.args[2]
     elseif onearg(e, :_I_)
         @warn "_I_() for escaping variables is deprecated, use cols() instead"
         addkey!(membernames, :($(e.args[2])))
     elseif onearg(e, :cols)
+        @warn "cols() for escaping variables is deprecated, use \$ instead"
```
nalimilan, line 26:
```suggestion
        @warn "cols(x) for escaping variables is deprecated, use \$x instead"
```

src/parsing.jl
```diff
@@ -44,15 +51,18 @@ function is_simple_broadcast_call(expr::Expr)
...
-        all(x -> x isa QuoteNode || onearg(x, :cols), expr.args[2].args)
+        all(x -> x isa QuoteNode || onearg(x, :cols) || is_column_expr(x), expr.args[2].args)
 end
 
 function args_to_selectors(v)
     t = map(v) do arg
         if arg isa QuoteNode
             arg
         elseif onearg(arg, :cols)
+            @warn "cols is deprecated, use \$ instead"
```
nalimilan, line 62:
```suggestion
            @warn "cols(x) is deprecated, use \$x instead"
```

src/parsing.jl
```diff
@@ -189,20 +199,20 @@ function fun_to_vec(ex::Expr;
...
+    # :y = :x # Simple pair
+    # :y = $:x # Extract and return simple pair (no function)
+    # $:y = :x # Simple pair
+    # $:y = $:x # Simple pair
+    # :y = f(:x) # re-write as simple call
+    # :y = f($:x) # re-write as simple call, interpolation elsewhere
+    # :y = :x + 1 # re-write as complicated call
+    # :y = $:x + 1 # re-write as complicated call, interpolation elsewhere
+    # $:y = f(:x) # re-write as simple call, unblock extract function
+    # $:y = f(cols(:x)) # re-write as simple call, unblock, interpolation elsewhere
```
nalimilan, line 213:
```suggestion
    # $:y = f($:x) # re-write as simple call, unblock, interpolation elsewhere
```

src/parsing.jl
```diff
@@ -237,8 +253,19 @@ function fun_to_vec(ex::Expr;
...
+
+    # fix cols
+    if onearg(lhs, :cols)
+        lhs = Expr(:$, lhs.args[2])
+    end
+
     rhs = MacroTools.unblock(ex.args[2])
 
+    if onearg(rhs, :cols)
+        @warn "cols is deprecated, use \$ instead"
```
nalimilan, line 265:
```suggestion
        @warn "cols(x) is deprecated, use \$x instead"
```

nalimilan (approved):
Looks good, just a minor comment. The issues you mentioned are annoying, it would be nice to find a way to escape/protect `$` from other macros so that we don't have to keep `cols` forever.

pdeffebach:
Yes. I don't have any ideas on how to do that at the moment, but I will ask people who are more experienced in metaprogramming. Maybe John Myles White can return to his roots and help DataFramesMeta.jl again.

nalimilan:
> Maybe John Myles White can return to his roots and help DataFramesMeta.jl again.

Actually John hasn't contributed to this package at all. It was created by Tom Short.

pdeffebach:
Ah sorry! I got my packages confused. 

I'm going to merge! This will create merge conflicts I will have to sort out in #267

PR #278, opened 2021-08-01
Rename `@with` to `@withcols`

### LINKED ISSUE #252: Re-name `@with`
StaticModules.jl has a macro `@with` defined, and Infiltrator.jl is (maybe) defining a `@with`. 

Ultimately `@with` just seems like too general a name for a macro that is so data-frames specific. It's also hard to imagine a `With.jl` package we could base this on because `@with` creates an anonymous function whereas a type-stable object can get by with just a `let` block.

pdeffebach:
Resolves #252 . 

The main motivation for this change is [StaticModules.jl](https://github.com/MasonProtter/StaticModules.jl) which exports the `@with` macro which has a more intuitive functionality than our `@with`. It simply replaces references to fields within `x` to `x.`. 

If this package, or similar ones, take off, then `DataFramesMeta.@with` will be both less general and inferior, since we will still require `:x` instead of `x` everywhere. You won't be able to use `StaticModules.@with` with DataFrames because the type instability of columns. 

cc @masonprotter @bkamins  and for good measure @jkrumbiegel

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -16,7 +16,7 @@ In addition, DataFramesMeta provides
 * Row-wise versions of the above macros in the form of `@rtransform`, `@rtransform!`,
   `@rselect`, `@rselect!`, `@rorderby`, `@rsubset`, and `@rsubset!`.
 * `@by`, for grouping and combining a data frame in a single step
-* `@with`, for working with the columns of a data frame with high performance and 
+* `@withcols`, for working with the columns of a data frame with high performance and 
   convenient syntax
 * `@eachrow` and `@eachrow!` for looping through rows in data frame, again with high performance and 
   convenient syntax. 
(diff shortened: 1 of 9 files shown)


bkamins:
Having thought about this `@with` in DataFramesMeta.jl is mostly for interactive use. While StaticModules.jl is for normal coding.
I do not think it is very likely they would be both used.

Now IF they are both used then can you do
```
using DataFramesMeta
import StaticModules
```
?

And all will work (except you have to write `StaticModules.@with`)?

pdeffebach:
StaticModules is also very useful for interactive use I would say. If you have a typed table, it will work fine. Or a large named tuple. These are scenarios I find myself in at the REPL. 

The main thing is that StaticModules doesn't have very many users, I think. I'm not sure how relevant that is given the implementation is so good.

bkamins:
OK - let us wait for @MasonProtter to comment.

MasonProtter:
Sorry, I was away and missed this convo. I don't really use data frames all that often (though maybe I should). I don't think it's too big a deal to just make people who have both packages loaded to declare which macro they want. 

I don't anticipate too much trouble with both packages using the name, but who knows. I will say though, that I think in this context, `@withcols` is a more evocative name for what the macro is doing, but 🤷.

nalimilan:
I agree the risk of conflicts is pretty low. Also, names like `@select`, `@transform`, `@subset` are quite general too, so if we start this path there's almost no end to it...

pdeffebach:
Sounds good. Closing this.

PR #279, opened 2021-08-01
Add dplyr tutorial port, clean up docs so documenter is happy

pdeffebach:


docs/src/dplyr.md
```diff
@@ -0,0 +1,293 @@
...
+
+This is a port of the HarvardX series PH525x Genomics class tutorial on dplyr. View the original [here](https://genomicsclass.github.io/book/pages/dplyr_tutorial.html) and the source [here](https://github.com/genomicsclass/labs/blob/master/intro/dplyr_tutorial.Rmd).
+
+## What is DataFramesMeta.jl?
+
+DataFramesMeta.jl is a Julia package to transform and summarize tabular data. It provides a more convenient syntax to work with DataFrames from [DataFrames.jl](https://github.com/JuliaData/DataFrames.jl). For a deeper explanation of DataFramesMeta.jl, see the [documentation](https://github.com/JuliaData/DataFramesMeta.jl). 
+
+DataFramesMeta.jl is heavily inspired by R's `dplyr`. If you are familiar with `dplyr` this guide should get you up to speed with DataFramesMeta.jl. 
+
+## Why Is It Useful?
```
bkamins, line 11: capitalization of words in headings needs to be reviewed (not a native so I might miss some rules, but it seems that now it is inconsistent across headers)

docs/src/dplyr.md
```diff
@@ -0,0 +1,293 @@
+# DataFramesMeta.jl tutorial
+
+This is a port of the HarvardX series PH525x Genomics class tutorial on dplyr. View the original [here](https://genomicsclass.github.io/book/pages/dplyr_tutorial.html) and the source [here](https://github.com/genomicsclass/labs/blob/master/intro/dplyr_tutorial.Rmd).
+
+## What is DataFramesMeta.jl?
+
+DataFramesMeta.jl is a Julia package to transform and summarize tabular data. It provides a more convenient syntax to work with DataFrames from [DataFrames.jl](https://github.com/JuliaData/DataFrames.jl). For a deeper explanation of DataFramesMeta.jl, see the [documentation](https://github.com/JuliaData/DataFramesMeta.jl).
```
bkamins, line 7: maybe add that this is a [DSL](https://en.wikipedia.org/wiki/Domain-specific_language). The syntax is more convenient at the cost of syntax not being valid Julia code.

On the other hand DataFramesMeta.jl concepts try to mirror DataFrames.jl concepts (which is important I think for learning and using both)

docs/src/dplyr.md
```diff
@@ -0,0 +1,293 @@
...
+
+## What is DataFramesMeta.jl?
+
+DataFramesMeta.jl is a Julia package to transform and summarize tabular data. It provides a more convenient syntax to work with DataFrames from [DataFrames.jl](https://github.com/JuliaData/DataFrames.jl). For a deeper explanation of DataFramesMeta.jl, see the [documentation](https://github.com/JuliaData/DataFramesMeta.jl). 
+
+DataFramesMeta.jl is heavily inspired by R's `dplyr`. If you are familiar with `dplyr` this guide should get you up to speed with DataFramesMeta.jl. 
+
+## Why Is It Useful?
+
+Like dplyr, the DataFramesMeta.jl package contains a set of macros (or "verbs") that perform common data manipulation operations such as filtering for rows, selecting specific columns, re-ordering rows, adding new columns and summarizing data.
```
bkamins, line 13: I would add a link to dplyr

docs/src/dplyr.md
```diff
@@ -0,0 +1,293 @@
...
+
+DataFramesMeta.jl is a Julia package to transform and summarize tabular data. It provides a more convenient syntax to work with DataFrames from [DataFrames.jl](https://github.com/JuliaData/DataFrames.jl). For a deeper explanation of DataFramesMeta.jl, see the [documentation](https://github.com/JuliaData/DataFramesMeta.jl). 
+
+DataFramesMeta.jl is heavily inspired by R's `dplyr`. If you are familiar with `dplyr` this guide should get you up to speed with DataFramesMeta.jl. 
+
+## Why Is It Useful?
+
+Like dplyr, the DataFramesMeta.jl package contains a set of macros (or "verbs") that perform common data manipulation operations such as filtering for rows, selecting specific columns, re-ordering rows, adding new columns and summarizing data. 
+
+In addition, DataFramesMeta.jl contains a useful operation `@combine` to perform another common task which is the "split-apply-combine" concept. We will discuss that in a little bit.
```
bkamins, line 15: this sentence is not clear to me and seems more detailed than the previous. Especially as in the previous you have written "summarizing data".
Also - if you keep this maybe give a link to "split-apply-combine" so people reading it know what we mean (not all of them might know it)

docs/src/dplyr.md
```diff
@@ -0,0 +1,293 @@
...
+
+## Why Is It Useful?
+
+Like dplyr, the DataFramesMeta.jl package contains a set of macros (or "verbs") that perform common data manipulation operations such as filtering for rows, selecting specific columns, re-ordering rows, adding new columns and summarizing data. 
+
+In addition, DataFramesMeta.jl contains a useful operation `@combine` to perform another common task which is the "split-apply-combine" concept. We will discuss that in a little bit. 
+
+## How Does It Compare To Using Base Functions in Julia and in DataFrames.jl?
+
+If you are familiar with Julia, you are probably familiar with base Julia functions such  `map`, and `broadcast` (akin to `*apply` in R). These functions are convenient to use, but are designed to work with `Vector`s, not tabular data.
```
bkamins, line 19:
```suggestion
If you are familiar with Julia, you are probably familiar with base Julia functions such  `map`, and `broadcast` (akin to `*apply` in R). These functions are convenient to use, but are designed to work with arrays, not tabular data.
```

docs/src/dplyr.md
```diff
@@ -0,0 +1,293 @@
...
+
+DataFrames provides the functions `select`, `transform`, and more to work with data frames. Unlike `map` and `broadcast`, these functions are designed to work with tabular data, but have a complicated syntax.
```
bkamins, line 21:
```suggestion
DataFrames.jl provides the functions `select`, `transform`, and more to work with data frames. Unlike `map` and `broadcast`, these functions are designed to work with tabular data, but have a complicated syntax. 
```
also maybe add `combine` here for completeness

docs/src/dplyr.md
```diff
@@ -0,0 +1,293 @@
...
+
+In addition, DataFramesMeta.jl contains a useful operation `@combine` to perform another common task which is the "split-apply-combine" concept. We will discuss that in a little bit. 
+
+## How Does It Compare To Using Base Functions in Julia and in DataFrames.jl?
+
+If you are familiar with Julia, you are probably familiar with base Julia functions such  `map`, and `broadcast` (akin to `*apply` in R). These functions are convenient to use, but are designed to work with `Vector`s, not tabular data.
+
+DataFrames provides the functions `select`, `transform`, and more to work with data frames. Unlike `map` and `broadcast`, these functions are designed to work with tabular data, but have a complicated syntax. 
+
+DataFramesMeta.jl provides a convenient syntax for working with the vectors in a `DataFrame` so that you get all the convenience of Base Julia and DataFrames combined.
```
bkamins, line 23:
```suggestion
DataFramesMeta.jl provides a convenient syntax for working with the vectors in a data frame so that you get all the convenience of Base Julia and DataFrames combined. 
```

docs/src/dplyr.md
```diff
@@ -0,0 +1,293 @@
...
+DataFramesMeta.jl provides a convenient syntax for working with the vectors in a `DataFrame` so that you get all the convenience of Base Julia and DataFrames combined. 
+
+## How Do I Get DataFramesMeta.jl? 
+
+To install DataFramesMeta.jl:
+
+```julia
+import Pkg
+Pkg.activate(; temp=true) # activate a temprary environment for this tutorial
+Pkg.add("DataFramesMeta");
```
bkamins, line 32: probably installing DataFrames.jl would be a good idea too (even if it is not strictly required - I do not know what you want to show later, but e.g. `DataFrame` constructor might be useful)

docs/src/dplyr.md
```diff
@@ -0,0 +1,293 @@
...
+
+```@example 1
+using CSV, HTTP, Statistics
+```
+
+We will use [CSV.jl](https://csv.juliadata.org/stable/) and [HTTP.jl](https://juliaweb.github.io/HTTP.jl/stable/) for downloading our dataset from the internet.
+
+# Data: Mammals Sleep
+
+The `msleep` (mammals sleep) data set contains the sleep times and weights for a set of mammals and is available in the dagdata repository on github. This data set contains 83 rows and 11 variables.
```
bkamins, line 57:
```suggestion
The `msleep` (mammals sleep) data set contains the sleep times and weights for a set of mammals and is available in the dagdata repository on GitHub. This data set contains 83 rows and 11 variables.  
```

docs/src/dplyr.md
```diff
@@ -0,0 +1,293 @@
...
+using CSV, HTTP, Statistics
+```
+
+We will use [CSV.jl](https://csv.juliadata.org/stable/) and [HTTP.jl](https://juliaweb.github.io/HTTP.jl/stable/) for downloading our dataset from the internet.
+
+# Data: Mammals Sleep
+
+The `msleep` (mammals sleep) data set contains the sleep times and weights for a set of mammals and is available in the dagdata repository on github. This data set contains 83 rows and 11 variables.  
+
+We can loads the data directly into a DataFrame from the `url`.
```
bkamins, line 59:
```suggestion
We can load the data directly into a `DataFrame` from the `url`. 
```

docs/src/dplyr.md
bkamins: Ah - I see you re-export DataFrames.jl. Maybe mention it?

bkamins: also maybe somewhere in the top mention how much the user is assumed to know about DataFrames.jl. E.g. such things as how to refer to column names (that you can use `Symbol` or string)

docs/src/dplyr.md
```diff
@@ -0,0 +1,293 @@
...
+`:sleep_rem` | rem sleep, in hours
+`:sleep_cycle` | length of sleep cycle, in hours
+`:awake` | amount of time spent awake, in hours
+`:brainwt` | brain weight in kilograms
+`:bodywt` | body weight in kilograms
+
+
+# Important DataFramesMeta.jl Verbs To Remember
+
+dplyr verbs | Description
```
bkamins, line 85: why do you call them dplyr verbs?

docs/src/dplyr.md
```diff
@@ -0,0 +1,293 @@
...
+--- | ---
+`@select` | select columns
```
bkamins, line 87: I would put `@select` next to `@transform`

docs/src/dplyr.md
```diff
@@ -0,0 +1,293 @@
...
+dplyr verbs | Description
+--- | ---
+`@select` | select columns 
+`@subset` | filter rows
+`@orderby` | re-order or arrange rows
+`@transform` | create new columns
+`@combine` | summarise values
+`groupby` | allows for group operations in the "split-apply-combine" concept
+
+DataFramesMeta.jl also provides `@rselect`, `@rsubset`, `@rorderby`, and `@rtransform` for operations which act row-wise. We will expore the distinction between column-wise and row-wise transformations later in this turorial.
```
bkamins, line 94: maybe use term "whole-column" rather than "column-wise"? Alan Edelman was confused by "col-wise" (as it seems that one operation works vertically and the other horizontally which is not the case)

docs/src/dplyr.md
```diff
@@ -0,0 +1,293 @@
...
+
+## Selecting Columns Using `@select`
+
+Select a set of columns: the `:name` and the `:sleep_total` columns. 
+
+```@example 1
+sleepData = @select msleep :name :sleep_total
+```
+
+To select all the columns *except* a specific column, use the `Not` function for inverse selection. We preface the `Not` with `$` because it does not reference a column directly as a `Symbol`.
```
bkamins, line 108: the explanation of `$` is not clear. The reader is not clear what would happen if you skipped `$`.

docs/src/dplyr.md
```diff
@@ -0,0 +1,293 @@
...
+@select msleep $(Between(:name, :order))
+```
+
+To select all columns that start with the character string `"sl"` use regular expressions. 
+
+```@example 1
+@select msleep $(r"^sl")
+```
+
+Regular expressions are powerful, but can be difficult for new users to understand. Here are some quick tips.
```
bkamins, line 126: add a link maybe

docs/src/dplyr.md
```diff
@@ -0,0 +1,293 @@
...
+
+```@example 1
+@select msleep $(r"^sl")
+```
+
+Regular expressions are powerful, but can be difficult for new users to understand. Here are some quick tips.  
+
+1. `r"^abc"` = Starts with `"abc"`
+2. `r"abc$"` = Ends with `"abc"`
+3. `r"abc"` = Contains `"abc"` anywhere.
```
bkamins, line 130: natural questions:
1. how to select a column by its number?
2. can I use strings as column names

docs/src/dplyr.md
```diff
@@ -0,0 +1,293 @@
...
+In the above expression, the `.>=` means we "broadcast" the `>=` comparison across the whole column. We can use a simpler syntax, `@rsubset` which automatically broadcasts all operations.
+
+```@example 1
+@rsubset msleep :sleep_total > 16
+```
+
+Subset the rows for mammals that sleep a total of more than 16 hours *and* have a body weight of greater than 1 kilogram. For this we put multiple operations on separate lines in a single block. 
+
+```@example 1
+@rsubset msleep begin
```
bkamins, line 149: the reader is not clear if `@rsubset(msleep, :sleep_total >= 16, :bodywt >= 1)` would work or not.

docs/src/dplyr.md
```diff
@@ -0,0 +1,293 @@
...
+
+```@example 1
+@rsubset msleep begin 
+    :sleep_total >= 16 
+    :bodywt >= 1
+end
+```
+
+Filter the rows for mammals in the Perissodactyla and Primates taxonomic order
+
```
bkamins, line 156: maybe add `Set` wrapper, as this pattern is inefficient and I would prefer that it is not shown to users.

docs/src/dplyr.md
```diff
@@ -0,0 +1,293 @@
...
+Before we go any further, let's introduce the `@chain` macro from [Chain.jl](https://github.com/jkrumbiegel/Chain.jl), which is re-exported by DataFramesMeta.jl. `@chain` allows for you to pipe the output from one operation into the input of another operation.  The idea of piping is to read the functions from left to right. The syntax and design of `@chain` is very similar to `%>%` which users of dplyr are familiar with.
+
+To show it's usefulness, let's use `@select` and `@rsubset` one after the other. 
+
+```
+msleep_1 = @select msleep :name :sleep_total
+msleep_2 = @rsubset msleep_1 :sleep_total > 16
+```
+
+Now in this case, we will pipe the msleep data frame to the function that will select two columns (name and sleep\_total) and then pipe the new data frame to the `@rsubset` opertaion. This method involves awkwardly creating and naming temporary data frames. This can be avoided with `@chain`.
```
bkamins, line 174:
```suggestion
Now in this case, we will pipe the msleep data frame to the function that will select two columns (`:name` and `:sleep_total`) and then pipe the new data frame to the `@rsubset` operation. This method involves awkwardly creating and naming temporary data frames. This can be avoided with `@chain`.
```

docs/src/dplyr.md
```diff
@@ -0,0 +1,293 @@
...
+@orderby msleep :order
+```
+
+Now we will select three columns from msleep, arrange the rows by the taxonomic order and then arrange the rows by sleep\_total. Finally, keep the first 10 rows of the data frame.
+
+```@example 1
+@chain msleep begin 
+    @select :name :order :sleep_total
+    @orderby :order :sleep_total
+    first(10)
```
bkamins, line 203: why `first` has no `@` should be explained.

docs/src/dplyr.md
```diff
@@ -0,0 +1,293 @@
...
+```
+
+Something slightly more complicated: same as above, except arrange the rows in the `:sleep_total` column in a descending order. For this, use the function `sortperm` with the keyword argument `rev=true`.
+
+```@example 1
+@chain msleep begin 
+    @select :name :order :sleep_total
+    @orderby begin 
+        :order 
+        sortperm(:sleep_total, rev=true)
```
bkamins, line 224: is this really needed? there is no way to pass `rev` to `@orderby`?

docs/src/dplyr.md
```diff
@@ -0,0 +1,293 @@
...
+    @rsubset :sleep_total >= 16
+end
+```
+
+## Create New Columns Using `@transform` and `@rtransform`
+
+The `@transform` macro will add new columns to the data frame. Like with other macros, use `@rtransform` to operate row-wise. Create a new column called `:rem_proportion`, which is the ratio of rem sleep to total amount of sleep. 
+
+```@example 1
+@rtransform msleep :rem_proportion = :sleep_rem / :sleep_total
```
bkamins, line 235: I would also give `@transform` example to show the difference again.

docs/src/dplyr.md
```diff
@@ -0,0 +1,293 @@
...
+
+```@example 1
+@chain msleep @combine :avg_sleep = mean(:sleep_total)
+```
+
+There are many other summary statistics you could consider such `std`, `minimum`, `maximum`, `median`, `sum`, `length` (returns the length of vector), `first` (returns first value in vector), and `last` (returns last value in vector).
+
+```@example 1
+@combine msleep begin 
+    avg_sleep = mean(:sleep_total)
```
bkamins, line 259: don't you need `:` prefix in target column names? Maybe also explicitly comment on it?

docs/src/dplyr.md
```diff
@@ -0,0 +1,293 @@
...
+```@example 1
+@combine msleep begin 
+    avg_sleep = mean(:sleep_total) 
+    min_sleep = minimum(:sleep_total)
+    max_sleep = maximum(:sleep_total)
+    total = length(:sleep_total)
+end
+```
+
+## Group operations using `group`
```
bkamins, line 266:
```suggestion
## Group operations using `groupby`
```

docs/src/dplyr.md
```diff
@@ -0,0 +1,293 @@
...
+
+The `groupby` verb is an important function in DataFrames.jl (it does not live in DataFramesMeta.jl). As we mentioned before it's related to concept of "split-apply-combine". We literally want to split the data frame by some variable (e.g. taxonomic order), apply a function to the individual data frames and then combine the output.   
+
+Let's do that: split the `msleep` data frame by the taxonomic order, then ask for the same summary statistics as above. We expect a set of summary statistics for each taxonomic order. 
+
+```@example 1
+@chain msleep begin 
+    groupby(:order)
+    @combine begin 
+        avg_sleep = mean(:sleep_total)
```
bkamins, line 276: again missing `:`

log.txt
```diff
@@ -0,0 +1,11 @@
+Doctests: DataFramesMeta: Test Failed at /home/peterwd/.julia/packages/Documenter/oBZFM/src/Documenter.jl:870
```
bkamins, line 1: I would not put this log file in the PR

src/DataFramesMeta.jl
```diff
@@ -16,7 +16,7 @@ export @with,
        @transform, @select, @transform!, @select!,
        @rtransform, @rselect, @rtransform!, @rselect!,
        @eachrow, @eachrow!,
-       @byrow,
+       @byrow, @passmissing,
```
bkamins, line 19: I would not put changes in the core in docs PR.

bkamins (commented):
Looks good. I have left some mostly stylistic comments. Thank you!

docs/src/dplyr.md
pdeffebach: The [base tutorial](https://genomicsclass.github.io/book/pages/dplyr_tutorial.html) this came from uses the term "verb". I think the author likes the term because it sounds less technical than "function".

docs/src/dplyr.md
pdeffebach: Fixing this. But we should merge a PR special-casing `Not`, `Between`, `Regex`, and `r"..."` so we don't have to worry about this.

docs/src/dplyr.md
pdeffebach: It's even worse than that. The expression is doing 

```
(x -> x in ["a", "b"]).(df.order)
```

so a new vector is being allocated for every row...

And we can't even use `$` with `@.` because we use `$` for column references!

Not sure what to do to be honest. I don't want to do a complicated broadcast expression.

docs/src/dplyr.md
pdeffebach: No, we don't have keyword arguments yet.

docs/src/dplyr.md
pdeffebach: Added a little bit about strings on the LHS.

docs/src/dplyr.md
pdeffebach: Thanks. added.

docs/src/dplyr.md
pdeffebach: Added both examples.

docs/src/dplyr.md
pdeffebach: good point. Hopefully the language is clearer.

docs/src/dplyr.md
pdeffebach: Yes, I now note that they should be familiar with what a symbol is.

docs/src/dplyr.md
pdeffebach: Hopefully it is clearer now.

docs/src/dplyr.md
pdeffebach: Should be clearer now.

docs/src/dplyr.md
pdeffebach: Added an explanation.

docs/src/dplyr.md
pdeffebach: Added an example showing whole-column transform.

pdeffebach:
Thanks for the review! Should be much improved now.

docs/src/dplyr.md
bkamins: I am OK with `verb`, I am not clear why you use term "dplyr" - it seems these DataFramesMeta.jl verbs.

docs/src/dplyr.md
bkamins: what if you defined `cond = Set(["a", "b"])` as a variable and interpolated it?

docs/src/dplyr.md
pdeffebach: Oh that was a typo, sorry.

docs/src/dplyr.md
pdeffebach: No interpolation needed, it just goes right in. I will add that.

bkamins: but without interpolation one needs `const cond = ...` or wrap it in `let` to ensure type stability.

pdeffebach: Maybe I should introduce `@aside`? Would get rid of those problems.

bkamins: If it solves the issue - I would add it. Having said that - we can keep the pattern you originally propose (just comment it is not efficient, so it should be used only when one has small data)

pdeffebach: Okay I've wrapped in a `let` block. It's not ideal but i guess it's okay for now.

docs/src/dplyr.md
```diff
@@ -0,0 +1,355 @@
...
+```
+
+as well as vectors of variable names 
+
+```@repl 1
+varnames = ["name", "sleep_total"]
+@select msleep $varnames
+```
+
+Similarly, to select the first column, use the syntax `$1`.
```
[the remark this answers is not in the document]
pdeffebach: yes.

Right now, the parsing for selecting columns is exactly the same as working with anonymous functions. So since `@transform df :y = :x .+ 1` would be ambiguous if we allowed `1` to be a column selector in the anonymous function, we need the same thing when doing `select`. 

Not ideal, though. We can change this before 1.0.

docs/src/dplyr.md
```diff
@@ -0,0 +1,355 @@
...
+Many DataFrames.jl macros come in two forms, a version which operates on columns as a whole and a version which operations row-wise, prefixed by `r`.
+
+DataFramesMeta.jl macro | By-row version | Description | `dplyr` equivalent
+--- | --- | --- | ---
+`@select` | `@rselect`| select columns | `select`
+`@transform` | `@rtransform` | create new columns | `mutate`
+`@subset` | `@rsubset` | filter rows | `filter`
+`@orderby` | `@rorderby` | re-order or arrange rows | `arrange`
+`@combine` | | summarise values | `summarize`
+`groupby` | | allows for group operations in the "split-apply-combine" concept | `group_by`
```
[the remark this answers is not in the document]
pdeffebach: added.

docs/src/dplyr.md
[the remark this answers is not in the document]
bkamins: I guess it is :).

docs/src/dplyr.md
```diff
@@ -0,0 +1,357 @@
...
+```
+msleep_1 = @select msleep :name :sleep_total
+msleep_2 = @rsubset msleep_1 :sleep_total > 16
+```
+
+Now in this case, we will pipe the msleep data frame to the function that will select two columns (`:name` and `:sleep_total`) and then pipe the new data frame to the `@rsubset` opertaion. This method involves awkwardly creating and naming temporary data frames. We can avoid this with `@chain`.
+
+```@repl 1
+@chain msleep begin 
+  @select :name :sleep_total
```
bkamins, line 218: indentation is inconsistent

bkamins (approved):

docs/src/dplyr.md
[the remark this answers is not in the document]
bkamins: I think `$1` makes sense - it just probably should be well explained somewhere.

pdeffebach:
Thanks!

PR #284, opened 2021-08-08
Write up rules regarding multi-column indexing

pdeffebach:
You were correct, @bkamins , the rules are more complicated than I initially realized. So much that it's hard to put them all in a bulleted list. 

I think I have a good idea of the gaps and marked the appropriate places as unstable, and we can iteratively improve the code so the rules are clear and consistent.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -595,7 +595,7 @@ In summary
   In transformation operations, i.e. `@transform :y = f(:x)`, the same rules on the right hand side also apply to the left hand side. For example, `@transform $"y" = f(:x)` will work. 
 
 * Arguments wrapped entirely in `$` or `$()` are passed directly to the underlying DataFrames.jl functions. Because of this, *in addition to* the single-column selectors listed above, multi-argument selectors are allowed. These include, but are not limited to
-    * Vectors of `Symbol`s, `$[:x, :y]`, strings, `$["x", "y"]`, or `$[1, 2]`
+    * Vectors of `Symbol`s, `$[:x, :y]`, strings, `$["x", "y"]`, or integers `$[1, 2]`
     * Regular expressions, `$(r"^a")`
     * Filtering column selectors, such as `$(Not(:x))` and `$(Between(:a, :z))`
 
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -492,6 +492,14 @@ end
 end
 ```
 
+To reference columns with more complicated expressions, you must wrap column references in parentheses. 
+
+```
+@transform df :a + $("a column name" * " in two parts")
+@transform df :a + $(get_column_name(x))
+```
+
+
 ## Constructing multi-column arguments and `src => fun => dest` calls using `$`
 
 If an argument is entirely wrapped in `$()`, the result bypasses the anonymous function creation of DataFramesMeta.jl and is passed to the underling DataFrames.jl function directly. Importantly, this allows for multi-column selection and passing `src => fun => dest` calls from the DataFrames.jl "mini-language" directly.
@@ -510,7 +518,7 @@ which selects the columns `:a` and `:b` in the data frame. We can generate this 
 @select df $[:a, :b]
 ```
 
-Similarly, to selct all columns beginning with the letter `"a"`, wrap a regular expression in `$()`. 
+Similarly, to selct all columns beginning with the letter `"a"`, wrap a regular expression in `$()`. As mentioned above, because the regex is a complicated syntax, we need to wrap it in parentheses. 
 
 ```julia
 @select df $(r"^a")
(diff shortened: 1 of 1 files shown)


docs/src/index.md
```diff
@@ -492,7 +492,97 @@ end
...
+which selects the columns `:a` and `:b` in the data frame. We can generate this command in DataFramesMeta.jl with
+
+```julia
+@select df $[:a, :b]
+```
+
+Similarly, to selct all columns beginning with the letter `"a"`, wrap a regular expression in `$()`. 
+
+```julia
+@select df $(r"^a")
```
bkamins, line 516: why are `(` and `)` needed here and were not needed above?

docs/src/index.md
```diff
@@ -492,7 +492,97 @@ end
...
+```julia
+@select df :y = f($[:a, :b])
+```
+
+will fail. 
+
+Not all functions in DataFrames.jl allow for multi-column selectors, so detailed knowledge of the underlying functions DataFrames.jl may be required. For example, the call 
+
+```julia
+subset(df, [:a, :b])
```
bkamins, line 532: we potentially could allow multi-column selectors in `subset`, but I thought they would not be very useful in practice and the code would be more complicated. Do you think they should be added?

docs/src/index.md
```diff
@@ -492,7 +492,97 @@ end
...
+```
+
+will fail in DataFrames.jl, bcause `DataFrames.subset` does not support vectors of column names. Likewise, `@subset df $[:a, :b]` will fail. The macros which support multi-column selectors are 
+
+* `@select`
+* `@transform` multi-argument selectors are ignored
```
bkamins, line 538: rather they are no-op. They are processed, but do not have any effect.

docs/src/index.md
```diff
@@ -492,7 +492,97 @@ end
...
+* `@combine`
+* `@by`
+
+
+Because arguments wrapped entirely in `$()` get passed directly to underlying DataFrames.jl functions. This allows the use of the DataFrames.jl "mini-language" consisting of `src => fun => dest` pairs inside dataFramesMeta macros. For example, you can do the following:
```
bkamins, line 543:
```suggestion
Because arguments wrapped entirely in `$()` get passed directly to underlying DataFrames.jl functions. This allows the use of the DataFrames.jl "mini-language" consisting of `src => fun => dest` pairs inside DataFramesMeta.jl macros. For example, you can do the following:
```

docs/src/index.md
```diff
@@ -492,7 +492,97 @@ end
...
+
+!!! warning
+    The macros `@orderby` and `@with` do not transparently call underlying DataFrames.jl functions. Escaping entire transformations should be considered unstable and may change in future versions.
+
+!!! warning
+    Row-wise macros such as `@rtransform` and `@rsubset` will not automatically wrap functions in `src => fun => dest` in `ByRow`. 
+
+In summary
+
+* To refer to single columns, use `Symbol`s, i.e. `:x`
```
bkamins, line 581: what about referring to columns with:
* string literals
* variables bound to strings
* integer literals
* variables bound to integers
?

bkamins:
Thank you very much for the write up. This was my experience from DataFrames.jl that things are usually more complex than they initially apperar.

docs/src/index.md
pdeffebach: Added an explanation here and above.

docs/src/index.md
pdeffebach: No, I don't think it should be added. It's not intuitive that the columns should be "and"-ed.

docs/src/index.md
pdeffebach: Okay I have added a more detailed summary at the end. Should be better now.

docs/src/index.md
bkamins: OK. So we are on the same page.

bkamins:
@pdeffebach: Regarding your last two comments I think you have not pushed the changes to GitHub. Right?

docs/src/index.md
```diff
@@ -492,7 +492,119 @@ end
...
+* All arguments that are not *entirely* escaped with `$` or `$()` construct anonymous functions. Inside these expressions only single-column selectors are allowed. This includes
+    * `Symbol`s, i.e. `:x` and `:y`
+    * Strings, escaped with `$`, i.e. `$"A string"` or `$("A string with many" * "parts")`
+    * Integers, escaped with `$`, i.e. `$1`
+    * Any single-column variable representing one of the above, escaped with `$`, i.e. `$x`
+  
+  In transformation operations, i.e. `@transform :y = f(:x)`, the same rules on the right hand side also apply to the left hand side. For example, `@transform $"y" = f(:x)` will work. 
+
+* Arguments wrapped entirely in `$` or `$()` are passed directly to the underlying DataFrames.jl functions. Because of this, *in addition to* the single-column selectors listed above, multi-argument selectors are allowed. These include, but are not limited to
+    * Vectors of `Symbol`s, `$[:x, :y]`, strings, `$["x", "y"]`, or `$[1, 2]`
```
pdeffebach, line 598:
```suggestion
    * Vectors of `Symbol`s, `$[:x, :y]`, strings, `$["x", "y"]`, or integers `$[1, 2]`
```

pdeffebach:
Sorry about that. Updated now.

bkamins (approved):
Great - now I think I understand `$`.

PR #285, opened 2021-08-10
Fix escaping `$`

pdeffebach:
Rules: 

Inside `.md` files, you can use `$` however much you want. 

Inside docstrings, we have the global variable `const DOLLAR = raw"$"`. We interpolate `$DOLLAR` everywhere we want a raw `$`, which is ironic.

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -15,7 +15,7 @@ DataFrames's `source => fun => destination` syntax.
 Parsing follows the same convention as other DataFramesMeta.jl macros, such as `@with`. All
 terms in the expression that are `Symbol`s are treated as columns in the data frame, except
 `Symbol`s wrapped in `^`. To use a variable representing a column name, wrap the variable
-in `\$`.
+in `$DOLLAR`.
 
 `@col` constructs an anonymous function `fun` based on the given expression. It then creates
 a `source => fun => destination` pair that is suitable for the `select`, `transform`, and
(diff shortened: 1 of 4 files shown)


pdeffebach:
@bkamins PR is ready for a review. I built the docs and things look good.

src/macros.jl
```diff
@@ -266,17 +266,17 @@ julia> @time @with df :a .+ expensive();
 ```
 
   This problem comes up when using the `@.` macro as well,
-  but can easily be fixed with `\$`.
+  but can easily be fixed with `$DOLLAR`. Because $DOLLAR is currently
```
bkamins, line 269:
```suggestion
  but can easily be fixed with `$DOLLAR`. Because `$DOLLAR` is currently
```

src/macros.jl
```diff
@@ -1550,7 +1550,7 @@ function combine_helper(x, args...; deprecation_warning = false)
         !(fe isa QuoteNode || onearg(fe, :cols) || is_column_expr(fe)) &&
         !(fe.head == :(=) || fe.head == :kw)
 
-        @warn "Returning a Table object from @by and @combine now requires `\$AsTable` on the LHS."
+        @warn "Returning a Table object from @by and @combine now requires `$DOLLARAsTable` on the LHS."
```
bkamins, line 1553:
```suggestion
        @warn "Returning a Table object from @by and @combine now requires `$(DOLLAR)AsTable` on the LHS."
```

bkamins (approved):
Looks good. I have made two small suggestions.

PR #291, opened 2021-08-21
add astable macro

pdeffebach:
Not yet ready for a review. 

To simulate `@t` from DataFrameMacros

```
julia> df = DataFrame(a = rand(5), b = rand(5));

julia> @transform df @astable begin 
           :x = :b
           $"hello" = :a
       end
5×4 DataFrame
 Row │ a         b         x         hello    
     │ Float64   Float64   Float64   Float64  
─────┼────────────────────────────────────────
   1 │ 0.495296  0.788492  0.788492  0.495296
   2 │ 0.935795  0.895012  0.895012  0.935795
   3 │ 0.340286  0.32458   0.32458   0.340286
   4 │ 0.911353  0.585055  0.585055  0.911353
   5 │ 0.74855   0.655892  0.655892  0.74855
```

### DIFF
--- a/src/parsing.jl
+++ b/src/parsing.jl
@@ -36,6 +36,33 @@ function replace_syms!(e::Expr, membernames)
     end
 end
 
+is_col_lhs(x) = false
+function is_col_lhs(x::Expr)
+    if x.head == :(=)
+        xa = x.args[1]
+        return xa isa QuoteNode || is_column_expr(xa)
+    else
+        return false
+    end
+end
+
+function modifiy_column_assignments(expr)
+    membernames = Dict{Any, Symbol}()
+
+    new_expr = MacroTools.postwalk(expr) do x
+        if x isa Expr && x.head == :(=)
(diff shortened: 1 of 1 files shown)


pdeffebach:
A note on the API: As you may know I also maintain [AddToField.jl](https://github.com/pdeffebach/AddToField.jl/) with syntax

```
@addnt begin 
    @add x = 1
    @add y = 5
end
```

To make the named tuple `(x = 1, y = 5)`. 

This functionality is similar, but a bit more DataFrames specific. It allows for consistency by keeping `:x` and `$x` referring to columns in the DataFrame. We can't do that in an external macro because we need to treat the column identifiers on the LHS in a special way.

pdeffebach:
Okay I'm thinking I will need to port over more of DataFrameMacros.jl's parsing code, in particular the stuff that uses `MacroTools.postwalk`. I will think more about this tomorrow.

bkamins:
Thank you for working on this!

pdeffebach:
Closed in favor of #298

PR #294, opened 2021-09-07
Add exclamations

### LINKED ISSUE #293: `@rtransform!` and `@rselect!` aren't actually modifying dataframe
```
julia> df = DataFrame(a = 1, b = 2)
1×2 DataFrame
 Row │ a      b     
     │ Int64  Int64 
─────┼──────────────
   1 │     1      2

julia> xx = ["a", "b"];

julia> for x in xx
           @rtransform! df $x = $x + 100
       end

julia> df
1×2 DataFrame
 Row │ a      b     
     │ Int64  Int64 
─────┼──────────────
   1 │     1      2
```

pdeffebach:
Fixes #293 

@bkamins @nalimilan please review and merge!

Tests pass. This is a major oversight and needs to be fixed quickly. 

Thank you!

### DIFF
--- a/.github/workflows/Documenter.yml
+++ b/.github/workflows/Documenter.yml
@@ -12,9 +12,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - uses: julia-actions/setup-julia@latest
+      - uses: julia-actions/julia-buildpkg@latest
+      - uses: julia-actions/julia-docdeploy@latest
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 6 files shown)


bkamins (approved):
I am approving to speed up process.
You need to bump the version and update the docs before merging (I will do this if you do not do this soon).

bkamins:
There is some issue with Documenter in CI. But when I run it locally it works.

pdeffebach:
If I merge and the build fails, will it use the old version of the docs?

bkamins:
I think so. Let us try to make Documenter work for a bit more 😄.

pdeffebach:
Jacob's snippet worked. 

Thanks for the fast turnaround! Registration [here](https://github.com/JuliaRegistries/General/pull/44464).

PR #298, opened 2021-09-14
Another attempt at an astable flag

pdeffebach:
This attempt should be a more robust strategy than the current `@t` flag in DataFrameMacros.jl. 

I will outline it in more detail shortly.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -404,7 +404,7 @@ new variables in the same operation, yet have them share
 information. 
 
 In a single block, all assignments of the form `:y = f(:x)` 
-or `$y = f(:x)` at the top-level generate new columns. In the 2nd example, `y`
+or `$y = f(:x)` at the top-level generate new columns. In the second example, `y`
 must be a string or `Symbol`. 
 
 ```
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -405,7 +405,7 @@ information.
 
 In a single block, all assignments of the form `:y = f(:x)` 
 or `$y = f(:x)` at the top-level generate new columns. In the 2nd example, `y`
-must be a string, `Symbol`. 
+must be a string or `Symbol`. 
 
 ```
 julia> df = DataFrame(a = [1, 2, 3], b = [400, 500, 600]);
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -22,6 +22,7 @@ In addition, DataFramesMeta provides
   convenient syntax. 
 * `@byrow` for applying functions to each row of a data frame (only supported inside other macros).
 * `@passmissing` for propagating missing values inside row-wise DataFramesMeta.jl transformations.
+* `@astable` to create multiple columns within a single transformation.
 * `@chain`, from [Chain.jl](https://github.com/jkrumbiegel/Chain.jl) for piping the above macros together, similar to [magrittr](https://cran.r-project.org/web/packages/magrittr/vignettes/magrittr.html)'s
   `%>%` in R. 
 
@@ -396,11 +397,37 @@ julia> @rtransform df @passmissing x = parse(Int, :x_str)
    3 │ missing  missing
 ```
 
+## Creating multiple columns at once with `@astable`
+
+Often new variables may depend on the same intermediate calculations. `@astable` makes it easy to create multiple
+new variables in the same operation, yet have them share
+information. 
+
+In a single block, all assignments of the form `:y = f(:x)` 
+or `$y = f(:x)` at the top-level are generate new columns.
+
+```
+julia> df = DataFrame(a = [1, 2, 3], b = [400, 500, 600]);
+
+julia> @transform df @astable begin 
+           ex = extrema(:b)
+           :b_first = :b .- first(ex)
+           :b_last = :b .- last(ex)
+       end
+3×4 DataFrame
+ Row │ a      b      b_first  b_last 
+     │ Int64  Int64  Int64    Int64  
+─────┼───────────────────────────────
+   1 │     1    400        0    -200
+   2 │     2    500      100    -100
+   3 │     3    600      200       0
+```
+
+
 ## [Working with column names programmatically with `$`](@id dollar)
 
 DataFramesMeta provides the special syntax `$` for referring to 
-columns in a data frame via a `Symbol`, string, or column position as either
-a literal or a variable. 
+columns in a data frame via a `Symbol`, string, or column position as either a literal or a variable. 
 
 ```julia
 df = DataFrame(A = 1:3, B = [2, 1, 2])
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -404,7 +404,8 @@ new variables in the same operation, yet have them share
 information. 
 
 In a single block, all assignments of the form `:y = f(:x)` 
-or `$y = f(:x)` at the top-level are generate new columns.
+or `$y = f(:x)` at the top-level generate new columns. In the 2nd example, `y`
+must be a string, `Symbol`. 
 
 ```
 julia> df = DataFrame(a = [1, 2, 3], b = [400, 500, 600]);
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -385,7 +385,7 @@ function f(a)
     (; x = x_t, z = z_t)
 end
 
-transform(df, [:a] => f => AsTable)
+transform(df, [:a] => ByRow(f) => AsTable)
 ```
 
 `@astable` is useful when performing intermediate calculations
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -282,11 +282,10 @@ macro byrow(args...)
     throw(ArgumentError("@byrow is deprecated outside of DataFramesMeta macros."))
 end
 
-
 """
     @passmissing(args...)
 
-Propograte missing values inside DataFramesMeta.jl macros.
+Propagrate missing values inside DataFramesMeta.jl macros.
 
 
 `@passmissing` is not a "real" Julia macro but rather serves as a "flag"
@@ -357,7 +356,7 @@ Return a `NamedTuple` from a single transformation inside the DataFramesMeta.jl
 macros, `@select`, `@transform`, and their mutating and row-wise equivalents.
 
 `@astable` acts on a single block. It works through all top-level expressions
-and collects all such expressions of the form `:y = ...` or `$y = ...`, i.e. assignments to a
+and collects all such expressions of the form `:y = ...` or `$(DOLLAR)y = ...`, i.e. assignments to a
 `Symbol` or an escaped column identifier, which is a syntax error outside of
 DataFramesMeta.jl macros. At the end of the expression, all assignments are collected
 into a `NamedTuple` to be used with the `AsTable` destination in the DataFrames.jl
@@ -427,7 +426,7 @@ Column assignment in `@astable` follows similar rules as
 column assignment in other DataFramesMeta.jl macros. The left-
 -hand-side of a column assignment can be either a `Symbol` or any
 expression which evaluates to a `Symbol` or `AbstractString`. For example
-`:y = ...`, and `$y = ...` are both valid ways of assigning a new column.
+`:y = ...`, and `$(DOLLAR)y = ...` are both valid ways of assigning a new column.
 However unlike other DataFramesMeta.jl macros, multi-column assignments via
 `AsTable` are disallowed. The following will fail.
 
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -284,7 +284,7 @@ end
 
 
 """
-    passmissing(args...)
+    @passmissing(args...)
 
 Propograte missing values inside DataFramesMeta.jl macros.
 
@@ -353,7 +353,8 @@ end
 """
     @astable(args...)
 
-Return a `NamedTuple` from a single transformation inside DataFramesMeta.jl macros.
+Return a `NamedTuple` from a single transformation inside the DataFramesMeta.jl
+macros, `@select`, `@transform`, and their mutating and row-wise equivalents.
 
 `@astable` acts on a single block. It works through all top-level expressions
 and collects all such expressions of the form `:y = ...`, i.e. assignments to a
@@ -424,7 +425,10 @@ end
 
 Column assignment in `@astable` follows the same rules as
 column assignment more generally. Construct a new column
-from a string by escaping it with `$DOLLAR`.
+from a string by escaping it with `$DOLLAR`, which can be a
+`Symbol` or an `AbstractString`. References to existing
+columns may be a `Symbol`, `AbstractString`, or an
+integer.
 
 ### Examples
 
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -427,15 +427,16 @@ julia> d = @rtransform df @astable begin
 julia> df = DataFrame(a = [1, 1, 2, 2], b = [5, 6, 70, 80]);
 
 julia> @by df :a @astable begin
-           $(DOLLAR)"Mean of b" = mean(:b)
-           $(DOLLAR)"Standard deviation of b" = std(:b)
+            ex = extrema(:b)
+            :min_b = first(ex)
+            :max_b = last(ex)
        end
 2×3 DataFrame
- Row │ a      Mean of b  Standard deviation of b
-     │ Int64  Float64    Float64
-─────┼───────────────────────────────────────────
-   1 │     1        5.5                 0.707107
-   2 │     2       75.0                 7.07107
+ Row │ a      min_b  max_b
+     │ Int64  Int64  Int64
+─────┼─────────────────────
+   1 │     1      5      6
+   2 │     2     70     80
 ```
 
 """
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -353,7 +353,7 @@ end
 """
     astable(args...)
 
-Return a `NamedTuple` from a transformation inside DataFramesMeta.jl macros.
+Return a `NamedTuple` from a single transformation inside DataFramesMeta.jl macros.
 
 `@astable` acts on a single block. It works through all top-level expressions
 and collects all such expressions of the form `:y = ...`, i.e. assignments to a
@@ -388,8 +388,24 @@ end
 transform(df, [:a] => ByRow(f) => AsTable)
 ```
 
-`@astable` is useful when performing intermediate calculations
-yet store their results in new columns. For example, the following fails.
+`@astable` has two major advantages at the cost of increasing complexity.
+First, `@astable` makes it easy to create multiple columns from a single
+transformation, which share a scope. For example, `@astable` allows
+for the following
+
+```
+@transform df @astable begin
+    m = mean(:x)
+    :x_demeaned = :x .- m
+    :x2_demeaned = :x2 .- m
+end
+```
+
+The creation of `:x_demeaned` and `:x2_demeaned` both share the variable `m`,
+which does not need to be calculated twice.
+
+Second, `@astable` is useful when performing intermediate calculations
+and storing their results in new columns. For example, the following fails.
 
 ```
 @rtransform df begin
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -285,7 +285,7 @@ end
 """
     @passmissing(args...)
 
-Propagrate missing values inside DataFramesMeta.jl macros.
+Propagate missing values inside DataFramesMeta.jl macros.
 
 
 `@passmissing` is not a "real" Julia macro but rather serves as a "flag"
@@ -349,8 +349,8 @@ macro passmissing(args...)
     throw(ArgumentError("@passmissing only works inside DataFramesMeta macros."))
 end
 
-global astable_docstring_snippet = """
-    Transformations can also use the macro-flag `@astable` for creating multiple
+const astable_docstring_snippet = """
+    Transformations can also use the macro-flag [`@astable`](@ref) for creating multiple
     new columns at once and letting transformations share the same name-space.
     See `? @astable` for more details.
     """
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -351,7 +351,7 @@ macro passmissing(args...)
 end
 
 """
-    astable(args...)
+    @astable(args...)
 
 Return a `NamedTuple` from a single transformation inside DataFramesMeta.jl macros.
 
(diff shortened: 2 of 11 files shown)


pdeffebach:
More progress! Not ready for a review yet as i have not added a complete test set or begun documenting.

pdeffebach:
Ready for a review!

This is a big PR. In the process, I have to finally remove some deprecated functionality for grouped data frames. In particular

```
@by df :a (n1 = :a, n2 = :b)
```

used to work. In the past release we had a warning that this needed to be 

```
@by df :a $AsTable = (n1 = :a, n2 = :b)
```

and now we throw an error in the original expression. 

Here is the docstring for `@astable`

---

## `astable(args...)`

Return a `NamedTuple` from a transformation inside DataFramesMeta.jl macros.

`@astable` acts on a single block. It works through all top-level expressions
and collects all such expressions of the form `:y = x`, i.e. assignments to a
`Symbol`, which is a syntax error outside of the macro. At the end of the
expression, all assignments are collected into a `NamedTuple` to be used
with the `AsTable` destination in the DataFrames.jl transformation
mini-language.

Concretely, the expressions

```
df = DataFrame(a = 1)

@rtransform df @astable begin
    :x = 1
    y = 50
    :z = :x + y + :a
end
```

becomes the pair

```
function f(a)
    x_t = 1
    y = 50
    z_t = x_t + y + a

    (; x = x_t, z = z_t)
end

transform(df, [:a] => f => AsTable)
```

`@astable` is useful when performing intermediate calculations
yet store their results in new columns. For example, the following fails.

```
@rtransform df begin
    :new_col_1 = :x + :y
    :new_col_2 = :new_col_1 + :z
end
```

This because DataFrames.jl does not guarantee sequential evaluation of
transformations. `@astable` solves this problem

@rtransform df @astable begin
    :new_col_1 = :x + :y
    :new_col_2 = :new_col_1 + :z
end

### Examples

```
julia> df = DataFrame(a = [1, 2, 3], b = [4, 5, 6]);

julia> d = @rtransform df @astable begin
           :x = 1
           y = 5
           :z = :x + y
       end
3×4 DataFrame
 Row │ a      b      x      z
     │ Int64  Int64  Int64  Int64
   1 │     1      4      1      6
   2 │     2      5      1      6
   3 │     3      6      1      6

julia> df = DataFrame(a = [1, 1, 2, 2], b = [5, 6, 70, 80]);

julia> @by df :a @astable begin
           $(DOLLAR)"Mean of b" = mean(:b)
           $(DOLLAR)"Standard deviation of b" = std(:b)
       end
2×3 DataFrame
 Row │ a      Mean of b  Standard deviation of b
     │ Int64  Float64    Float64
─────┼───────────────────────────────────────────
   1 │     1        5.5                 0.707107
   2 │     2       75.0                 7.07107
```

---

This implementation is more complicated than that of `@t` from DataFrameMacros.jl. In DataFrameMacros.jl, the following will fail

```
df = DataFrame(a = 1)
@transform df @t begin 
    :x = 1
    b + :x 
end
```

this is because it sends `:x` to `src` in `src => fun => dest` even though it doesn't exist in the DataFrame. `@astable` does not have this problem, at the cost of a more complicated implementation. 

cc @bkamins @nalimilan

bkamins:
> `transform(df, [:a] => f => AsTable)`

I would think it should be `transform(df, [:a] => ByRow(f) => AsTable)`?

pdeffebach:
Good catch. Will update.

src/macros.jl
```diff
@@ -350,6 +350,99 @@ macro passmissing(args...)
...
+   1 │     1      4      1      6
+   2 │     2      5      1      6
+   3 │     3      6      1      6
+
+julia> df = DataFrame(a = [1, 1, 2, 2], b = [5, 6, 70, 80]);
+
+julia> @by df :a @astable begin
+           $(DOLLAR)"Mean of b" = mean(:b)
+           $(DOLLAR)"Standard deviation of b" = std(:b)
+       end
```
nalimilan, line 432: This example can be achieved without `@astable`, right? Maybe do `m = mean(:b); std(:b, mean=m)` to illustrate  the power of this function? Or, simpler, call `extrema(:b)` to create two columns.

Also, I wouldn't use long column names with spaces in them: better illustrate a single feature at a time.

src/parsing_astable.jl
```diff
@@ -0,0 +1,95 @@
+function conditionally_add_symbols!(inputs_to_function, lhs_assignments, col)
```
nalimilan, line 1: How about making the signature more explicit? At least, `lhs_assignments::AbstractDict` would make things clearer. If you rely on the order, better even use `::OrderedCollections.OrderedDict`. Same below.

src/parsing_astable.jl
```diff
@@ -0,0 +1,95 @@
...
+    # if it's already been assigned at top-level,
+    # don't add it to the inputs
+    if haskey(lhs_assignments, col)
+        return lhs_assignments[col]
+    else
+        return addkey!(inputs_to_function, col)
+    end
```
nalimilan, line 8:
```suggestion
    return get!(lhs_assignments, col) do
        gensym()
    end
```

src/parsing_astable.jl
```diff
@@ -0,0 +1,95 @@
...
+    exprs = map(ex.args) do arg
+        if is_column_assigment(arg)
+            lhs = get_column_expr(arg.args[1])
+            rhs = arg.args[2]
+            new_ex = replace_syms_astable!(inputs_to_function, lhs_assignments, arg.args[2])
+            if haskey(inputs_to_function, lhs)
+                new_lhs = inputs_to_function[lhs]
+                lhs_assignments[lhs] = new_lhs
+            else
+                new_lhs = addkey!(lhs_assignments, lhs)
```
nalimilan, line 65: `addkey!` checks whether `lhs_assignments` contains `lhs`. If you already know that's not the case, better assign `gensym()` directly.

src/parsing.jl
```diff
@@ -269,7 +270,13 @@ function fun_to_vec(ex::Expr;
...
     end
 
-    if no_dest
+    if final_flags[ASTABLE_SYM][]
+        src, fun = get_source_fun_astable(ex; exprflags = final_flags)
+
+        return :($src => $fun => AsTable)
+    end
+
+    if no_dest # subet and with
```
nalimilan, line 279:
```suggestion
    if no_dest # subset and with
```

test/astable_flag.jl
```diff
@@ -0,0 +1,124 @@
...
+
+const ≅ = isequal
+
+@testset "@astable with just assignments" begin
+    df = DataFrame(a = 1, b = 2)
+
+    d = @rtransform df @astable begin
+        :x = 1
+
+        nothing
```
nalimilan, line 15: Why return `nothing`?

src/macros.jl
```diff
@@ -350,6 +350,99 @@ macro passmissing(args...)
     throw(ArgumentError("@passmissing only works inside DataFramesMeta macros."))
 end
 
+"""
+    astable(args...)
+
+Return a `NamedTuple` from a transformation inside DataFramesMeta.jl macros.
+
+`@astable` acts on a single block. It works through all top-level expressions
+and collects all such expressions of the form `:y = x`, i.e. assignments to a
```
nalimilan, line 359:
```suggestion
and collects all such expressions of the form `:y = ...`, i.e. assignments to a
```

src/macros.jl
```diff
@@ -350,6 +350,99 @@ macro passmissing(args...)
...
+df = DataFrame(a = 1)
+
+@rtransform df @astable begin
+    :x = 1
+    y = 50
+    :z = :x + y + :a
+end
+```
+
+becomes the pair
```
nalimilan, line 377:
```suggestion
become the pair
```

src/macros.jl
```diff
@@ -350,6 +350,99 @@ macro passmissing(args...)
...
+
+```
+@rtransform df begin
+    :new_col_1 = :x + :y
+    :new_col_2 = :new_col_1 + :z
+end
+```
+
+This because DataFrames.jl does not guarantee sequential evaluation of
+transformations. `@astable` solves this problem
```
nalimilan, line 402: While this is an interesting side-effect, the main goal of `AsTable` is to allow returning multiple columns from a single "function". Probably worth mentioning? For example it's useful with `extrema` to compute the minimum and the maximum at the same time.

src/macros.jl
```diff
@@ -350,6 +350,99 @@ macro passmissing(args...)
     throw(ArgumentError("@passmissing only works inside DataFramesMeta macros."))
 end
 
+"""
```
nalimilan, line 353: I guess you'll also add docs to the manual?

src/macros.jl
```diff
@@ -350,6 +350,99 @@ macro passmissing(args...)
...
+    astable(args...)
+
+Return a `NamedTuple` from a transformation inside DataFramesMeta.jl macros.
+
+`@astable` acts on a single block. It works through all top-level expressions
+and collects all such expressions of the form `:y = x`, i.e. assignments to a
+`Symbol`, which is a syntax error outside of the macro. At the end of the
```
nalimilan, line 360: Outside of DataFramesMeta macros, right?

nalimilan (commented):
Looks pretty cool! This will require a breaking release, right?

Using `@astable` to ensure operations are run sequentially is clever. The name is a bit surprising for this, but well... I also hope the compilation overhead isn't too large.

pdeffebach:
> Using @astable to ensure operations are run sequentially is clever. The name is a bit surprising for this, but well... I also hope the compilation overhead isn't too large.

This macro does more than that. It allows for local variables to per persistent as well. If I were to just force sequential transform calls, I would just create a `transform` call for each `:y = f(:x)` expression.

src/macros.jl
pdeffebach: great. changed.

src/parsing_astable.jl
pdeffebach: added.

test/astable_flag.jl
pdeffebach: I wanted to make sure it could handle an expression that wasn't just assignments.

src/parsing_astable.jl
pdeffebach: Ah this isn't quite right. Notice the two different dicts used, `lhs_assignments` and `inputs_to_function`.

pdeffebach:
This is ready for another round of reviews. I have added docs to the manual as well.

docs/src/index.md
```diff
@@ -396,11 +397,37 @@ julia> @rtransform df @passmissing x = parse(Int, :x_str)
...
 ```
 
+## Creating multiple columns at once with `@astable`
+
+Often new variables may depend on the same intermediate calculations. `@astable` makes it easy to create multiple
+new variables in the same operation, yet have them share
+information. 
+
+In a single block, all assignments of the form `:y = f(:x)` 
+or `$y = f(:x)` at the top-level are generate new columns.
```
bkamins, line 407:
```suggestion
or `$y = f(:x)` at the top-level generate new columns.
```

bkamins: maybe add what `$y` has to resolve to (I understand it has to be `Symbol`, or strings are also accepted?)

src/macros.jl
```diff
@@ -350,6 +350,120 @@ macro passmissing(args...)
     throw(ArgumentError("@passmissing only works inside DataFramesMeta macros."))
 end
 
+"""
+    astable(args...)
```
bkamins, line 354:
```suggestion
    @astable(args...)
```

src/macros.jl
```diff
@@ -350,6 +350,120 @@ macro passmissing(args...)
...
+
+```
+@transform df @astable begin
+    m = mean(:x)
+    :x_demeaned = :x .- m
+    :x2_demeaned = :x2 .- m
+end
+```
+
+The creation of `:x_demeaned` and `:x2_demeaned` both share the variable `m`,
```
bkamins, line 404: maybe add that the original data frame has to have `:x` and `:x2` columns (to make things clear).

src/macros.jl
```diff
@@ -350,6 +350,120 @@ macro passmissing(args...)
...
+transformations. `@astable` solves this problem
+
+@rtransform df @astable begin
+    :new_col_1 = :x + :y
+    :new_col_2 = :new_col_1 + :z
+end
+
+Column assignment in `@astable` follows the same rules as
+column assignment more generally. Construct a new column
+from a string by escaping it with `$DOLLAR`.
```
bkamins, line 427: can you add an example of this?

src/macros.jl
```diff
@@ -391,7 +392,7 @@ transform(df, [:a] => ByRow(f) => AsTable)
 `@astable` has two major advantages at the cost of increasing complexity.
 First, `@astable` makes it easy to create multiple columns from a single
 transformation, which share a scope. For example, `@astable` allows
-for the following
+for the following (where `:x` and `:x_2` exist in the `DataFrame` already).
```
bkamins, line 395:
```suggestion
for the following (where `:x` and `:x_2` exist in the data frame already).
```

docs/src/index.md
```diff
@@ -396,11 +397,38 @@ julia> @rtransform df @passmissing x = parse(Int, :x_str)
...
+   3 │     3    600      200       0
+```
+
+
 ## [Working with column names programmatically with `$`](@id dollar)
 
 DataFramesMeta provides the special syntax `$` for referring to 
-columns in a data frame via a `Symbol`, string, or column position as either
-a literal or a variable. 
+columns in a data frame via a `Symbol`, string, or column position as either a literal or a variable.
```
bkamins, line 431: While we are at it given our recent discussion on Discourse, I think it is essential to mention when the `$` reference is resolved.
Also maybe add an example when macros are used within a function? I think these are cases not trivial. This can be another PR of course

src/macros.jl
```diff
@@ -350,6 +350,138 @@ macro passmissing(args...)
...
+transformations. `@astable` solves this problem
+
+@rtransform df @astable begin
+    :new_col_1 = :x + :y
+    :new_col_2 = :new_col_1 + :z
+end
+
+Column assignment in `@astable` follows the same rules as
+column assignment more generally. Construct a new column
+from a string by escaping it with `$DOLLAR`, which can be a
```
bkamins, line 428: "from a string" here is unclear.

src/macros.jl
```diff
@@ -350,6 +350,138 @@ macro passmissing(args...)
...
 end
 
+"""
+    @astable(args...)
+
+Return a `NamedTuple` from a single transformation inside the DataFramesMeta.jl
+macros, `@select`, `@transform`, and their mutating and row-wise equivalents.
+
+`@astable` acts on a single block. It works through all top-level expressions
+and collects all such expressions of the form `:y = ...`, i.e. assignments to a
```
bkamins, line 360: also mention the `$` form on LHS.

src/macros.jl
```diff
@@ -350,6 +350,138 @@ macro passmissing(args...)
...
+2×3 DataFrame
+ Row │ a      min_b  max_b
+     │ Int64  Int64  Int64
+─────┼─────────────────────
+   1 │     1      5      6
+   2 │     2     70     80
+
+julia> @rtransform df @astable begin
+           f_a = first(:a)
+           $(DOLLAR)new_col = :a + :b + f_a
```
bkamins, line 467: where is `new_col` defined?

docs/src/index.md
pdeffebach: I will do this as another PR. In summary, you can't use other macros which use `$`. I will try and sort out if I can carve out an exception.

src/macros.jl
pdeffebach: added.

src/macros.jl
pdeffebach: I have re-phrased. Though I think people reading this are probably familiar with column assignment so I don't want to go into it in too much detail here.

src/macros.jl
pdeffebach: Sorry. Fixed.

pdeffebach:
Thanks, I've responded to the most recent round of reviews. 

I have added more tests. I can't think of any new tests to add at the moment, they seem pretty well covered.

pdeffebach:
cc @jkrumbiegel, if you want to review.

src/macros.jl
```diff
@@ -284,7 +284,7 @@ end
 
 
 """
-    passmissing(args...)
+    @passmissing(args...)
 
 Propograte missing values inside DataFramesMeta.jl macros.
```
nalimilan, line 289:
```suggestion
Propagate missing values inside DataFramesMeta.jl macros.
```

src/macros.jl
nalimilan: I still don't see it.

src/parsing_astable.jl
```diff
@@ -43,9 +50,15 @@ function is_column_assigment(ex::Expr)
...
 # Taken from MacroTools.jl
-# No docstring so assumed untable
+# No docstring so assumed unstable
 block(ex) = isexpr(ex, :block) ? ex : :($ex;)
 
+sym_or_str_to_sym(x::Union{AbstractString, Symbol}) = Symbol(x)
+function sym_or_str_to_sym(x)
+    e = "New columns created inside @astable must be Symbols or AbstractStrings"
+    throw(ArgumentError(e))
+end
```
nalimilan, line 60:
```suggestion
sym_or_str_to_sym(x) =
    throw(ArgumentError("New columns created inside @astable must be Symbols or AbstractStrings"))
```

test/astable_flag.jl
```diff
@@ -120,5 +120,50 @@ end
...
+
+    gd = groupby(df, :a)
+
+    d = @combine gd @astable begin
+        ex = extrema(:b)
+        :b_min = ex[1]
+        :b_max = ex[2]
+    end
+
+    @test sort(d.b_min) == [5, 7]
```
nalimilan, line 134: This kind of test is quite fragile. Better sort the whole data frame and compare it to the reference to make sure that groups and values match.

nalimilan:
> This macro does more than that. It allows for local variables to per persistent as well. If I were to just force sequential transform calls, I would just create a `transform` call for each `:y = f(:x)` expression.

Yes I know. What I'm saying is that mentioning sequential operations as the main justification for it was a bit weird.

pdeffebach:
I don't understand the changes to th egithub review interface, but I think i've addressed lingering issues. 

I just added some checks to make sure you can't do `@passmissing` and `@astable` at the same time. I think this is intuitive in the `@byrow` case, but I want to finish #276 , which makes `@passmissing` work on column-wise transformations, before I do this.

pdeffebach:
Tests pass and this can be merged.

docs/src/index.md
bkamins: To be clear why I stress it so much. With DataFrames.jl my answer to users is: if you learn Julia Base then you will know exactly how DataFrames.jl works. With DataFramesMeta.jl unfortunately this is not the case as it is a DSL so we need to be very precise how things work in documentation.

src/macros.jl
bkamins: I see it. It is fixed :)

test/astable_flag.jl
```diff
@@ -0,0 +1,176 @@
...
+        :x = [5]
+        y = 100
+        nothing
+    end
+
+    @test d == DataFrame(x = 5)
+end
+
+@testset "@astable with just assignments, mutating" begin
+    # After finalizing above testset
```
bkamins, line 46: @pdeffebach - this seems to be WIP?

test/astable_flag.jl
```diff
@@ -0,0 +1,178 @@
...
+using Statistics
+
+const ≅ = isequal
+
+@testset "@astable with just assignments" begin
+    df = DataFrame(a = 1, b = 2)
+
+    d = @rtransform df @astable begin
+        :x = 1
+
```
bkamins, line 14: newline is not needed. Also I am not clear why you add `nothing` here and below? Does it change anything?

test/astable_flag.jl
```diff
@@ -0,0 +1,178 @@
...
+        :x = 1
+        y = 5
+        :z = :x + y
+    end
+
+    @test d == DataFrame(x = 1, z = 6)
+
+    x_str = "x"
+    d = @rtransform df @astable begin
+        $x_str = 1
```
bkamins, line 107: this test is not good as you are assigning `1` to `:x` column, and it already was `1` so we do not see if it works correctly. Same below.

pdeffebach (commented):
Thanks for the round of reviews. I think I have addressed everything. Sorry about forgetting some of the tests.

bkamins (approved):
Looks good. Just please check why Documenter fails before merging. Thank you!

src/macros.jl
```diff
@@ -282,11 +282,10 @@ macro byrow(args...)
     throw(ArgumentError("@byrow is deprecated outside of DataFramesMeta macros."))
 end
 
-
 """
     @passmissing(args...)
 
-Propograte missing values inside DataFramesMeta.jl macros.
+Propagrate missing values inside DataFramesMeta.jl macros.
```
nalimilan, line 288:
```suggestion
Propagate missing values inside DataFramesMeta.jl macros.
```

src/macros.jl
```diff
@@ -350,18 +349,24 @@ macro passmissing(args...)
     throw(ArgumentError("@passmissing only works inside DataFramesMeta macros."))
 end
 
+global astable_docstring_snippet = """
```
nalimilan, line 352: Did you mean this?
```suggestion
const astable_docstring_snippet = """
```

src/macros.jl
```diff
@@ -350,18 +349,24 @@ macro passmissing(args...)
...
+    Transformations can also use the macro-flag `@astable` for creating multiple
```
nalimilan, line 353:
```suggestion
    Transformations can also use the macro-flag [`@astable`](@ref) for creating multiple
```

src/parsing.jl
```diff
@@ -224,6 +226,8 @@ function get_source_fun(function_expr; exprflags = deepcopy(DEFAULT_FLAGS))
         end
     end
 
+    println(MacroTools.prettify(fun))
+
```
nalimilan, line 230:
```suggestion
```

pdeffebach:
Thanks for the review. @nalimilan ready to merge?

nalimilan (approved):

bkamins:
Bravo!

PR #305, opened 2021-10-27
Suggestions for Introduction 

jd-foster:
I've tested the examples given then added corrections where I thought there to be inconsistencies, as well as some changes to help the flow of the explanation.


pdeffebach (approved):

pdeffebach:
Thanks!

PR #307, opened 2021-11-10
Allow AsTable on RHS in transformations

pdeffebach:
After the recent improvements to `AsTable` in DataFrames, I guess it's time to make it easier to work with this construct in DataFramesMeta.jl. 

No tests added yet.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -525,15 +525,15 @@ To reference columns with more complicated expressions, you must wrap column ref
 @transform df :a + $(get_column_name(x))
 ```
 
-### Operations with multiple columns at once using `AsTable` inside operations
+## Operations with multiple columns at once using `AsTable` inside operations
 
 In operations, it is also allowed to use `AsTable(cols)` to work with
 multiple columns at once, where the columns are grouped together in a
 `NamedTuple`. When `AsTable(cols)` appears in a operation, no
 other columns may be referenced in the block.
 
-`AsTable` on the right-hand-side also allows the use of the special
-column selectors `Not`, `Between`, and regular expressions. As well
+`AsTable` on the right-hand side also allows the use of the special
+column selectors `Not`, `Between`, and regular expressions, as well
 as working with lists of variables programmatically. 
 
 For example, consider a collection of column names `vars`, such that
@@ -596,7 +596,9 @@ inside the expression. The command
 
 will fail. 
 
-Finally, note that everyting inside `AsTable` is escaped by default. There is no ned to use `$` inside `AsTable` on the right-hand-side. For example
+Finally, note that everyting inside `AsTable` is escaped by default.
+There is no ned to use `$` inside `AsTable` on the right-hand side.
+For example
 
 ```
 :y = first(AsTable("a"))
@@ -609,15 +611,15 @@ will work as expected.
 
 At this point we have seen `AsTable` appear in three places:
 
-1. `AsTable` on the left-hand-side of transformations: `$AsTable = f(:a, :b)`
+1. `AsTable` on the left-hand side of transformations: `$AsTable = f(:a, :b)`
 2. The macro-flag `@astable` within the transformation. 
-3. `AsTable(cols)` on the right hand side for multi-column transformations. 
+3. `AsTable(cols)` on the right-hand side for multi-column transformations. 
 
 The differences between the three is summarized below
 
 | Operation         | Purpose                                                                             | Example           | Notes |
 |-------------------|-------------------------------------------------------------------------------------|-------------------|-------|
-| `$AsTable` on LHS | Create multiple columns at once, whose column names are only known programmatically | <pre lang=julia>$AsTable = f(:y)</pre> | Requires escaping with `$` until deprecation period ends for unquoted column names on LHS.      |
+| `$AsTable` on LHS | Create multiple columns at once, whose column names are only known programmatically | <pre lang="julia">$AsTable = f(:y)</pre> | Requires escaping with `$` until deprecation period ends for unquoted column names on LHS.      |
 | `@astable`        | Create multiple columns at once where number of columns is known in advance         | <pre lang="julia">@astable begin<br>    :y = :x - 1<br>    :z =  :y - 1<br>end</pre>
 | `AsTable` on RHS  | Work with multiple columns at once                                                  |<pre lang="julia">:y = sum(AsTable(Between("a", "z")))</pre> | Requires input columns, unlike on LHS |
 
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -427,104 +427,6 @@ julia> @transform df @astable begin
    3 │     3    600      200       0
 ```
 
-
-## [Working with column names programmatically with `$`](@id dollar)
-
-DataFramesMeta provides the special syntax `$` for referring to 
-columns in a data frame via a `Symbol`, string, or column position as either a literal or a variable. 
-
-```julia
-df = DataFrame(A = 1:3, :B = [2, 1, 2])
-
-nameA = :A
-df2 = @transform(df, :C = :B - $nameA)
-
-nameA_string = "A"
-df3 = @transform(df, :C = :B - $nameA_string)
-
-nameB = "B"
-df4 = @eachrow df begin 
-    :A = $nameB
-end
-```
-
-`$` can also be used to create new columns in a data frame. 
-
-```julia
-df = DataFrame(A = 1:3, B = [2, 1, 2])
-
-newcol = "C"
-@select(df, $newcol = :A + :B)
-
-@by(df, :B, $("A complicated" * " new name") = first(:A))
-
-nameC = "C"
-df3 = @eachrow df begin 
-    @newcol $nameC::Vector{Int}
-    $nameC = :A
-end
-```
-
-DataFramesMeta macros do not allow mixing of integer column references with references 
-of other types. This means `@transform(df, :y = :A + $2)`, attempting to add the columns 
-`df[!, :A]` and `df[!, 2]`, will fail. This is because in DataFrames, the command 
-
-```julia
-transform(df, [:A, 2] => (+) => :y)
-``` 
-
-will fail, as DataFrames requires the "source" column identifiers in a 
-`source => fun => dest` pair to all have the same type. DataFramesMeta adds one exception
-to this rule. `Symbol`s and strings are allowed to be mixed inside DataFramesMeta macros. 
-Consequently, 
-
-```
-@transform(df, :y = :A + $"B")
-```
-
-will not error even though 
-
-```
-transform(df, [:A, "B"] => (+) => :y)
-```
-
-will error in DataFrames. 
-
-For consistency, this restriction in the input column types also applies to `@with`
-and `@eachrow`. You cannot mix integer column references with `Symbol` or string column 
-references in `@with` and `@eachrow` in any part of the expression, but you can mix 
-`Symbol`s and strings. The following will fail:
-
-```julia
-df = DataFrame(A = 1:3, B = [2, 1, 2])
-@eachrow df begin 
-    :A = $2
-end
-
-@with df begin 
-    $1 + $"A"
-end
-```
-
-while the following will work without error
-
-```julia
-@eachrow df begin 
-    $1 + $2
-end
-
-@with df begin 
-    $1 + $2
-end
-```
-
-To reference columns with more complicated expressions, you must wrap column references in parentheses. 
-
-```
-@transform df :a + $("a column name" * " in two parts")
-@transform df :a + $(get_column_name(x))
-```
-
 ## Operations with multiple columns at once using `AsTable` inside operations
 
 In operations, it is also allowed to use `AsTable(cols)` to work with
@@ -653,6 +555,103 @@ The differences between the three is summarized below
 | `$AsTable` on LHS | Create multiple columns at once, whose column names are only known programmatically |  Requires escaping with `$` until deprecation period ends for unquoted column names on LHS. |
 | `@astable`        | Create multiple columns at once where number of columns is known in advance         | |
 | `AsTable` on RHS  | Work with multiple columns at once                                                  | Requires input columns, unlike on LHS |
+
+## [Working with column names programmatically with `$`](@id dollar)
+
+DataFramesMeta provides the special syntax `$` for referring to 
+columns in a data frame via a `Symbol`, string, or column position as either a literal or a variable. 
+
+```julia
+df = DataFrame(A = 1:3, :B = [2, 1, 2])
+
+nameA = :A
+df2 = @transform(df, :C = :B - $nameA)
+
+nameA_string = "A"
+df3 = @transform(df, :C = :B - $nameA_string)
+
+nameB = "B"
+df4 = @eachrow df begin 
+    :A = $nameB
+end
+```
+
+`$` can also be used to create new columns in a data frame. 
+
+```julia
+df = DataFrame(A = 1:3, B = [2, 1, 2])
+
+newcol = "C"
+@select(df, $newcol = :A + :B)
+
+@by(df, :B, $("A complicated" * " new name") = first(:A))
+
+nameC = "C"
+df3 = @eachrow df begin 
+    @newcol $nameC::Vector{Int}
+    $nameC = :A
+end
+```
+
+DataFramesMeta macros do not allow mixing of integer column references with references 
+of other types. This means `@transform(df, :y = :A + $2)`, attempting to add the columns 
+`df[!, :A]` and `df[!, 2]`, will fail. This is because in DataFrames, the command 
+
+```julia
+transform(df, [:A, 2] => (+) => :y)
+``` 
+
+will fail, as DataFrames requires the "source" column identifiers in a 
+`source => fun => dest` pair to all have the same type. DataFramesMeta adds one exception
+to this rule. `Symbol`s and strings are allowed to be mixed inside DataFramesMeta macros. 
+Consequently, 
+
+```
+@transform(df, :y = :A + $"B")
+```
+
+will not error even though 
+
+```
+transform(df, [:A, "B"] => (+) => :y)
+```
+
+will error in DataFrames. 
+
+For consistency, this restriction in the input column types also applies to `@with`
+and `@eachrow`. You cannot mix integer column references with `Symbol` or string column 
+references in `@with` and `@eachrow` in any part of the expression, but you can mix 
+`Symbol`s and strings. The following will fail:
+
+```julia
+df = DataFrame(A = 1:3, B = [2, 1, 2])
+@eachrow df begin 
+    :A = $2
+end
+
+@with df begin 
+    $1 + $"A"
+end
+```
+
+while the following will work without error
+
+```julia
+@eachrow df begin 
+    $1 + $2
+end
+
+@with df begin 
+    $1 + $2
+end
+```
+
+To reference columns with more complicated expressions, you must wrap column references in parentheses. 
+
+```
+@transform df :a + $("a column name" * " in two parts")
+@transform df :a + $(get_column_name(x))
+```
 
 ## Using `src => fun => dest` calls using `$`
 
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -638,7 +638,7 @@ For example
 will work as expected.  
 
 
-### AsTable and `@astable`, explained
+## AsTable and `@astable`, explained
 
 At this point we have seen `AsTable` appear in three places:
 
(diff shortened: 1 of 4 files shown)


bkamins:
What would be the test cases for this?

pdeffebach:
a row-wise sum

```
cols = [:a, :b, :c, :d]
@rtransform df :y = sum(AsTable(cols))
```

bkamins:
Nice - probably documentation should be updated (or does it already have it?)

pdeffebach:
No I haven't done any of that yet. I'll ping you when its ready.

pdeffebach:
Okay I've added enough stuff so that you can see the API

```

julia> df = DataFrame(rand(10, 10), :auto);

julia> @rselect df :sum_all = sum(AsTable(names(df)))
10×1 DataFrame
 Row │ sum_all 
     │ Float64 
─────┼─────────
   1 │ 4.19633
   2 │ 4.69652
   3 │ 2.96451
   4 │ 4.32973
   5 │ 4.69788
   6 │ 3.68512
   7 │ 6.47631
   8 │ 5.80387
   9 │ 4.92043
  10 │ 4.83637
```

pdeffebach:
Okay ready for a review @bkamins @nalimilan 

If only to check if the API makes sense.

src/macros.jl
```diff
@@ -630,6 +630,39 @@ macro with(d, body)
     esc(with_helper(d, body))
 end
 
+# TODO: New docstring for @subset
+astable_rhs_docs = """
+In transformations, it is also allowed to use `AsTable(cols)` to work with
+multiple columns at once, where the columns are grouped together in a
+`NamedTuple`. When `AsTable(cols)` appears in a transformation, no
+other columns may be referenced in the block.
```
bkamins, line 638: how are then symbols outside of `AsTable` treated - they error or are treated as symbols?

bkamins: Also can `AsTable` appear in the specification multiple times?

Also is something like `target = AsTable(nms)` valid? And what should/can `target` be then?

pdeffebach: Good questions. 

The error would be the same as `transform(df, [AsTable([:a, :b]), :c] => ...)`

```
julia> @select df :y = begin 
           nothing
           AsTable([:x1, :x2])
       end
ERROR: ArgumentError: Table returned but a single output column was expected
Stacktrace:
```

but the following works

```
julia> @select df $AsTable = begin 
           nothing
           AsTable([:x1, :x2])
       end
10×2 DataFrame
 Row │ x1         x2       
     │ Float64    Float64  
─────┼─────────────────────
   1 │ 0.383283   0.394874
   2 │ 0.173797   0.285121
   3 │ 0.0262305  0.472488
   4 │ 0.0129008  0.990346
   5 │ 0.368682   0.247991
   6 │ 0.0210636  0.296448
   7 │ 0.473996   0.130082
   8 │ 0.680537   0.159793
   9 │ 0.0541043  0.572126
  10 │ 0.93897    0.050982
```

I have to fix the parsing to avoid that `nothing` in there. Will do

pdeffebach: I suppose I should also special-case `AsTable` on the LHS. But I can do that later.

bkamins: OK - thank you!

src/macros.jl
```diff
@@ -630,6 +630,39 @@ macro with(d, body)
...
+
+Using `AsTable` in this way is useful for working with many columns
+at once programatically. For example, to take a row-wise sum of the
```
nalimilan, line 641:
```suggestion
at once programmatically. For example, to compute the row-wise sum of the
```

src/macros.jl
```diff
@@ -630,6 +630,39 @@ macro with(d, body)
...
+columns `[:a, :b, :c, :d]`, write
+
+```
+nms = [:a, :b, :c, :d]
+@byrow :c = sum(AsTable(nms))
```
nalimilan, line 646: Why not just
```suggestion
@byrow :c = sum(AsTable([:a, :b, :c, :d]))
```

src/macros.jl
```diff
@@ -630,6 +630,39 @@ macro with(d, body)
...
+```
+
+The constructs the pair
```
nalimilan, line 649:
```suggestion
This constructs the pairs
```

src/macros.jl
```diff
@@ -630,6 +630,39 @@ macro with(d, body)
     esc(with_helper(d, body))
 end
 
+# TODO: New docstring for @subset
```
nalimilan, line 633: What does this mean?

nalimilan (commented):
Just a few stylistic comments.

src/macros.jl
pdeffebach: The current docstring only talks about transformations, i.e. creating new columns. But `@subset` and `@orderby` don't do that, and thus need their own docstrings.

pdeffebach: Actually, @nalimilan I could use your advice on what to do here. 

There is a trade-off between consolidating documentation, having too-long of docstrings, and having all the information in the docstring. 

Should I write variants of this text for every macro? Or put this in the documentation `.md` file separately?

bkamins: My opinion:
* Docstring should be a complete contract for a given functionality.
* If some functionality is shared you can give a reference to a different docstring.
* Explanations of why and how should be given in a manual.

pdeffebach: Okay I've added a pretty lengthy discussion in the `index.md` file. 

@nalimilan @bkamins this is ready for another review.

docs/src/index.md
```diff
@@ -525,11 +525,111 @@ To reference columns with more complicated expressions, you must wrap column ref
...
+
+1. `AsTable` on the left-hand-side of transformations: `$AsTable = f(:a, :b)`
+2. The macro-flag `@astable` within the transformation. 
+3. `AsTable(cols)` on the right hand side for multi-column transformations. 
+
+The differences between the three is summarized below
+
+| Operation         | Purpose                                                                             | Example           | Notes |
+|-------------------|-------------------------------------------------------------------------------------|-------------------|-------|
+| `$AsTable` on LHS | Create multiple columns at once, whose column names are only known programmatically | <pre lang=julia>$AsTable = f(:y)</pre> | Requires escaping with `$` until deprecation period ends for unquoted column names on LHS.      |
```
pdeffebach, line 620: Pretty hacky way to make a reasonable little table. But it looks nice when compiled.

docs/src/index.md
```diff
@@ -525,11 +525,111 @@ To reference columns with more complicated expressions, you must wrap column ref
...
+
+```
+df = DataFrame(a = [11, 14], b = [17, 10], c = [12, 5]);
+vars = ["a", "b"];
+```
+
+To make a new column which is the sum of `vars`, write
+
+```
+julia> @rtransform df :y = sum(AsTable(vars))
```
bkamins, line 549: maybe also add an example when just a column selector is passed to `AsTable` as it will be maybe more common?

docs/src/index.md
```diff
@@ -525,11 +525,111 @@ To reference columns with more complicated expressions, you must wrap column ref
...
+| `@astable`        | Create multiple columns at once where number of columns is known in advance         | <pre lang="julia">@astable begin<br>    :y = :x - 1<br>    :z =  :y - 1<br>end</pre>
+| `AsTable` on RHS  | Work with multiple columns at once                                                  |<pre lang="julia">:y = sum(AsTable(Between("a", "z")))</pre> | Requires input columns, unlike on LHS |
+
+
+### Using `src => fun => dest` calls using `$`
+
+If an argument is entirely wrapped in `$()`, the result bypasses the anonymous function 
+creation of DataFramesMeta.jl and is passed to the underling DataFrames.jl function 
+directly. Importantly, this allows for `src => fun => dest` calls from the DataFrames.jl 
+"mini-language" directly.
```
bkamins, line 630: again - maybe add an example? something like `[:a, :b] .=> [sum mean]`? as I guess this is hard to express in DataFramesMeta.jl?

docs/src/index.md
```diff
@@ -525,11 +525,111 @@ To reference columns with more complicated expressions, you must wrap column ref
...
+```
+
+will work as expected.  
+
+
+### AsTable and `@astable`, explained
+
+At this point we have seen `AsTable` appear in three places:
+
+1. `AsTable` on the left-hand-side of transformations: `$AsTable = f(:a, :b)`
```
nalimilan, line 612:
```suggestion
1. `AsTable` on the left-hand side of transformations: `$AsTable = f(:a, :b)`
```

docs/src/index.md
pdeffebach: It does not work, unfortunately. I will think of another way to do this.

docs/src/index.md
pdeffebach: Added. And another example about how function can now "see" the names of columns
.

pdeffebach:
Thanks for the review! 

I have addressed all comments. I also got rid of the code example in the table because I couldn't figure out a way to get it to work. 

I think this is ready to merge?

docs/src/index.md
```diff
@@ -525,11 +525,165 @@ To reference columns with more complicated expressions, you must wrap column ref
...
 ```
 
-## Constructing multi-column arguments and `src => fun => dest` calls using `$`
+## Operations with multiple columns at once using `AsTable` inside operations
 
-If an argument is entirely wrapped in `$()`, the result bypasses the anonymous function creation of DataFramesMeta.jl and is passed to the underling DataFrames.jl function directly. Importantly, this allows for multi-column selection and passing `src => fun => dest` calls from the DataFrames.jl "mini-language" directly.
+In operations, it is also allowed to use `AsTable(cols)` to work with
+multiple columns at once, where the columns are grouped together in a
+`NamedTuple`. When `AsTable(cols)` appears in a operation, no
+other columns may be referenced in the block.
```
bkamins, line 533: and it can appear only once - right?

docs/src/index.md
```diff
@@ -525,11 +525,165 @@ To reference columns with more complicated expressions, you must wrap column ref
...
+For example, consider a collection of column names `vars`, such that
+
+```
+df = DataFrame(a = [11, 14], b = [17, 10], c = [12, 5]);
+vars = ["a", "b"];
+```
+
+To make a new column which is the sum of `vars`, write
+
+```
```
bkamins, line 548: this is not for this PR if it is not covered, but do you resolve it to `AsTable(vars) => sum => :y` (so that we use an efficient path in DataFrames.jl)?

docs/src/index.md
```diff
@@ -525,11 +525,165 @@ To reference columns with more complicated expressions, you must wrap column ref
...
-If an argument is entirely wrapped in `$()`, the result bypasses the anonymous function creation of DataFramesMeta.jl and is passed to the underling DataFrames.jl function directly. Importantly, this allows for multi-column selection and passing `src => fun => dest` calls from the DataFrames.jl "mini-language" directly.
+In operations, it is also allowed to use `AsTable(cols)` to work with
+multiple columns at once, where the columns are grouped together in a
+`NamedTuple`. When `AsTable(cols)` appears in a operation, no
+other columns may be referenced in the block.
 
-### Multi-argument selectors 
+`AsTable` on the right-hand side also allows the use of the special
+column selectors `Not`, `Between`, and regular expressions, as well
+as working with lists of variables programmatically.
```
bkamins, line 537: I understand that anything inside `AsTable` is untouched by the macro - right? Maybe add this comment.

docs/src/index.md
```diff
@@ -525,11 +525,165 @@ To reference columns with more complicated expressions, you must wrap column ref
...
+   1 │    11     17     12     28
+   2 │    14     10      5     24
+```
+
+`AsTable` on the right-hand side also allows operations which can use the names of the variables. 
+
+```
+julia> function fun_with_new_name(x::NamedTuple)
+           nms = string.(propertynames(x))
+           new_name = join(nms, "_") * "_sum"
```
bkamins, line 575:
```suggestion
           new_name = Symbol(join(nms, "_"), "_sum")
```

docs/src/index.md
```diff
@@ -525,11 +525,165 @@ To reference columns with more complicated expressions, you must wrap column ref
...
+           s = sum(x)
+           (; Symbol(new_name) => s)
```
bkamins, line 577:
```suggestion
           s = sum(x)
           (; new_name => s)
```

docs/src/index.md
bkamins: I guess the answer is yes given your comment below.

docs/src/index.md
bkamins: OK - I see you answer this question below.

docs/src/index.md
```diff
@@ -525,11 +525,165 @@ To reference columns with more complicated expressions, you must wrap column ref
...
+```
+
+will fail. 
+
+Finally, note that everyting inside `AsTable` is escaped by default.
+There is no ned to use `$` inside `AsTable` on the right-hand side.
+For example
+
+```
+:y = first(AsTable("a"))
```
bkamins, line 635: why is this an example of the behavior you describe? I guess you wanted `:a` not `"a"` - right?

docs/src/index.md
```diff
@@ -525,11 +525,165 @@ To reference columns with more complicated expressions, you must wrap column ref
...
+julia> df = DataFrame(a = [1, 2], b = [30, 40]);
+
+julia> @transform df $([:a, :b] .=> [sum mean])
+2×6 DataFrame
+ Row │ a      b      a_sum  b_sum  a_mean   b_mean  
+     │ Int64  Int64  Int64  Int64  Float64  Float64 
+─────┼──────────────────────────────────────────────
+   1 │     1     30      3     70      1.5     35.0
+   2 │     2     40      3     70      1.5     35.0
+
```
bkamins, line 683:
```suggestion
```

src/macros.jl
```diff
@@ -349,7 +349,7 @@ macro passmissing(args...)
     throw(ArgumentError("@passmissing only works inside DataFramesMeta macros."))
 end
 
-const astable_docstring_snippet = """
+const astable_macro_flag_docs = """
```
bkamins, line 352: maybe uppercase the constant name?

src/macros.jl
```diff
@@ -625,12 +625,102 @@ julia> @with df @byrow :x * :y
     global scope, then a variable cannot be assigned without using the `global` keyword.
     If the parent scope is a local scope (inside a function or let block for example),
     the `global` keyword is not needed to assign to that parent scope.
+
+!!! note
```
bkamins, line 629: should this be indented by 4 spaces?

src/macros.jl
```diff
@@ -625,12 +625,102 @@ julia> @with df @byrow :x * :y
...
+    Using `AsTable` inside `@with` block is currently not supported.
 """
 macro with(d, body)
     esc(with_helper(d, body))
 end
 
+astable_rhs_orderby_docs = """
```
bkamins, line 636: make it a `const` and uppercase the name?

src/macros.jl
```diff
@@ -625,12 +625,102 @@ julia> @with df @byrow :x * :y
...
+sum of the columns `:a`, `:b`, and `:c`, write
+
+```
+@byrow sum(AsTable([:a, :b, :c]))
+```
+
+This constructs the pair
+
+```
+AsTable(AsTable([:a, :b, :c])) => ByRow(sum)
```
bkamins, line 653: probably a single `AsTable` should be here?

src/macros.jl
```diff
@@ -625,12 +625,102 @@ julia> @with df @byrow :x * :y
...
+`AsTable` on the right-hand side also allows the use of the special
+column selectors `Not`, `Between`, and regular expressions. For example,
+to order all rows by the product of all columns starting with `"a"`, write
+
+```
+@byrow prod(AsTable(r"^a"))
+```
+"""
+
+astable_rhs_subset_docs = """
```
bkamins, line 665: make it a `const` and uppercase.

src/macros.jl
```diff
@@ -625,12 +625,102 @@ julia> @with df @byrow :x * :y
...
+sum of the columns `:a`, `:b`, and `:c` is greater than `5`, write
+
+```
+@byrow sum(AsTable([:a, :b, :c])) > 5
+```
+
+This constructs the pair
+
+```
+AsTable([:a, :b, :c]) => ByRow(t -> sum(t) > 5)
```
bkamins, line 682: out of curiosity - and for the future - we  could think what syntax could be used in such cases to create:
```
AsTable([:a, :b, :c]) => ByRow((t -> sum(t) > 5)∘collect)
```

src/macros.jl
```diff
@@ -625,12 +625,102 @@ julia> @with df @byrow :x * :y
...
+column selectors `Not`, `Between`, and regular expressions. For example,
+to subset all rows where the product of all columns starting with `"a"`,
+is greater than `5`, write
 
+```
+@byrow prod(AsTable(r"^a")) > 5
+```
+"""
+
+astable_rhs_select_transform_docs = """
```
bkamins, line 695: make it a `const` and uppercase

bkamins:
Looks good. I made some cosmetic comments.

docs/src/index.md
```diff
@@ -627,7 +660,28 @@ The differences between the three is summarized below
 If an argument is entirely wrapped in `$()`, the result bypasses the anonymous function 
 creation of DataFramesMeta.jl and is passed to the underling DataFrames.jl function 
 directly. Importantly, this allows for `src => fun => dest` calls from the DataFrames.jl 
-"mini-language" directly.
+"mini-language" directly. One example where this is useful is calling multiple functions across multiple input parameters. For insteance, the `Pair`
```
nalimilan, line 663:
```suggestion
"mini-language" directly. One example where this is useful is calling multiple functions across multiple input parameters. For instance, the `Pair`
```

docs/src/index.md
```diff
@@ -627,7 +660,28 @@ The differences between the three is summarized below
...
+
+```
+[:a, :b] .=> [sum mean]
+```
+
+takes the `sum` and `mean` of both columns `:a` and `:b` separately. It is not possible to express this with DataFrames.jl. But the operation can easily be performed with `$`
```
nalimilan, line 669: Is this what you meant?
```suggestion
takes the `sum` and `mean` of both columns `:a` and `:b` separately. It is not possible to express this with DataFramesMeta.jl. But the operation can easily be performed with `$`
```

docs/src/index.md
```diff
@@ -525,11 +525,165 @@ To reference columns with more complicated expressions, you must wrap column ref
 @transform df :a + $(get_column_name(x))
 ```
 
-## Constructing multi-column arguments and `src => fun => dest` calls using `$`
+## Operations with multiple columns at once using `AsTable` inside operations
```
nalimilan, line 528: Again, why not put this section right after the one about `@astable`?

src/macros.jl
pdeffebach: added.

src/macros.jl
pdeffebach: I doubt it. No way of knowing how the function uses the named tuple and I don't want to special-case the `sum` as a syntax.

src/macros.jl
pdeffebach: added.

docs/src/index.md
pdeffebach: No. it can appear multiple times. But it has to look the exact same, you can't have, for example

```
@transform df :y = sum(AsTable([:a, :b])) + sum(AsTable(["a", "b"]))
```

There are similar restrictions elsewhere, for example with `@astable`.

docs/src/index.md
pdeffebach: No this is correct. Usually when working with `"a"`, you would need to write `$"a"`. I guess I could make it a variable you have to interpolate.

src/macros.jl
pdeffebach: Nope it works

pdeffebach: Though I may need you to write the implementation for this in the future since I will have to rely on DataFrames internals to really get this working.

docs/src/index.md
pdeffebach: Thanks. Re-arranged.

pdeffebach (commented):
Thanks for the review. I addressed all comments.

pdeffebach (commented):
Okay ready for review / merging.

src/macros.jl
bkamins: So why do you indent `!!! note` above (line 621) with 4 spaces and not here?

bkamins:
I went through the comments and resolved all that I thought were addressed. I left one related to indentation (a minor issue) + one regarding the order of presentation of things (so that @nalimilan can check if now we have what he wants).

Thank you for working on thins!

src/macros.jl
pdeffebach: I see for spaces. Can you add a change via github? (We can also worry about this in another PR)

src/macros.jl
```diff
@@ -625,12 +625,102 @@ julia> @with df @byrow :x * :y
     global scope, then a variable cannot be assigned without using the `global` keyword.
     If the parent scope is a local scope (inside a function or let block for example),
     the `global` keyword is not needed to assign to that parent scope.
+
+!!! note
+    Using `AsTable` inside `@with` block is currently not supported.
```
bkamins, line 630:
```suggestion
    !!! note
        Using `AsTable` inside `@with` block is currently not supported.
```

src/macros.jl
bkamins: ah - you are right. I have mistakenly mixed this with line 352.

bkamins (approved):

nalimilan (approved):

pdeffebach:
Awesome! Thanks for all the hard work reviewing this.

PR #315, opened 2021-12-09
Small docstring fix in src/macros.jl

storopoli:
I was seeing the help for `@subset` in a Julia REPL and noticed that there is a missing space in the docstring.


bkamins:
Thank you!

PR #317, opened 2021-12-27
Add support for f ∘ g 

pdeffebach:
This PR does *not* transform `f(g(:x))` to `(f ∘ g)(:x)`. 

Rather, it ensures that the expression

```
:y = (f ∘ g)(:a, :b)
```

gets turned into 

```
[:a, :b] => (f ∘ g ) => :y
```

which is a good first step for transforming `f(g(x))`. 

I'm leaving this undocumented for now, since it's a behind-the-scenes transformation.

### DIFF
--- a/src/parsing.jl
+++ b/src/parsing.jl
@@ -59,11 +59,19 @@ function replace_dotted!(membernames, e)
     Expr(:., x_new, y_new)
 end
 
+composed_or_symbol(x) = false
+composed_or_symbol(x::Symbol) = true
+function composed_or_symbol(x::Expr)
+    x.head == :call &&
+        x.args[1] == :∘ &&
+        all(composed_or_symbol, x.args[2:end])
+end
+
 is_simple_non_broadcast_call(x) = false
 function is_simple_non_broadcast_call(expr::Expr)
     expr.head == :call &&
         length(expr.args) >= 2 &&
-        expr.args[1] isa Symbol &&
+        composed_or_symbol(expr.args[1]) &&
         all(a -> get_column_expr(a) !== nothing, expr.args[2:end])
 end
 
@@ -71,7 +79,7 @@ is_simple_broadcast_call(x) = false
 function is_simple_broadcast_call(expr::Expr)
     expr.head == :. &&
         length(expr.args) == 2 &&
-        expr.args[1] isa Symbol &&
+        composed_or_symbol(expr.args[1]) &&
         expr.args[2] isa Expr &&
         expr.args[2].head == :tuple &&
         all(a -> get_column_expr(a) !== nothing, expr.args[2].args)
--- a/test/function_compilation.jl
+++ b/test/function_compilation.jl
@@ -215,7 +215,7 @@ end
             (slowtime[2] > fasttime[2]) || @warn("Slow compilation")
 
             fasttime = @timed @rselect df_wide :y = (sum ∘ skipmissing)(AsTable(:))
-            slowtime = @timed select(df_wide, AsTable(:) => ByRow(t -> sum(t)) => :y)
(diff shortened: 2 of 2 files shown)


pdeffebach:
@bkamins can I get an approval? I have code ready based on this which will do the more substantive change of transforming `f(g(:x, :y))` into `(f ∘ g)(:x, :y)`.

test/function_compilation.jl
```diff
@@ -4,16 +4,18 @@ using Test
 using DataFramesMeta
 
 @testset "function_compilation" begin
+    @eval begin
```
bkamins, line 7: what is the benefit of evaluating it on module level?

test/function_compilation.jl
```diff
@@ -187,4 +189,38 @@ using DataFramesMeta
         end
     end
 end
+
+@testset "composed compilation" begin
+    @eval begin
```
bkamins, line 194: same here - I do not understand the benefit of `@eval`?

test/function_compilation.jl
```diff
@@ -187,4 +189,38 @@ using DataFramesMeta
...
+
+        f(x) = identity(x)
+        g(x, y) = x + y
+
+        df_wide = DataFrame(rand(10, 1000), :auto)
+    end
+
+    for _ in 1:2
+        @eval begin
+            @test @select(df, :y = (f ∘ g)(:a, :b)).y == [3]
```
bkamins, line 205: Do you think testing timing is enough? Maybe also generated code should be tested?

src/parsing.jl
```diff
@@ -59,19 +59,27 @@ function replace_dotted!(membernames, e)
     Expr(:., x_new, y_new)
 end
 
+composed_or_symbol(x) = false
+composed_or_symbol(x::Symbol) = true
+function composed_or_symbol(x::Expr)
+    x.head == :call &&
+        x.args[1] == :∘ &&
+        all(composed_or_symbol, x.args[2:end])
```
bkamins, line 67: `[2:end]` selector is not needed as `:∘` is a symbol. Bit can stay as it is also.

test/function_compilation.jl
pdeffebach: The stuff inside the `for` loop needs to be `@eval`ed to prevent Julia from caching the anonymous functions inside the `for` loop. 

The `for` loop needs to happen to test that we compare the compiled times for the macros which should be `:x => f => :y` where `f` is the same and thus added to the method table. 

This logic is complicated and realistically I think we should only be checking the timings on the 2nd run, so we should add `if i == 2` somewhere. But it's working now and I am a bit scared to mess with it in this PR.

test/function_compilation.jl
pdeffebach: As mentioned above, `@eval` is necessary because we need to evaluate in global scope to prevent caching the anonymous function in the method table.

test/function_compilation.jl
bkamins: OK

pdeffebach:
No that's a good catch. Overall I'm not very good with macro hygiene in this package and should fix it.

bkamins:
I am not sure who is good at macro hygene 😄 (e.g. I do not write macros often and in consequence it is really hard for me to do it properly when I do).

test/function_compilation.jl
pdeffebach: We don't check generated code anywhere else. I think this should be enough.

pdeffebach:
Macro hygiene doesn't matter here, actually. Since the expression `f ∘ g` is just "moved" not changed in any way. And it is assumed to be callable both before and after the transformation.

test/function_compilation.jl
pdeffebach: Nevermind it's easy to add. Added. Will merge after tests.

PR #318, opened 2021-12-29
Updated minor typos in index.md

svilupp:
I've fixed a few minor typos, see the tracked file.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -337,8 +337,8 @@ example the result of `@transform(df, @byrow :y = f(:x))` and
 
 ## Propagating missing values with `@passmissing`
 
-Many Julia functions to not automatically propagate missing values. For instance, 
-`parse(Int, missing)` will error. 
+Many Julia functions do not automatically propagate missing values. For instance, 
+`parse(Int, missing)` will throw an error. 
 
 Missings.jl provides the `passmissing` function-wrapper to help get around these
 roadblocks: `passmissing(f)(args...)` will return `missing` if any of `args` is
@@ -610,13 +610,13 @@ Consequently,
 @transform(df, :y = :A + $"B")
 ```
 
-will not error even though 
+will not throw an error even though 
 
 ```
 transform(df, [:A, "B"] => (+) => :y)
 ```
 
-will error in DataFrames. 
+will throw an error in DataFrames. 
 
 For consistency, this restriction in the input column types also applies to `@with`
 and `@eachrow`. You cannot mix integer column references with `Symbol` or string column 
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -610,13 +610,13 @@ Consequently,
 @transform(df, :y = :A + $"B")
 ```
 
-will not throw an error even though 
+will not error even though 
 
 ```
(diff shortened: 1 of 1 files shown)


docs/src/index.md
```diff
@@ -337,8 +337,8 @@ example the result of `@transform(df, @byrow :y = f(:x))` and
 
 ## Propagating missing values with `@passmissing`
 
-Many Julia functions to not automatically propagate missing values. For instance, 
-`parse(Int, missing)` will error. 
+Many Julia functions do not automatically propagate missing values. For instance, 
+`parse(Int, missing)` will throw an error.
```
pdeffebach, line 341: I think "will error" is fine. 
```suggestion
`parse(Int, missing)` will error. 
```

docs/src/index.md
```diff
@@ -610,13 +610,13 @@ Consequently,
 @transform(df, :y = :A + $"B")
 ```
 
-will not error even though 
+will not throw an error even though
```
pdeffebach, line 613:
```suggestion
will not error even though 
```

docs/src/index.md
```diff
@@ -610,13 +610,13 @@ Consequently,
...
 
-will not error even though 
+will not throw an error even though 
 
 ```
 transform(df, [:A, "B"] => (+) => :y)
 ```
 
-will error in DataFrames. 
+will throw an error in DataFrames.
```
pdeffebach, line 619:
```suggestion
will error in DataFrames. 
```

docs/src/index.md
svilupp: I can leave it as is and close the PR, but I found the phrase distracting.

English is my second language, but I believe "error" is a noun ([MW on "error"](https://www.merriam-webster.com/dictionary/error#synonyms)) and the verb would be "to err" ([MW on "err"](https://www.merriam-webster.com/dictionary/err). 

I've also searched Google for the phrases "to error" and "will error" and haven't found any reputable reference.

bkamins:
I am not a native, but in computing "error" is often used as a verb AFAICT.

svilupp:
Noted. In which case, I'll remove "error as a verb" fixes. It was distracting me during the read, but it seems to be only my problem.

bkamins:
I would rely on the judgement of @pdeffebach who is a native speaker here. Though, maybe your proposals are better for non-natives? @pdeffebach - what do you think?

docs/src/index.md
svilupp: Updated as per @pdeffebach suggestion.

docs/src/index.md
svilupp: Updated as per @pdeffebach suggestion.

docs/src/index.md
svilupp: Updated as per @pdeffebach suggestion.

svilupp:
Already followed @pdeffebach 's suggestion - left only the fix from "to" --> "do", which is likely to have been a typo.

Apologies for wasting your time (and huge thank you for the amazing packages!!)

pdeffebach:
Thanks! Yes `error` as a verb is common, though I'm sure Milan actually wrote this.

PR #319, opened 2022-01-04
Composition of nested functions

pdeffebach:
This PR performs the following transformation

```
julia> DataFramesMeta.@col :y = first(last(:x))
[:x] => (first ∘ last => :y)
```

Tests are added and pass. I only need to fix one issue with macro hygiene, in case someone overloads `∘`.

### DIFF
--- a/src/parsing.jl
+++ b/src/parsing.jl
@@ -97,18 +97,16 @@ make_composed(x) = x
 function make_composed(x::Expr)
     funs = Any[]
     x_orig = x
+    nested_once = false
     while true
         if is_nested_fun(x)
             push!(funs, x.args[1])
             x = x.args[2]
-        elseif is_simple_non_broadcast_call(x)
-            if !isempty(funs)
+            nested_once = true
+        elseif is_simple_non_broadcast_call(x) && nested_once
                 push!(funs, x.args[1])
                 # ∘(f, g, h)(:x, :y, :z)
                 return Expr(:call, Expr(:call, ∘, funs...), x.args[2:end]...)
-            else
-                throw(Argumenterror("Not eligible for function composition"))
-            end
         else
             throw(Argumenterror("Not eligible for function composition"))
         end
--- a/src/parsing.jl
+++ b/src/parsing.jl
@@ -68,15 +68,15 @@ function composed_or_symbol(x::Expr)
 end
 
 is_call(x) = false
-is_call(x::Expr) = x.head == :call
+is_call(x::Expr) = x.head === :call
 
 is_nested_fun(x) = false
 function is_nested_fun(x::Expr)
-    x.head == :call &&
+    x.head === :call &&
     length(x.args) == 2 &&
     is_call(x.args[2]) &&
     # AsTable(:x) or `$(:x)`
-    get_column_expr(x.args[2]) == nothing
+    return get_column_expr(x.args[2]) === nothing
 end
 
 is_nested_fun_recursive(x, nested_once) = false
@@ -84,11 +84,7 @@ function is_nested_fun_recursive(x::Expr, nested_once)
     if is_nested_fun(x)
         return is_nested_fun_recursive(x.args[2], true)
     elseif is_simple_non_broadcast_call(x)
-        if nested_once
-            return true
-        else
-            return false
-        end
+        return nested_once
     else
         return false
     end
@@ -104,11 +100,11 @@ function make_composed(x::Expr)
             x = x.args[2]
             nested_once = true
         elseif is_simple_non_broadcast_call(x) && nested_once
-                push!(funs, x.args[1])
-                # ∘(f, g, h)(:x, :y, :z)
-                return Expr(:call, Expr(:call, ∘, funs...), x.args[2:end]...)
+            push!(funs, x.args[1])
+            # ∘(f, g, h)(:x, :y, :z)
+            return Expr(:call, Expr(:call, ∘, funs...), x.args[2:end]...)
         else
-            throw(Argumenterror("Not eligible for function composition"))
+            throw(ArgumentError("Not eligible for function composition"))
         end
     end
 end
--- a/src/parsing.jl
+++ b/src/parsing.jl
@@ -67,6 +67,39 @@ function composed_or_symbol(x::Expr)
         all(composed_or_symbol, x.args[2:end])
 end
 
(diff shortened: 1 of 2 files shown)


src/parsing.jl
```diff
@@ -67,6 +67,55 @@ function composed_or_symbol(x::Expr)
...
+function is_nested_fun(x::Expr)
+    x.head == :call &&
+    length(x.args) == 2 &&
+    is_call(x.args[2]) &&
+    # AsTable(:x) or `$(:x)`
+    get_column_expr(x.args[2]) == nothing
+end
+
+is_nested_fun_recursive(x, nested_once) = false
+function is_nested_fun_recursive(x::Expr, nested_once)
```
pdeffebach, line 83: Here and below are two recursive functions. `is_nested_fun_recursive` simply checks whether or not an expression is of the form `f(g(:x, :y))` and `make_composed` actually creates `(f ∘g)` . It may be too costly to do this recursive search twice but both should return relatively quickly.

pdeffebach:
I have fixed the macro hygiene issue with help from Slack. This is ready for a review.

src/parsing.jl
bkamins: I think performance should not be crucial here.

src/parsing.jl
```diff
@@ -67,6 +67,54 @@ function composed_or_symbol(x::Expr)
...
+is_call(x) = false
+is_call(x::Expr) = x.head == :call
+
+is_nested_fun(x) = false
+function is_nested_fun(x::Expr)
+    x.head == :call &&
+    length(x.args) == 2 &&
+    is_call(x.args[2]) &&
+    # AsTable(:x) or `$(:x)`
+    get_column_expr(x.args[2]) == nothing
```
bkamins, line 79: I understand you catch here cases where user passes things like `Cols(...)` or similar?

test/function_compilation.jl
```diff
@@ -218,11 +218,26 @@ end
...
 
             (slowtime[2] > fasttime[2]) || @warn("Slow compilation")
+
+            @test @select(df, :y = f(g(:a, :b))).y == [3]
+
+            fasttime = @timed @select(df, :y = f(g(:a, :b)))
+            slowtime = @timed select(df, [:a, :b] => ((a, b) -> f(g(a, b))) => :y )
+            (slowtime[2] > fasttime[2]) || @warn("Slow compilation")
+
+            fasttime = @timed @rselect df_wide :y = sum(skipmissing(AsTable(:)))
```
bkamins, line 228: I would add also correctness tests (not only timing).
Also I think it would be good to add more tests of different column selectors.

src/parsing.jl
pdeffebach: Actually `Cols` isn't recognized. It's just `AsTable` and `$(...)`. `AsTable` is the only way to do multiple column selection at the moment.

test/function_compilation.jl
pdeffebach: Add more correctness test and with different column selectors.

pdeffebach:
This should be ready for a review from @nalimilan

src/parsing.jl
```diff
@@ -67,6 +67,52 @@ function composed_or_symbol(x::Expr)
         all(composed_or_symbol, x.args[2:end])
 end
 
+is_call(x) = false
+is_call(x::Expr) = x.head == :call
```
nalimilan, line 71:
```suggestion
is_call(x::Expr) = x.head === :call
```

src/parsing.jl
```diff
@@ -67,6 +67,52 @@ function composed_or_symbol(x::Expr)
...
+
+is_nested_fun(x) = false
+function is_nested_fun(x::Expr)
+    x.head == :call &&
```
nalimilan, line 75:
```suggestion
    x.head === :call &&
```

src/parsing.jl
```diff
@@ -67,6 +67,52 @@ function composed_or_symbol(x::Expr)
...
+is_nested_fun_recursive(x, nested_once) = false
+function is_nested_fun_recursive(x::Expr, nested_once)
+    if is_nested_fun(x)
+        return is_nested_fun_recursive(x.args[2], true)
+    elseif is_simple_non_broadcast_call(x)
+        if nested_once
+            return true
+        else
+            return false
+        end
```
nalimilan, line 91:
```suggestion
        return nested_once
```

src/parsing.jl
```diff
@@ -67,6 +67,52 @@ function composed_or_symbol(x::Expr)
...
+is_call(x) = false
+is_call(x::Expr) = x.head == :call
+
+is_nested_fun(x) = false
+function is_nested_fun(x::Expr)
+    x.head == :call &&
+    length(x.args) == 2 &&
+    is_call(x.args[2]) &&
+    # AsTable(:x) or `$(:x)`
+    get_column_expr(x.args[2]) == nothing
```
nalimilan, line 79:
```suggestion
    return get_column_expr(x.args[2]) === nothing
```

src/parsing.jl
```diff
@@ -67,6 +67,52 @@ function composed_or_symbol(x::Expr)
...
+    nested_once = false
+    while true
+        if is_nested_fun(x)
+            push!(funs, x.args[1])
+            x = x.args[2]
+            nested_once = true
+        elseif is_simple_non_broadcast_call(x) && nested_once
+                push!(funs, x.args[1])
+                # ∘(f, g, h)(:x, :y, :z)
+                return Expr(:call, Expr(:call, ∘, funs...), x.args[2:end]...)
```
nalimilan, line 109:
```suggestion
            push!(funs, x.args[1])
            # ∘(f, g, h)(:x, :y, :z)
            return Expr(:call, Expr(:call, ∘, funs...), x.args[2:end]...)
```

src/parsing.jl
```diff
@@ -67,6 +67,52 @@ function composed_or_symbol(x::Expr)
...
+        else
+            throw(Argumenterror("Not eligible for function composition"))
```
nalimilan, line 111:
```suggestion
            throw(ArgumentError("Not eligible for function composition"))
```

nalimilan (commented):
Cool. Just a few minor things.

pdeffebach:
Thanks! Will merge.

Issue #321, opened 2022-01-11
More examples in docstrings


mahiki:
It would be very helpful if each macro/command had an example in the docstring.

For example, I've been having a lot of trouble using `@rtranform` to make a new column based on conditional aspects of other columns at the row level. The doctring in the REPL is:

```jl
help?> @rtransform
  @rtransform(x, args...)

  Row-wise version of @transform, i.e. all operations use @byrow by default. See @transform for details.
```

I didn't find the help I needed here, a good example would keep me moving along with my work.

I'm not a computer scientist, and its time-intensive to unravel how these excellent tools are implemented. I've invested time to apply julia in my work vs. the python ecosystem because of its great qualities, however the most consistent hurdle for me is the lack of examples.

Perhaps the usage is obvious to package developers, but for more pedestrian types like me nothing could be more illuminating than a good example.

bkamins:
We should improve docstrings, but in the mean time maybe this https://bkamins.github.io/julialang/2021/11/19/dfm.html would help you?

mahiki:
Ah yes, these are good, thank you @bkamins. To be fair the referred `@transform` and `@byrow` are helpful, it just requires more synthesis at an inconvenient time.

I guess the docstring for `@rtransform` could be like the following, should I submit a PR?

```jl
"""
    @rtransform(x, args...)

Row-wise version of @transform, i.e. all operations use @byrow by default. See @transform for details.

### Example
```jldoctest
julia> @rtransform(df, :a = 2 * :x, :b = :x * :y ^ 2)
df = DataFrame(x=1:5, y=11:15)
5×2 DataFrame
 Row │ x      y
     │ Int64  Int64
─────┼──────────────
   1 │     1     11
   2 │     2     12
   3 │     3     13
   4 │     4     14
   5 │     5     15

julia> @rtransform(df, :a = 2 * :x, :b = :x * :y ^ 2)
5×4 DataFrame
 Row │ x      y      a      b
     │ Int64  Int64  Int64  Int64
─────┼────────────────────────────
   1 │     1     11      2    121
   2 │     2     12      4    288
   3 │     3     13      6    507
   4 │     4     14      8    784
   5 │     5     15     10   1125
```
"""
```

mahiki:
I should definitely submit a PR. This is something I run into all the time, it would be fantastic to be in the habit of making contributions that people in the data engineering DE community would enjoy.

bkamins:
Looks good. Thank you!

[...]

PR #322, opened 2022-01-19
add ready examples to docstring for rselect, rtransform (#321)

mahiki:
I added REPL jldoctest examples for the `@rselect` and `@rtransform` macros to demonstrate the easy syntax for operations that need a ByRow treatment.  

I choose these examples to show an easy way to post a value to a new column based on the row value in another column or combination of columns.

### DIFF
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -1408,6 +1408,8 @@ See [`@transform`](@ref) for details.
 
 ### Examples
 ```jldoctest
+julia> using DataFramesMeta
+
 julia> df = DataFrame(x = 1:5, y = 11:15)
 5×2 DataFrame
  Row │ x      y
@@ -1687,6 +1689,8 @@ default. See [`@select`](@ref) for details.
 
 ### Examples
 ```jldoctest
+julia> using DataFramesMeta
+
 julia> df = DataFrame(x = 1:5, y = 10:14)
 5×2 DataFrame
  Row │ x      y
--- a/src/macros.jl
+++ b/src/macros.jl
@@ -927,9 +927,9 @@ julia> @rsubset df :A > 3
 
 julia> @rsubset df :A > 3 || :B == "pear"
 3×2 DataFrame
-Row │ A      B
-    │ Int64  String
-─────┼───────────────
+  Row │ A      B
+      │ Int64  String
+ ─────┼───────────────
     1 │     2  pear
     2 │     4  orange
     3 │     5  pear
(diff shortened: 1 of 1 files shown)


mahiki:
There was a [doctest failure in my macro examples for @rtransform](https://github.com/JuliaData/DataFramesMeta.jl/runs/4864357722?check_suite_focus=true#step:4:195), and [also @rselect](https://github.com/JuliaData/DataFramesMeta.jl/runs/4864357722?check_suite_focus=true#step:4:259). I'm not sure how this works so not able to determine what is missing.

The dataframe operations are correct in project REPL, I must be missing some syntax in the docsstring?

bkamins:
Probably `julia> using DataFramesMeta;` is needed (normally it should be set up to be globally loaded for all doctest strings, but I do not know if @pdeffebach did this)

mahiki:
Now I see. The other `jldoctest` sections are like this:

```
```jldoctest
julia> using DataFramesMeta      # <--- I didnt do this

julia> @transform df begin
           :a = 2 * :A
           :x = :A .+ :B
# ..etc
```

@pdeffebach I will push another commit and try to update the PR or file a new one. Alternately, I could make @bkamins suggested change to `make.jl`:

I think @bkamins  is saying that the following added to the `make.jl` file will alleviate the need to place `using DataFramesMeta for each doctest:

```julia
DocMeta.setdocmeta!(DataFramesMeta, :DocTestSetup, :(using DataFramesMeta); recursive=true)
```

Following the [DataFrames.jl/doc/make.jl file](https://github.com/JuliaData/DataFrames.jl/blob/main/docs/make.jl).

bkamins:
Yes, this is how we do it in https://github.com/JuliaData/DataFrames.jl/blob/5233159d470c8e5a9a2e2f93cd3ea67a7fff1769/docs/make.jl#L5.

pdeffebach:
Yes we should definitely fix that. But just do the `using DataFramesMeta` line for now and we will simplify the docstrings in a later PR.

pdeffebach:
Thanks a ton for this!

Do you think you could add these for `@rorderby` and `@rsubset` as well?

mahiki:
Added jldoctest examples for `@rorderby`, `@rsubset`. I was very glad to learn what the purpose of these macros is, its very convenient to write like this.

mahiki:
added minor change to df output.

mahiki:
Let me know if there are needed changes before running the workflow tests.

pdeffebach:
Sorry about the delay on this! Will merge tomorrow

pdeffebach (approved):

mahiki:
Awesome, ping me if you need some other help!

PR #323, opened 2022-01-20
Allow for keyword arguments

pdeffebach:
This just adds keyword arguments for calls of the form `@subset(df, :a; view = true)`, rather than the multi-line version

```
@subset df begin 
    :a 
end
```

I have added this for `@subset` but not others yet.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -403,7 +403,28 @@ julia> @rtransform df @passmissing :x = parse(Int, :x_str)
 ## Passing keyword arguments to underlying DataFrames.jl functions
 
 All DataFramesMeta.jl macros allow passing of keyword arguments to their DataFrames.jl
-function equivelents. This can be done in two ways. When inputs are given as multiple 
+function equivelents. 
+
+| Macro | Base DataFrames.jl called |
+|-------|---------------------------|
+| @subset | `subset` |
+| @subset! | `subset!` |
+| @rsubset | `subset` |
+| @rsubset! | `subset!` |
+| @orderby | None (no keyword arguments supported) |
+| @rorderby | None (no keyword arguments supported) |
+| @by | `combine` |
+| @combine | `combine` |
+| @transform | `transform` |
+| @transform! | `transform!` |
+| @rtransform | `transform` |
+| @rtransform! | `transform!` |
+| @select | `select` |
+| @select! | `select!` |
+| @rselect | `select` |
+| @rselect! | `select!` |
+
+This can be done in two ways. When inputs are given as multiple 
 arguments, they are added at the end after a semi-colon `;`, as in
 
 ```julia
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -403,9 +403,10 @@ julia> @rtransform df @passmissing :x = parse(Int, :x_str)
 ## Passing keyword arguments to underlying DataFrames.jl functions
 
 All DataFramesMeta.jl macros allow passing of keyword arguments to their DataFrames.jl
-function equivelents. 
+function equivelents. The table below describes the correspondence between DataFramesMeta.jl
+macros and the function that is actually called by the macro. 
 
-| Macro | Base DataFrames.jl called |
+| Macro | Base DataFrames.jl function called |
 |-------|---------------------------|
 | @subset | `subset` |
 | @subset! | `subset!` |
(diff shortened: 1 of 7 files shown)


pdeffebach:
@jkrumbiegel can I get a quick check on the implementation? It borrows from DataFrameMacros.jl. 

With your approval I will add the same structure to everything else.

jkrumbiegel:
Yeah looks ok from a quick glance. I still haven't decided how keyword arguments could work for non-parentheses style.

pdeffebach:
I guess the current code isn't very DRY, but each 4-line `*_helper` function is unique and it seems overly abstract to make an outer function right now. 

Will plug away with tests.

pdeffebach:
I think I have the pipeline figured out for how to process arguments. I've written it for `@rtransform` but need to port it to the rest of the macros.

bkamins:
Nice!

pdeffebach:
Okay! I think this is all done @bkamins @nalimilan @jkrumbiegel 

Please take a look!

docs/src/index.md
```diff
@@ -400,6 +400,71 @@ julia> @rtransform df @passmissing :x = parse(Int, :x_str)
    3 │ missing  missing
 ```
 
+## Passing keyword arguments to underlying DataFrames.jl functions
+
+All DataFramesMeta.jl macros allow passing of keyword arguments to their DataFrames.jl
```
bkamins, line 405: can you add a list of macros for which this is supported (e.g. does `@orderby` support it?)

src/macros.jl
```diff
@@ -728,18 +758,19 @@ write
 ##############################################################################
 
 function subset_helper(x, args...)
-    exprs, outer_flags = create_args_vector(args...)
+    x, exprs, outer_flags, kw = get_df_args_kwargs(x, args...; wrap_byrow = false)
+
     t = (fun_to_vec(ex; no_dest=true, outer_flags=outer_flags) for ex in exprs)
     quote
-        $subset($x, $(t...); skipmissing=true)
+        $subset($x, $(t...); (skipmissing = true,)..., $(kw...))
```
bkamins, line 765: is passing `skipmissing` explicitly tested? Why does `skipmissing` need to be in a tuple and later `...`?

src/macros.jl
```diff
@@ -1094,12 +1184,12 @@ end
 ## @orderby
 ##
 ##############################################################################
-
+# TODO: Determine which keyword arguments to expose to user
```
bkamins, line 1187: TODO left

src/macros.jl
pdeffebach: It is explicitely tested for throwing an error. Every keyword argument is tested. 

I do not know why exactly I need a tuple and splatting, but by omitting that I get 

```
julia> @subset(df, :x .== 1; skipmissing = false)
ERROR: syntax: keyword argument "skipmissing" repeated in call to "DataFrames.subset" around /home/peterwd/Documents/Development/DataFramesMeta/src/macros.jl:765
Stacktrace:
 [1] top-level scope
   @ REPL[8]:1
```

so it seems necessary.

src/macros.jl
pdeffebach: Deleted the TODO, as it was unnecessary. 

We will think about keyword arguments for `@orderby` later.

src/macros.jl
```diff
@@ -1962,6 +2154,24 @@ and
 
 $ASTABLE_MACRO_FLAG_DOCS
 
+`@by` accepts the same keyword arguments as `DataFrames.combine` and can be added in
```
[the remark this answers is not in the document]
pdeffebach: Added!

pdeffebach:
Thanks!

pdeffebach:
Okay I'm ready to merge this. There will always be more docs to add but I think this is now pretty comprehensive.

src/macros.jl
bkamins: ah - OK. So that is why it is needed. If you pass it explicitly uniqueness is tested, if you pass collections they are merged. This is strange, as this is not required in other contexts:
```
julia> (;a=1, [:a => 2]...)
(a = 2,)
```
(and you get merging without having to wrap the first in a tuple.

bkamins (approved):

PR #332, opened 2022-07-03
Update Project.toml

zsz00:


nalimilan:
Indeed, as Chain is reexported then allowing a new version which is breaking requires tagging a breaking version of DataFramesMeta too. Better disallow 0.4 too, otherwise users can silently get a different API just because Chain changes version.

pdeffebach:
Thanks! @zsz00 can you make these changes?

pdeffebach:
yes please.

zsz00:
Done

zsz00:
can you realse a 0.12.0 ?

pdeffebach:
Released! Thanks.

PR #335, opened 2022-07-08
 Fix eachrow bug #334

### LINKED ISSUE #334: Problem with eachrow and eachrow! + break/continue
Here is a MWE:
```
julia> df = DataFrame(a=1:2, b=2:3);

julia> @eachrow! df begin
               for i in 1:1
                   continue
                   println("and this is super bad as inner loop is affected")
               end
               continue
               println("should not happen")
           end;
and this is super bad as inner loop is affected
should not happen
and this is super bad as inner loop is affected
should not happen
```

pdeffebach:
Fixes #334 

Culprit was code from 0.6 which I believe is not needed.

### DIFF
--- a/test/eachrow.jl
+++ b/test/eachrow.jl
@@ -195,4 +195,31 @@ end
     @test res.b == [2, 5]
 end
 
+@testset "#334" begin
+    df = DataFrame(x = [1, 2, 3])
+    s = Ref(0)
+    @eachrow df begin
+        continue
+        s[] = 100
+    end
+    @test s[] == 0
+
+    s = Ref(0)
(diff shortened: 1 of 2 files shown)


test/eachrow.jl
```diff
@@ -195,4 +195,31 @@ end
     @test res.b == [2, 5]
 end
 
+@testset "#334" begin
```
bkamins, line 198: maybe update the docstring of `@eachrow` and `@eachrow!` with information that what this macro does is that it creates a for loop so, in particular, `break` and `continue` may be used?

bkamins (approved):

PR #338, opened 2022-10-03
Replace `sortperm` by alternative pattern

bkamins:
Fixes https://github.com/JuliaData/DataFramesMeta.jl/issues/337

### DIFF
--- a/docs/src/dplyr.md
+++ b/docs/src/dplyr.md
@@ -270,7 +270,7 @@ end
 ```
 
 Alternatively, using the StatsBase.jl package you could have written `ordinalrank(:sleep_total, rev=true)`.
-Finally, if you wanted to avoid using an additional package you could use the following expression `invperm([sortperm(:sleep_total, rev=true))`.
+Finally, if you wanted to avoid using an additional package you could use the following expression `invperm(sortperm(:sleep_total, rev=true))`.
 
 ## Create New Columns Using `@transform` and `@rtransform`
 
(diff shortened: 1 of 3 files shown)


docs/src/dplyr.md
```diff
@@ -256,19 +256,22 @@ Same as above, except here we filter the rows for mammals that sleep for 16 or m
...
-        sortperm(:sleep_total, rev=true)
+        :order
+        -:sleep_total
     end 
     @rsubset :sleep_total >= 16
 end
 ```
 
+Alternatively, using the StatsBase.jl package you could have written `ordinalrank(:sleep_total, rev=true)`.
+Finally, if you wanted to avoid using an additional package you could use the following expression `(1:length(:sleep_total))[sortperm(:sleep_total, rev=true)]`.
```
bkamins, line 273:
```suggestion
Finally, if you wanted to avoid using an additional package you could use the following expression `invperm([sortperm(:sleep_total, rev=true))`.
```

pdeffebach:
Can you add an example with strings so that its clear `ordinalrank` is what to do with strings?

bkamins:
I have added it to the manual (I assume you wanted this).

In general we might think about a more flexible API for `@orderby` allowing to pass additional parameters (like `rev`) directly as now it is quite cumbersome. If you want to pass `by`, `lt` etc. you essentially need to use `sortperm` and `invperm` trick as `sortperm` accepts all arguments.

pdeffebach:
> `@orderby` allowing to pass additional parameters (like `rev`) directly as now it is quite cumbersome. If you want to pass `by`, `lt` etc. you essentially need to use `sortperm` and `invperm` trick as `sortperm` accepts all arguments.

Ehhh this doesn't seem so bad. But I'll make an issue to keep track. 

Thanks for the PR!

PR #340, opened 2022-10-07
Add `unique` macros 

### LINKED ISSUE #233: Add a `unique` macro
It turns out you can do transformations in `unique` with a DataFrame. See [here](https://github.com/JuliaData/DataFrames.jl/issues/2651). 

So we can easily add a macro for this.

MatthewRGonzalez:
This pull request resolves #233 by providing new functionalities to `DataFramesMeta.jl`

## New Macros 
The following macros are added 

* `@unique` and `@runique` 
* `@unique!` and `@runique!`

These macros extend the functionality of `DataFrames.unique` and `DataFrames.unique!` to the `DataFramesMeta.jl` ecosystem. Both general and row-wise versions are included for consistency with similar macro functions.

## Tests
* General tests are added in `unique.jl`
* Tested with `byrow` in `byrow.jl`


src/macros.jl
```diff
@@ -2161,3 +2161,277 @@ julia> @by df :a begin
...
+@unique(df, @byrow :x + :y)
+```
+
+becomes
+
+```
+unique(df, :x => ByRow((x,y) -> x + y))
+```
+
+a transformation which cannot be conveniently expressed
```
pdeffebach, line 2224: You can delete this line.

src/macros.jl
```diff
@@ -2161,3 +2161,277 @@ julia> @by df :a begin
...
+    t = (fun_to_vec(ex; no_dest=true, outer_flags=outer_flags) for ex in exprs)
+    quote
+        $DataFrames.unique($x, $(t...))
+    end
+end
+
+"""
+    @unique(x, args)
+
+Select unique rows in `AbstractDataFrame`s.
```
pdeffebach, line 2182: Can you also communicate that `args` can be transformations? 

What happens if multiple arguments are passed?  Currently it errors 

```
julia> @unique(df, :x, :b)
ERROR: MethodError: no method matching unique(::DataFrame, ::Symbol, ::Symbol)
```

Maybe just allow one argument for consistency? Or we need to do some other implementation which uses `select` under the hood and allows multiple arguments. 

I think that's the best way forward.

src/macros.jl
```diff
@@ -2161,3 +2161,277 @@ julia> @by df :a begin
...
+"""
+macro unique(x, args...)
+    esc(unique_helper(x, args...))
+end
+
+function runique_helper(x, args...)
+    exprs, outer_flags = create_args_vector(args...; wrap_byrow=true)
+    t = (fun_to_vec(ex; no_dest=true, outer_flags=outer_flags) for ex in exprs)
+    quote
+        $DataFrames.unique($x, $(t...))
```
pdeffebach, line 2262: @bkamins Looking at the docs for `unique`, the fact that transformations can be passed to `unique` is not explicitly documented. 
Instead of using `unique`, do 

```
df2 = select(df, t...; copycols = false)
rowidxs = (!).(nonunique(df2))
df[rowidxs, :]
```

test/unique.jl
```diff
@@ -0,0 +1,190 @@
...
+    df = DataFrame(A=[1, 1, 3, missing], B=[1, 1, 2, 1])
+
+    @test @unique(df) ≅ unique(df)
+
+    @test @unique(df, :A .+ :B) ≅ unique(df, [:A, :B] => (x, y) -> x .+ y)
+    @test @unique(df).A isa Vector{Union{Missing,Int}}
+
+    @test @unique(df, $:A .+ :B) ≅ unique(df, [:A, :B] => (x, y) -> x .+ y)
+    @test @unique(df, $:A.+$:B) ≅ unique(df, [:A, :B] => (x, y) -> x .+ y)
+    @test @unique(df, :A.+$:B) ≅ unique(df, [:A, :B] => (x, y) -> x .+ y)
```
pdeffebach, line 20: Can you test with string column names?

test/unique.jl
```diff
@@ -0,0 +1,190 @@
...
+
+
+end
+
+
+@testset "unique with :block" begin
+    df = DataFrame(A=[1, 1, 3, missing], B=[1, 1, 2, 1])
+
+    d = @unique df begin
+        :A .+ :B
```
pdeffebach, line 55: You didn't test with multiple arguments ;)

```
julia> @unique df begin 
           :x
           :b
       end
ERROR: MethodError: no method matching unique(::DataFrame, ::Symbol, ::Symbol)
```

pdeffebach (changes requested):
Thanks!

I think we should change the implementation a bit to focus on `select` and mirror the internals of DataFrames.jl [here](https://github.com/JuliaData/DataFrames.jl/blob/897a8cbd8c3335239c8f050a1cfa10e559f6ca45/src/abstractdataframe/abstractdataframe.jl#L1502). 

Also, this means that 

```
@unique(df, [:a, :b])
```

means something very different from 

```
unique(df, [:a, :b]
```

that's okay, but this should be documented.

src/macros.jl
bkamins: `cols` in `unique` is documented as:

> in selected columns or their transformations

so it explicitly mentions that transformations are allowed. But maybe a better wording could be used (this was written a long time ago, when we have a much less precise vocabulary in DataFrames.jl 😄). If you have a proposal out of hand then please comment.

MatthewRGonzalez:
Hi, @pdeffebach. Thank you for your thoughtful feedback!

I've addressed these requested changes in 88a779b2cf164265f5a3577e7832dbafa2a8535b. In this commit, I implemented the `select` method you described above and also added a condition to handle cases where `t...` is not present. This additional condition allows `@unique(df)` to operate in the same manner as `unique(df)`

The documentation is updated to describe the difference between `@unique` and `unique`, and also describes that arguments can be transformations. Tests are also updated to include multiple arguments and string column names.

Please note that I used `gensym` in the helper functions in an attempt to keep the macros hygienic. If there's a more appropriate or concise method that you'd prefer, please let me know!

src/macros.jl
```diff
@@ -2161,3 +2161,312 @@ julia> @by df :a begin
...
+## @unique - distinct row selection
+##
+##############################################################################
+
+function unique_helper(x, args...)
+    exprs, outer_flags = create_args_vector(args...)
+    t = (fun_to_vec(ex; no_dest=true, outer_flags=outer_flags) for ex in exprs)
+    tmp = gensym("tmp")
+    quote            
+        $(tmp) = DataFrames.select($x, $(t...); copycols = false)
```
pdeffebach, line 2176: Can you put this all into a separate function? That way we can dispatch on `::AbstractDataFrame` and avoid thinking about `GroupedDataFrame`s

src/macros.jl
```diff
@@ -2161,3 +2161,312 @@ julia> @by df :a begin
...
+end
+
+"""
+    @unique(d, args...)
+
+Return the first occurence of unique rows in an `AbstractDataFrame` according 
+to given combinations of values in selected columns or their transformation. 
+`args` can be any column selector or transformation accepted by `select`.
+
+The implementation of `@unique` differs from the `unique` function in DataFrames.jl.
```
pdeffebach, line 2193: This paragraph is too technical. 

Please create a `### Details` section beneath `### Returns` to discuss the implementation. The main thing the user needs to know is that 

```
unique(df, [:a, :b])
```

is different from `@unique(df, [:a, :b])`

src/macros.jl
```diff
@@ -2161,3 +2161,312 @@ julia> @by df :a begin
...
+```julia
+@unique df begin 
+    :x + :y
+end
+```
+
+and 
+
+```
+@unique df :x + :y
```
pdeffebach, line 2222:
```suggestion
@unique(df, :x + :y)
```

src/macros.jl
pdeffebach: as in

```
function DataFramesMeta.make_unique(df, args...)
    ...
end

function unique_helper(x, args...)
    ...
    quote
        $make_unique($x, $t...)
    end
end
```

src/macros.jl
```diff
@@ -2161,3 +2161,312 @@ julia> @by df :a begin
...
+   4 │     4     2
+   5 │     5     1
+   
+julia> @runique(df, :x + :y)
+5×2 DataFrame
+ Row │ x      y
+     │ Int64  Int64
+─────┼──────────────
+   1 │     1     5
+````
```
pdeffebach, line 2312:
```suggestion
```
```

File: src/macros.jl [hunk 10]
```diff
@@ -2161,3 +2161,312 @@ julia> @by df :a begin
...
+macro unique(x, args...)
+    esc(unique_helper(x, args...))
+end
+
+function runique_helper(x, args...)
+    exprs, outer_flags = create_args_vector(args...; wrap_byrow=true)
+    t = (fun_to_vec(ex; no_dest=true, outer_flags=outer_flags) for ex in exprs)
+    tmp = gensym("tmp")
+    quote                    
+        if isempty($(args))
```
pdeffebach, line 2275: Having a separate function where we re-implement `DataFrames.unique` also helps reduce code-duplication.

src/macros.jl
pdeffebach: Okay I have the solution. 

Follow the implementation of `@orderby` and use `gensym_names = true`. That will fix the errors.

MatthewRGonzalez: I'm happy to be wrong, but I find that`@orderby(df, [:x, :z, :d])` results in the same error as above on `v0.12.0`, and that following the implementation of `@orderby` doesn't seem to resolve these errors.

pdeffebach: That's fine. If the user understands that it uses `select` machinery under the hood, and we communicate that `unique(df, [:x, :y, :z])` is fundamentally different than `@unique(df, [:x, :y, :z])` then we are fine. 

The problem is that 

```
@unique(df, [:x, :y, :z])
```

creates a three element array and tries to make that a column. If the data frame is a different length, it will error. 

The major problem here is that `@unique(df, [:x, :y, :z])` *will* work if the DataFrames has 3 rows... 

I think this is okay though. As long as users know what it's doing.

pdeffebach: I wonder if this deviates enough from `DataFrames.unique` to merit a new name? Similar to `sort` and `@orderby`. Maybe we should call this `@distinct`? I hope we don't get sued by rstudio for that. 

cc @bkamins @nalimilan

bkamins: So the deviation is that you want to allow `@distinct(df, :a, :b, :c)`? While `@distinct(df, $[:a, :b, :c])` is required as `@distinct(df, [:a, :b, :c])` errors? Then I think it would be better to have a new name.

pdeffebach: Yes. 

I agree, a new name is probably better here. @MatthewRGonzalez can you change to `@distinct`?

MatthewRGonzalez: Sure. I'll commit with some updated documentation later on.

pdeffebach: Thanks!

src/macros.jl
```diff
@@ -2161,3 +2161,343 @@ julia> @by df :a begin
...
+
+### Details
+
+The implementation of `@distinct` differs from the `unique` function in DataFrames.jl. 
+When `args` are present, `@distinct` relies upon an internal `select` call which produces 
+an intermediate data frame containing columns of `df` specified by `args`. The unique rows
+of `df` are thus determined by this intermediate data frame. This focus on `select` allows 
+for multiple arguments to be passed conveniently in the form of column names or transformations.
+
+Users should be cautious when passing function arguments as vectors. E.g., `@distinct(df, $[:x,:y])`
```
pdeffebach, line 2261:
```suggestion
Users should be cautious when passing function arguments as vectors. E.g., `@distinct(df, $DOLLAR[:x,:y])`
```

src/macros.jl
```diff
@@ -2161,3 +2161,343 @@ julia> @by df :a begin
...
+
+### Details
+
+The implementation of `@distinct!` differs from the `unique` function in DataFrames.jl. 
+When `args` are present, `@distinct!` relies upon an internal `select` call which produces 
+an intermediate data frame containing columns of `df` specified by `args`. The unique rows
+of `df` are thus determined by this intermediate data frame. This focus on `select` allows 
+for multiple arguments to be conveniently passed in the form of column names or transformations.
+
+Users should be cautious when passing function arguments as vectors. E.g., `@distinct(df, $[:x,:y])`
```
pdeffebach, line 2427:
```suggestion
Users should be cautious when passing function arguments as vectors. E.g., `@distinct(df, $DOLLAR[:x,:y])`
```

src/macros.jl
```diff
@@ -2161,3 +2161,343 @@ julia> @by df :a begin
...
+function rdistinct!_helper(x, args...)
+    exprs, outer_flags = create_args_vector(args...; wrap_byrow=true)
+    t = (fun_to_vec(ex; no_dest=true, outer_flags=outer_flags) for ex in exprs)
+    quote                    
+        $DataFramesMeta.make_distinct!($x, $(t...))    
+    end            
+end
+
+"""
+    rdistinct(d, args...)
```
pdeffebach, line 2473:
```suggestion
    rdistinct!(d, args...)
```

src/macros.jl
```diff
@@ -2161,3 +2161,343 @@ julia> @by df :a begin
...
+   4 │     4     2
+   5 │     5     1
+   
+julia> @rdistinct!(df, :x + :y)
+5×2 DataFrame
+ Row │ x      y
+     │ Int64  Int64
+─────┼──────────────
+   1 │     1     5
+````
```
pdeffebach, line 2499:
```suggestion
```
```

pdeffebach (changes requested):
Looks good!

Docs could be re-written a little but we can do that later. 

Happy to merge after you make the following small changes.

pdeffebach:
JK I figured out how to push to this branch so I will merge when tests pass. 

Thanks for your hard work on this!

MatthewRGonzalez:
@pdeffebach It looks like the checks failed due to issues with the`create_args_vector()`. I'm thinking this is probably due to the use of `create_args_vector()` instead of `get_df_args_kwargs()` introduced in  #323. Sorry for not catching this sooner-- I'll resolve it.

pdeffebach:
Tests pass locally... which is confusing.

If you feel comfortable rebasing and changing the logic of the macro to mirror the new implementations of `@select` etc. that might help. 

But rebasing is hard and you've done a lot of work on this so far, so I can take it from here if you need me to.

bkamins:
> But rebasing is hard and you've done a lot of work on this so far, so I can take it from here if you need me to.

Rebasing made easy is:
1. on your local git squash all commits in the PR into a single commit
2. rebase

The point is that if you do it this way you only need to resolve conflicts once and against the final version of the code. This is much easier than sequentially resolving all commits.

MatthewRGonzalez:
@pdeffebach I followed the advice above to update this PR. The tests pass locally after rebasing and I've updated the macro logic (e04fb01) to mirror the new implementations of `@select` etc. I apologize for the additional clutter. Please let me know if this is what you had in mind!

bkamins:
@pdeffebach can check best the details of the implementation, but I would like to thank you for an excellent and big PR (and tests now also pass CI 😄).

pdeffebach:
Thanks!

Issue #328, opened 2022-05-02
Error `.*` not defined with `@by`


junder873:
When I upgraded to v0.11, I ran into an error when using `@by`. For a MWE:

```julia
df= DataFrame(
    x = [1, 1, 2, 2, 3, 3],
    y = [true, false, true, false, true, false],
    z = [true, true, true, false, false, false]
)
@by(
    df,
    :x,
    :a = maximum(:y .* :z)
)
# ERROR: UndefVarError: .* not defined
# Stacktrace:
#  [1] top-level scope
#    @ C:\Users\underwjo\.julia\packages\DataFramesMeta\kWkhX\src\macros.jl:2056
```

Interestingly, when I do a simple check on `:z`, it works:
```julia
@by(
    df,
    :x,
    :a = maximum(:y .* (:z .== true))
)
# 3×2 DataFrame
#  Row │ x      a     
#      │ Int64  Bool  
# ─────┼──────────────
#    1 │     1   true
#    2 │     2   true
#    3 │     3  false
```

pdeffebach:
Thank you for the bug report!

I think I understand what's going on but it will take a few days to look into it more fully.

bkamins:
@pdeffebach The issue is caused by the automatic-composition mechanism you have recently introduced (that avoids creation of anonymous functions):
```
julia> @macroexpand @by(
           df,
           :x,
           :a = maximum(:y .* :z)
       )
quote
    #= C:\Users\bogum\.julia\packages\DataFramesMeta\kWkhX\src\macros.jl:2056 =#
    (DataFrames).combine((DataFrames.groupby)(df, :x), DataFramesMeta.make_source_concrete([:y, :z]) => ((∘)(maximum, .*) => :a))
end
```

pdeffebach:
It's more mysterious than that, though. On 1.7.2

```
julia> @by(
           df,
           :x,
           :a = maximum(:y .* :z)
       )
ERROR: UndefVarError: .* not defined
Stacktrace:
 [1] top-level scope
   @ ~/.julia/packages/DataFramesMeta/kWkhX/src/macros.jl:2056

julia> (DataFrames).combine((DataFrames.groupby)(df, :x), DataFramesMeta.make_source_concrete([:y, :z]) => ((∘)(maximum, .*) => :a))
3×2 DataFrame
 Row │ x      a
     │ Int64  Bool
─────┼──────────────
   1 │     1   true
   2 │     2   true
[...]

PR #341, opened 2022-10-27
Fix parsing bug #328

pdeffebach:


src/parsing.jl
```diff
@@ -244,7 +257,8 @@ function get_source_fun(function_expr; exprflags = deepcopy(DEFAULT_FLAGS))
         source = args_to_selectors(function_expr.args[2:end])
```
bkamins, line 257: @pdeffebach `get_source_fun` is key and complex function. Can we document in e.g. a comment the logic of processing and all cases it considers? Thank you!

pdeffebach: Yes. One quick question in addition. Is there any scenario where `[:x, :y] => ByRow(+)` is preferred over `[:x, :y] => .+` (assuming parsing works properly). 

Currently for non-nested functions, we convert `.+` to `ByRow(+)`, and the same for `.-`, `.*`, etc. 

It seems simpler to just maintain the base broadcasting, unless there are specializations for `ByRow` that I am missing that make things faster.

pdeffebach:
Filed an issue in the main Julia repo [here](https://github.com/JuliaLang/julia/issues/47408). 

Once they confirm my intuition is correct I will document and merge.

src/parsing.jl
bkamins: `ByRow` uses `map`. Normally broadcasting and `map` should give the same result and have the same performance (the reason is that you always pass vectors to them in your use case). However, in some particular cases there might be some small differences, e.g. a different type of vector can be allocated. Here is an example:
```
julia> (1:3) .+ (1:3)
2:2:6

julia> map(+, 1:3, 1:3)
3-element Vector{Int64}:
 2
 4
 6
```
(again - in DataFrames.jl this will not happen as we do not allow ranges, but this shows that there might be slight differences)

PR #343, opened 2022-11-03
add rename macro, docs, and tests

### LINKED ISSUE #204: Add the `@rename` macro
It's not strictly necessary, because `DataFrames.rename` exists, but it is nice to have the keyword argument form.

MatthewRGonzalez:
This pull request resolves #204 by providing new functionalities to `DataFramesMeta.jl`

## New Macros 
The following macros are added 

* `@rename` and `@rename!` 

These macros extend the functionality of `DataFrames.rename` and `DataFrames.rename!` to the `DataFramesMeta.jl` ecosystem. These macros also allow arbitrary RHS expressions (per https://github.com/JuliaData/DataFramesMeta.jl/issues/204#issuecomment-1297767454). 

## Tests
* General tests are added in `rename.jl`
* Tests pass locally in `runtests.jl` 

For convenience, I added a [simple helper function](https://github.com/MatthewRGonzalez/DataFramesMeta.jl/blob/9f9e69f1313790b42b5247ebcb54c1ba43cce88f/src/macros.jl#L2758-L2764) that converts pair elements of the final argument transformation to strings. I'm happy to make any modifications. Thanks!


src/macros.jl
```diff
@@ -2748,3 +2748,192 @@ julia> @rdistinct!(df, :x + :y)
...
+* `d` : an AbstractDataFrame
+* `args...` : expressions of the form `:new = :old` specifying the change of a column's name 
+from "old" to "new"
+
+### Returns
+
+* `::AbstractDataFrame`
+
+Inputs to `@rename` can come in two formats: a `begin ... end` block, or as a series of
+arguments and keyword-like arguments. For example, the following are equivalent:
```
bkamins, line 2795: it is not clear for me in the docstring what "arguments" means and what "keyword-like arguments" means. In particular if we have a form `lhs = rhs` what values are accepted on both sides.

src/macros.jl
```diff
@@ -2748,3 +2748,192 @@ julia> @rdistinct!(df, :x + :y)
...
+ Row │ new1       old_col2   old_col3 
+     │ Float64    Float64    Float64  
+─────┼────────────────────────────────
+   1 │ 0.0176206  0.493592   0.348072
+   2 │ 0.861545   0.512254   0.85763
+   3 │ 0.263082   0.0267507  0.696494
+   4 │ 0.643179   0.299391   0.780125
+   5 │ 0.731267   0.18905    0.767292
+
+julia> @rename(df, :new1 = :old_col1, :new2 = $DOLLAR"old_col2")
```
bkamins, line 2824: Can you please explain what `$DOLLAR` does here an below? Thank you!

src/macros.jl
bkamins: it would be also good too mention that whole expressions can be escaped by `$` (like in the tests).

MatthewRGonzalez: Hi, @bkamins. Thanks for the review! I've updated the docs to clarify phrasing regarding the points you raised above. I'm happy to continue revising as you see fit. Thanks!

bkamins: Looks good. Thank you!

pdeffebach:
Thanks for your work on this!

I think we need to re-think how it's done, however. 

Currently, the whole RHS of the expression is passed to the *whole* `fun_to_vec` machinery.  That is, for complicated expressions, we make anonymous functions and pass those to the renaming function. 

```
julia> MacroTools.@macroexpand(@rename df :new = :old_col1 * :old_col3) |> MacroTools.prettify
:(make_renamed(df, DataFramesMeta.make_source_concrete([:old_col1, :old_col3]) => ((*) => :new)))

julia> @rename df :new = :old_col1 * :old_col3
ERROR: ArgumentError: Tried renaming :[:old_col1, :old_col3] to :(*) => :new, when :[:old_col1, :old_col3] does not exist in the data frame.
```

This seems like it could be a source of bugs. 

That said, I like how we restrict the RHS to be a "valid" column identifier. I think it's good to be consistent about differences between columns and data, even if some users might find it annoying. 

Can you try writing a helper function instead of `fun_to_vec` which just checks that the LHS and the RHS are both column identifiers and then makes a pair?

MatthewRGonzalez:
> Can you try writing a helper function instead of fun_to_vec which just checks that the LHS and the RHS are both column identifiers and then makes a pair?

Sure!  I added this a helper function called `fun_to_pair` in `parsing.jl`. I basically simplify `fun_to_vec` and recycle `get_column_expr()` to determine the validity of column identifiers. Please let me know if this is in line with you're expectations. Thanks!

src/macros.jl
```diff
@@ -2748,3 +2748,215 @@ julia> @rdistinct!(df, :x + :y)
...
+"""
+macro rename(x, args...)
+    esc(rename_helper(x, args...))        
+end
+
+function rename!_helper(x, args...)
+    x, exprs, outer_flags, kw = get_df_args_kwargs(x, args...; wrap_byrow = false)    
+    t = (fun_to_pair(ex) for ex in exprs)            
+    quote            
+        $DataFrames.rename!($x, $(t...))
```
pdeffebach, line 2862: Extended from comments below. Instead of doing `String(lhs) => String(rhs)` in the parsing stage, let's clean up the inputs all together. 

I would allocate an intermediate array of renaming calls. then write a function that turns every vector of pairs into a vector of pairs of strings. Then pass that to `rename`. Something like

```
julia> function pairs_to_str_pairs(args...)
           map(args) do arg
               if !(arg isa Pair)
                   throw(ArgumentError("Non-pair created in @rename"))
               end
                   
               string(first(arg)) => string(last(arg))
           end
       end;
```

then do 

```
        $DataFrames.rename!($x, $pairs_to_str_pairs($(t...)...))
```

This means you don't have to do `MacroTools.postwalk` in the parsing stage. See the function `make_source_concrete` for inspiration. It's used when we have more than one input column, i.e. `@rtransform df :x = :a + :b`

Edge case: What happens when we want to rename the first column, i.e. `@rename df :a = $1`. I want this to work, particularly because `rename(df, 1 => :a)` works. But I don't want to mess around with `names(df, 1)` in this stage. So its a tough balance. Fortunately we get to point fingers at DataFrames.jl for any limitations.

pdeffebach (commented):
Thanks!

src/parsing.jl
```diff
@@ -387,6 +387,54 @@ fun_to_vec(ex::QuoteNode;
...
+
+        s = get_column_expr(t)
+        if s === nothing
+            throw(ArgumentError("Invalid column identifier on LHS in DataFramesMeta.jl macro"))
+        end
+        
+        s
+    end
+
+    rhs = MacroTools.unblock(ex.args[2])
```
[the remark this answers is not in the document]
MatthewRGonzalez: The unblock allows for the use of `begin ... end` in cases such as the following: 

```
@rename df :newcol = begin
    "old_col" * "1"
end
```

src/macros.jl
MatthewRGonzalez: I've resolved your comments aside from this edge case. Could we handle this case with a modified version 
of `pairs_to_str_pairs`? 

```
function pairs_to_str_pairs(args...)
           map(args) do arg
               if !(arg isa Pair)
                   throw(ArgumentError("Non-pair created in @rename"))
               end              
               if first(arg) isa Int            
                   return first(arg) => string(last(arg))
               end      
               string(first(arg)) => string(last(arg))
           end
end
```

src/parsing.jl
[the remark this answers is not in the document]
pdeffebach: Wouldn't you need a `$` in there though? Since we don't allow arbitrary expressions on the RHS that are not wrapped in `$()`. So this seems redundant.

src/macros.jl
pdeffebach: Yes this seems good!

Remember, this is done at call-time, not at the parsing stage.

pdeffebach:
Thanks!

Sorry, I thought you had pushed your changes. Please push them and I will take a look.

src/parsing.jl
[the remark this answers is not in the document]
MatthewRGonzalez: You're right. I edited the above to include `$`. If I remove the unblock as in the following:
```
rhs = ex.args[2]
rhs_col = get_column_expr(rhs)
if rhs_col === nothing
    throw(ArgumentError("Invalid column identifier on RHS in DataFramesMeta.jl macro"))
end

# parsing.jl:424
```

then 
```
@rename df :newcol = begin
    $("old_col" * "1")
end
```

throws `Invalid column identifier on RHS in DataFramesMeta.jl macro`. I'll look this over more.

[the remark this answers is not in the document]
pdeffebach: I think its okay to disallow the second case. People can do 

```
$( begin 


)```

if they want to.

File: test/rename.jl [hunk 5]
```diff
@@ -0,0 +1,171 @@
...
+using Statistics
+
+const ≅ = isequal
+
+@testset "rename" begin
+
+    df = DataFrame(old_col1 = rand(10), old_col2 = rand([missing;2:10],10),old_col3 = rand(10));
+
+    # rename test set
+    @test @rename(df, :new1 = :old_col1) ≅ rename(df, :old_col1 => :new1)
```
pdeffebach, line 15: Can you add some tests with `$"a string" = ...` i.e. `$` on the LHS?

pdeffebach:
@MatthewRGonzalez Is this ready for a review?

MatthewRGonzalez:
@pdeffebach yes, thanks!

src/parsing.jl
```diff
@@ -387,6 +387,62 @@ fun_to_vec(ex::QuoteNode;
...
+
+function pairs_to_str_pairs(args...)
+    
+    map(args) do arg        
+        if !(arg isa Pair)            
+            throw(ArgumentError("Non-pair created in @rename"))
+        end         
+
+        if first(arg) isa Int            
+            return first(arg) => string(last(arg))
```
pdeffebach: @MatthewRGonzalez 

Thank you for your work. One small thing. I think `string` is too strong here. Here's the issue. `:(AsTable(:x))` passes `get_column_expr` because we allow `AsTable` to be used as a column identifier identifier. inside transformations. So you end up calling `string(:old_col) => string(AsTable(:x))`

The problem is that `string(AsTable(:x))` gets turned into `"AsTable(x)"`. 

The solution is two-fold

1. Make a new function called `get_column_expr_rename` which only looks for `:x` and `$(...)`, no `AsTable`
2. Add more error handling in the `pairs_to_str_pairs` function. 

I've done both and added more tests. I've also re-written the docstring. 

So I think this is good to merge!

MatthewRGonzalez: Good catch, @pdeffebach. Thanks for adding the style fixes, error handling, and adding `get_column_expr_rename` as well!

PR #346, opened 2022-12-08
Fix #328 parsing bug try 2

### LINKED ISSUE #328: Error `.*` not defined with `@by`

pdeffebach:
Just takes the bugfix from #341 

I cluttered up #341 with docstrings, and the diff is actually a little important. 

Ideal scenario: create `Expr(., f)` 

This works on 1.6+ but tests currently pass on 1.0 and 1.5. Including this fix would require changing compatability. This is a bug-fix so I don't want to do that. 

Current scenario: use `$ByRow($f)`

This is not should  to `Expr(., f)` except for extremely contrived edge cases which I do not know about. 

Once we release 1.0, we will depend on DataFrames.jl version which require 1.6 and above, and therefore depend on 1.6. 

The docstrings and explanations will be added in a later PR.

### DIFF
--- a/src/parsing.jl
+++ b/src/parsing.jl
@@ -255,7 +255,7 @@ function get_source_fun(function_expr; exprflags = deepcopy(DEFAULT_FLAGS))
         # extract source symbols from quotenodes
         source = args_to_selectors(function_expr.args[2].args)
         fun_t = function_expr.args[1]
-        fun = :(DataFrames.ByRow($fun_t))
+        fun = :($ByRow($fun_t))
     elseif is_nested_fun_recursive(function_expr, false)
         composed_expr = make_composed(function_expr)
         # Repeat clean up from simple non-broadcast above
--- a/src/parsing.jl
+++ b/src/parsing.jl
@@ -94,8 +94,7 @@ fix_simple_dot(x) = x
 function fix_simple_dot(x::Symbol)
     if startswith(string(x), '.')
         f_sym_without_dot = Symbol(chop(string(x), head = 1, tail = 0))
-        #return Expr(:., f_sym_without_dot)
-        return :($ByRow($f_sym_without_dot))
+        return Expr(:., f_sym_without_dot)
     else
         return x
     end
@@ -107,13 +106,12 @@ function make_composed(x::Expr)
     x_orig = x
     nested_once = false
     while true
+        fun = fix_simple_dot(x.args[1])
         if is_nested_fun(x)
-            fun = x.args[1]
             push!(funs, fun)
             x = x.args[2]
             nested_once = true
         elseif is_simple_non_broadcast_call(x) && nested_once
-            fun = fix_simple_dot(x.args[1])
             push!(funs, fun)
             # ∘(f, g, h)(:x, :y, :z)
             x = Expr(:call, Expr(:call, ∘, funs...), x.args[2:end]...)
@@ -266,7 +264,7 @@ function get_source_fun(function_expr; exprflags = deepcopy(DEFAULT_FLAGS))
         # extract source symbols from quotenodes
         source = args_to_selectors(function_expr.args[2].args)
         fun_t = function_expr.args[1]
-        fun = :($ByRow($fun_t))
+        fun = Expr(:., fun_t)
     elseif is_nested_fun_recursive(function_expr, false)
         composed_expr = make_composed(function_expr)
         # Repeat clean up from simple non-broadcast above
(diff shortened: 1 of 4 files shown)


pdeffebach:
@bkamins Let's merge this and add docs separately.

src/parsing.jl
```diff
@@ -89,26 +89,42 @@ function is_nested_fun_recursive(x::Expr, nested_once)
         return false
     end
 end
+
+fix_simple_dot(x) = x
+function fix_simple_dot(x::Symbol)
+    if startswith(string(x), '.')
+        f_sym_without_dot = Symbol(chop(string(x), head = 1, tail = 0))
+        #return Expr(:., f_sym_without_dot)
+        return :($ByRow($f_sym_without_dot))
```
bkamins, line 98: how do we know it is a `ByRow` call?

bkamins:
Looks almost good. This:
```
julia> x = [true, false]
2-element Vector{Bool}:
 1
 0

julia> y = .-(.+(x))
2-element Vector{Int64}:
 -1
  0

julia> df = DataFrame(x=x)
2×1 DataFrame
 Row │ x
     │ Bool
─────┼───────
   1 │  true
   2 │ false

julia> @select df :y = .-(.+(:x))
ERROR: UndefVarError: .- not defined
```
fails, but maybe we are OK, as it is a corner case.

The only situation where I think it could be useful is negation with `!`, e.g.:
```
julia> df = DataFrame(x=[true, false, false])
3×1 DataFrame
 Row │ x
     │ Bool
─────┼───────
   1 │  true
   2 │ false
   3 │ false

julia> @select df :y = (x -> .!(x))(reverse(:x))
3×1 DataFrame
 Row │ y
     │ Bool
─────┼───────
   1 │  true
   2 │  true
   3 │ false

julia> @select df :y = .!(reverse(:x))
ERROR: UndefVarError: .! not defined
```

pdeffebach:
Thank you! I was missing a `fix_simple_dot` somewhere. 

Ultimately, though, we actually have to change support to only include 1.6+. See the following

```
julia> ByRow(+)(1)
ERROR: MethodError: no method matching (::ByRow{typeof(+)})(::Int64)
Closest candidates are:
  (::ByRow)() at ~/.julia/packages/Tables/fouQB/src/utils.jl:241
  (::ByRow)(::NamedTuple{(), Tuple{}}) at ~/.julia/packages/Tables/fouQB/src/utils.jl:242
  (::ByRow)(::NamedTuple{names, T} where {N, names, T<:Tuple{Vararg{AbstractVector, N}}}) at ~/.julia/packages/Tables/fouQB/src/utils.jl:234
  ...
Stacktrace:
 [1] top-level scope
   @ REPL[20]:1

julia> .+(1)
1
```

This comes into play when we see 

```
@select df :z = .+(first(:x))
```

Where we have 

```
julia> .+(first([1]))
1

julia> ByRow(+)(first([1]))
ERROR: MethodError: no method matching (::ByRow{typeof(+)})(::Int64)
Closest candidates are:
  (::ByRow)() at ~/.julia/packages/Tables/fouQB/src/utils.jl:241
  (::ByRow)(::NamedTuple{(), Tuple{}}) at ~/.julia/packages/Tables/fouQB/src/utils.jl:242
  (::ByRow)(::NamedTuple{names, T} where {N, names, T<:Tuple{Vararg{AbstractVector, N}}}) at ~/.julia/packages/Tables/fouQB/src/utils.jl:234
  ...
Stacktrace:
 [1] top-level scope
   @ REPL[23]:1
```

so I need to lower things down to `Expr(., f)` and remove support for `< 1.6.0`

bkamins:
Yes, support for Julia 1.6 can be required.

Also you are right, by requiring to use `.` not `ByRow`. The reason, again, corner cases, is that `ByRow` has special meaning for `NamedTuple` also.

As for tests - maybe add some tests with data frames having more than 1 row.

pdeffebach:
Okay tests pass with this change. I think we should merge and then I can add the docs from #341 later.

bkamins (approved):

PR #366, opened 2023-10-31
Fix error with ByRow on import

### LINKED ISSUE #365: `ByRow` not defined when importing DataFramesMeta
Hello, I was trying to do this:
```
import DataFrames
import DataFramesMeta
df = DataFrames.DataFrame(a = 1:10)
DataFramesMeta.@rsubset!(df, :a < 6)
```
But I get: 
```
ERROR: UndefVarError: `ByRow` not defined
```
I'm using Julia v1.9.3 with DataFrames v1.6.1 and DataFramesMeta v0.14.0.
Am I going about this wrong?

pdeffebach:
fixes #365

### DIFF
--- a/src/parsing.jl
+++ b/src/parsing.jl
@@ -295,9 +295,9 @@ function get_source_fun(function_expr; exprflags = deepcopy(DEFAULT_FLAGS))
 
     if exprflags[BYROW_SYM][]
         if exprflags[PASSMISSING_SYM][]
-            fun = :(ByRow(Missings.passmissing($fun)))
+            fun = :($ByRow($(Missings.passmissing)($fun)))
         else
-            fun = :(ByRow($fun))
+            fun = :($ByRow($fun))
         end
     end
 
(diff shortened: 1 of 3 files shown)


pdeffebach:
Requesting a quick review. I went and added `$` in front of all `ByRow` and `AsTable` calls. I also special-cased `AsTable` so that `AsTable` on the right-hand-side, such as a row-sum, works even if DataFramesMeta is only `imported`. 

```
julia> import DataFramesMeta

julia> import DataFrames

julia> df = DataFrames.DataFrame(x1 = 1, x2 = 2, x3 = 3);

julia> DataFramesMeta.@rtransform df :z = sum(AsTable([:x1, :x2, :x3]))
1×4 DataFrame
 Row │ x1     x2     x3     z     
     │ Int64  Int64  Int64  Int64 
   1 │     1      2      3      6
```


I figures this is okay since we "own" the expression we are operating on. But long term, it might be nice to make this something DataFramesMeta.jl owns more concretely.

bkamins (approved):
looks OK

pdeffebach:
Failures on nightly are unrelated. We should merge.

PR #367, opened 2023-10-31
Remove @with, add @attach

### LINKED ISSUE #364: `@with` macro clashes with `Base.@with` in Julia 1.11+
Julia 1.11 is the first version to export `Base.@with`:
```
help?> @with
  @with vars... expr

  Macro version of with(f, vars...) but with expr instead of f function. This is similar to using with with a do block, but avoids creating a closure.
```

This causes a name conflict with `DataFramesMeta.@with` (i.e., error when the latter used in a module). MWE:
```julia
julia> module Foo
    using DataFrames, DataFramesMeta

    const df = DataFrame(a = [1, 2, 3])
    print() = @with df println.(:a)
end
WARNING: both DataFramesMeta and Base export "@with"; uses of it in module Foo must be qualified
ERROR: LoadError: UndefVarError: `@with` not defined
in expression starting at REPL[14]:6

```

pdeffebach:
fixes #364


src/macros.jl
```diff
@@ -663,6 +663,17 @@ macro attach(d, body)
     esc(attach_helper(d, body))
 end
 
+"""
+    @with(x, args...)
+
+Deprecated version of `@attach`, see `?@with` for details.
+"""
+macro with(x, args...)
+    @warn "`@with is deprecated, use `@with` instead."
```
bkamins, line 672:
```suggestion
    @warn "`@with is deprecated, use `@attach` instead."
```

src/macros.jl
```diff
@@ -663,6 +663,17 @@ macro attach(d, body)
     esc(attach_helper(d, body))
 end
 
+"""
+    @with(x, args...)
+
+Deprecated version of `@attach`, see `?@with` for details.
```
bkamins, line 669:
```suggestion
Deprecated version of `@attach`, see `?@attach` for details.
```

pdeffebach:
Tests pass

bkamins:
There are some doctest failures

bkamins (approved):

pdeffebach:
This can be a small version bump, right? Since it just deprecates, we don't need to make a `0.15.0` version.

bkamins:
we add new `@attach`. In the past, we would e.g. in DataFrames.jl make 0.15 release. But recently people were saying that such minor changes are OK to be made in 0.14.2.

pdeffebach:
Okay I'm just going to merge this with a minor version bump. I don't want to release too many breaking changes and I want people to see the deprecation. 

I will also announce on discourse soon.

bkamins:
OK

nalimilan:
Sorry, for some reason I missed this. It's really unfortunate that Julia exports this general name now. Have you considered calling it `@within` instead of `@attach` (which indeed has a bad reputation in R as it attaches copies of columns to the global environment)?

pdeffebach:
`@within` is a good name. I guess I still prefer `@attach` since

1. It's a verb, kind of like `@select`, `@transform` etc. 
2. I like the idea of us re-doing an R feature, but "correctly", without the scoping and performance drawbacks of `attach`.

pdeffebach:
Closing this due to https://github.com/JuliaLang/julia/pull/53004#issuecomment-1981109599 being merged

PR #372, opened 2023-12-18
Allow for Between, Not, All, Cols in `@select`

pdeffebach:


pdeffebach:
Tests fail because of `@linq`. This means we finally get to deprecate `@linq` completely and remove it!

bkamins:
I would remove `@linq`.

bkamins (approved):
Maybe add a comment that anything within these calls is left untouched. What I mean is that `Cols(:A)` does not do re-interpretation of `:A`.

pdeffebach:
Doctest failures are fixed in #367 so I'm comfortable merging.

pdeffebach:
Actually, I'm going to change the implementation slightly so `@groupby` works better.

pdeffebach:
@bkamins okay to merge? This is a pretty simple PR. Will open a new PR for `@groupby` using this shortly.

bkamins:
Looks good

PR #373, opened 2023-12-18
add groupby and docs

pdeffebach:
Adding a `@groupby` convenience macro has been suggested before (though I can't find the exact issues). Currently, you have to write

```
@chain df begin
    groupby([:a, :b])
    @transform ...
end
```

The extra `[` and the lack of `@` is frustrating and gets you out of the "flow" of a chained operation. 

This PR allows

```
@chain df begin
  @groupby :a :b 
  @transfrom ...
end
```

which is cleaner

This macro also provides a few convenience syntaxes

1. Allows for mixing `String` and `Symbol` column selectors
2. Allows for "block" format
3. Automatically concatenates to allow for scalar and vector selection

Unlike DataFrameMacros.jl's `@groupby` is does *not* allow for new column creation. This is intentional. I don't want `@groupby` to create any copies or have any ambiguity about when it creates a copy, which DataFrameMacros.jl's `@groupby` does (see [here](https://github.com/jkrumbiegel/DataFrameMacros.jl/blob/58716b0921cce53b93312bcc2881ec5f587df5b8/src/DataFrameMacros.jl#L76))`. 

With my implementation, users can mindlessly write `@groupby` without worrying about performance at all.


bkamins (approved):
Looks good, but I would recommend to make it more precise how `args...` are transformed to a single argument to `groupby` as it was not fully clear to me. What if someone e.g. writes
`@groupby(df, Cols(r"a"), Not("b"))`.

pdeffebach:
@bkamins Ready for a review. 

I opted against adding any escaping rules for `@groupby df x`. `@groupby df x y` is now exactly the same as `groupby(df, Cols(x, y))`. No `$` or anything else needed. 

1. This corresponds with `@by`, which is our other maco which allows grouping. 
2. This allows for `@groupby df [:a, :b]` easily. Users don't *really* need to do `[:a, :b]` in `@groupby` when they could just do `@groupby df :a :b`, but they *do* need to do that in `@by`, which doesn't allow grouping in multiple expressions. 
3. It's simpler and easier to explain to users.

bkamins (approved):

pdeffebach:
Thanks!

PR #379, opened 2024-02-26
Make a release for 0.15.0

pdeffebach:
We've added two new features since the last release

* Added support for `Not`, `All`, and `Cols` in `@select`, making it easier to keep or drop many columns at once. ([#372](https://github.com/JuliaData/DataFramesMeta.jl/pull/372))
* Added the `@groupby` macro, which provides an easier syntax for grouping data frames ([#373](https://github.com/JuliaData/DataFramesMeta.jl/pull/373))

I always think I will wait until the next feature (metadata syntax, or `@when`) until the next release. But realistically we should keep a faster clip of releases. 

I put 0.15.0 as the release number because of the new features. We really need to get to 1.0. Which I think I will do after metadata is merged. 

@bkamins

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,8 @@
+# DataFramesMeta v0.15.0 Release notes
+
+* Added support for `Not`, `All`, and `Cols` in `@select`, making it easier to keep or drop many columns at once. ([#372](https://github.com/JuliaData/DataFramesMeta.jl/pull/372))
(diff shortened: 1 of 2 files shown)


pdeffebach:
Test failure is due to `julia-1.11`. But I'm pretty sure `@with` is going to get un-exported in `v1.11` so I'm not going to change it yet.

bkamins (approved):

PR #384, opened 2024-03-01
Small macro hygiene fixes

pdeffebach:


pdeffebach:
Tests are hanging, but pass locally.

PR #391, opened 2024-03-18
Bump Chain.jl compat

pdeffebach:
Chain.jl has been at 0.6 for a month now That's a problem! If you add `Chain`, then `DataFamesMeta`, you get a super old version.


pdeffebach:
This Chain.jl release is only breaking in the most contrived way possible. Our current version bump is fine.

bkamins (approved):
We could also allow both 0.5 and 0.6, but this is minor

PR #400, opened 2024-06-26
Fix keyword argument bug 

### LINKED ISSUE #399: kwargs not passed to subset in `@subset` with multiple conditions
See the example below:

```julia
df = DataFrame(a = [1, 2, missing], b = [4, 5, 6])
@subset(df, :a .== 1; view = true)
# return a SubDataFrame

@subset(df, :a .== 1; skipmissing = false)
# throw an error as there is a missing value in the condition

@subset(df, :a.== 1, :b .== 4; view = true)
# return a DataFrame rather than a SubDataFrame!!!

@subset(df, :a.== 1, :b .== 4; skipmissing = false)
# return a DataFrame! No error was thrown.
```

Version info: Julia 1.10

  [a93c6f00] DataFrames v1.6.1
  [1313f7d8] DataFramesMeta v0.15.2

pdeffebach:
Fixes #399

### DIFF
--- a/test/keyword_arguments.jl
+++ b/test/keyword_arguments.jl
@@ -401,4 +401,28 @@ end
     @test df2 == correct
 end
 
+@testset "Multiple arguments #399" begin
+    correct = df[df.a .== 1, :]
+    correct_view = view(df, df.a .== 1, :)
+
+    df2 = @subset(df, :a .== 1, :b .== 3; view = true)
+    @test df2 ≈ correct_view
+
+    @test_throws ArgumentError @subset(df, :a .== 1, :b .== 3; skipmissing = false)
+    @test_throws ArgumentError @subset(df, :a .== 1, :b .== 3; skipmissing = false, view = true)
+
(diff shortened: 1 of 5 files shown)


pdeffebach:
@bkamins This is a bug-fix that is relatively urgent, see #399 

I think this fix is good though. Tests pass both locally and on github actions.

test/keyword_arguments.jl
```diff
@@ -401,4 +401,28 @@ end
     @test df2 == correct
 end
 
+@testset "Multiple arguments #399" begin
+    correct = df[df.a .== 1, :]
+    correct_view = view(df, df.a .== 1, :)
+
+    df2 = @subset(df, :a .== 1, :b .== 3; view = true)
+    @test df2 ≈ correct_view
```
bkamins, line 409: Can you additionally check type of `df2`?

bkamins (approved):

test/keyword_arguments.jl
pdeffebach: Sorry, I missed this!

Issue #404, opened 2024-11-11
`@select!` errors where `select!` does not with multi-column selection


nathanrboyer:
I get an unexpected error with `@select!` below, and the error message reads to me like it should be working.

```julia-repl
julia> using DataFramesMeta

julia> df = DataFrame(x=1:3, y=4:6, z=7:9)
3×3 DataFrame
 Row │ x      y      z
     │ Int64  Int64  Int64
─────┼─────────────────────
   1 │     1      4      7
   2 │     2      5      8
   3 │     3      6      9

julia> @select!(df, Not(:x), :x)
ERROR: LoadError: ArgumentError: Multi-column references outside of @select, @rselect, @select! and @rselect! must be wrapped in AsTable
Stacktrace:
 [1] get_column_expr(e::Expr; allow_multicol::Bool)
   @ DataFramesMeta C:\Users\nboyer.AIP\.julia\packages\DataFramesMeta\Ga6VV\src\parsing.jl:37
 [2] get_column_expr
   @ C:\Users\nboyer.AIP\.julia\packages\DataFramesMeta\Ga6VV\src\parsing.jl:22 [inlined]
 [3] fun_to_vec(ex::Expr; gensym_names::Bool, outer_flags::@NamedTuple{…}, no_dest::Bool, allow_multicol::Bool)
   @ DataFramesMeta C:\Users\nboyer.AIP\.julia\packages\DataFramesMeta\Ga6VV\src\parsing.jl:359
 [4] fun_to_vec
   @ C:\Users\nboyer.AIP\.julia\packages\DataFramesMeta\Ga6VV\src\parsing.jl:326 [inlined]
 [5] (::DataFramesMeta.var"#56#57"{@NamedTuple{…}})(ex::Expr)
   @ DataFramesMeta .\none:0
 [6] iterate(::Base.Generator{Vector{Any}, DataFramesMeta.var"#56#57"{@NamedTuple{…}}})
   @ Base .\generator.jl:48
 [7] select!_helper(::Symbol, ::Expr, ::Vararg{Any})
   @ DataFramesMeta C:\Users\nboyer.AIP\.julia\packages\DataFramesMeta\Ga6VV\src\macros.jl:1987
 [8] var"@select!"(__source__::LineNumberNode, __module__::Module, x::Any, args::Vararg{Any})
[...]

PR #405, opened 2024-11-14
Fix multi-column select! #404

pdeffebach:


pdeffebach:
@bkamins small bug-fix. Will merge.

bkamins (approved):

PR #415, opened 2025-10-26
Update to PrettyTables.jl v3

pdeffebach:
cc @bkamins 

I noticed some issue on my work computer where PrettyTables.jl v3 was accidentally loaded, even though the Project.toml didn't allow for it? I have to explore it more. 

However this is a fix to support PrettyTables.jl *both* 2 and 3, using `@static if` copying DataFrames.jl. 

I think DataFramesMeta.jl has been a little out-of-sync with DataFrames.jl for a while due to this.


bkamins (approved):
Looks good, though @ronisbr could also have a look at it :). Thank you!

ronisbr:
Hi! Sorry for the delay.

My suggestion is to change the new call to use the argument `show_first_column_label_only = true` instead of providing a new vector with column labels. This argument is the same as the previous `show_subheader = false`.

Issue #259, opened 2021-06-20
Bug with `@where true`


pdeffebach:
When we have `@where true`, `true` is not a `QuoteNode` or a `Expr`. We don't have a method for it. 

We should fix this, combined with #253 to just return `Return(x)` for anything *not* an expression, i.e. string, ints, or booleans.

pdeffebach:
It looks like `Returns` is in 1.7. Should be easy to add.

pdeffebach:
This is actually kind of complicated, because `subset df true` and `select df 1` have different meanings. Do we even want `select df 1` to work? What about `select df t`? 

We need a keyword argument `allow_literals`, I think. And it's *only* true for `@subset` and `@orderby`. But even then... I don't know maybe we want `^(true)` instead.

PR #420, opened 2026-05-12
Byrow-aware literal handling for @subset / @rsubset (#259)

kimjune01:
## Summary

`@subset(df, true)` previously threw an opaque `MethodError` (no `fun_to_vec` dispatch for literals), which is the bug in #259.

This handles literals in a way that mirrors the underlying `DataFrames.subset` exactly, keyed on the `@byrow` flag:

- **`@subset(df, true)`** (not byrow): a bare literal is a scalar predicate, which `subset` rejects just as `subset(df, [] => Returns(true))` does. It now raises a clear `ArgumentError` in place of the `MethodError`.
- **`@rsubset(df, true)`** (byrow): a per-row literal is well defined and matches `subset(df, [] => ByRow(Returns(true)))`, which works, so it returns the expected rows.

`ByRow` is applied only when the `@byrow` flag is set. Applying it unconditionally (the first approach here) made `@subset` accept what the function rejects, which is the divergence raised in review.

## Test plan

- `@subset(df, true)` / `false` / grouped / `@subset!` raise at macro expansion (asserted via `@test_throws LoadError @eval ...`, matching the suite's convention for invalid expansions).
- `@rsubset(df, true)` returns all rows, `@rsubset(df, false)` returns none, including the grouped case.
- Real column predicates (`@subset(df, :A .> 1)`) unaffected.

There was no test coverage for literal predicates in either mode before, so this adds it for both.

### DIFF
--- a/src/parsing.jl
+++ b/src/parsing.jl
@@ -421,6 +421,22 @@ fun_to_vec(ex::QuoteNode;
            outer_flags::Union{NamedTuple, Nothing}=nothing,
            allow_multicol::Bool = false) = ex
 
+# Catch-all method for literal values (Bool, Int, String, etc.)
+# Wraps them in Returns to create a constant function
+# This allows syntax like @subset(df, true) or @subset(df, false)
+function fun_to_vec(ex;
+                    no_dest::Bool=false,
+                    gensym_names::Bool=false,
+                    outer_flags::Union{NamedTuple, Nothing}=nothing,
+                    allow_multicol::Bool = false)
+    if no_dest
+        # For @subset and @with, return a constant function
+        return :([] => Returns($ex))
+    else
+        throw(ArgumentError("Literal values are only supported in @subset, @rsubset, @with, and similar macros"))
+    end
+end
+
 
 """
     rename_kw_to_pair(ex::Expr)
--- a/src/parsing.jl
+++ b/src/parsing.jl
@@ -422,7 +422,10 @@ fun_to_vec(ex::QuoteNode;
            allow_multicol::Bool = false) = ex
 
 # Catch-all method for literal values (Bool, Int, String, etc.)
-# Wraps them in Returns to create a constant function
+# Wraps them in ByRow(Returns(...)) to create a constant vector function.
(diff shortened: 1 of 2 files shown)


pdeffebach:
Thanks! But I'm not 100% sure this is right. `@subset(df, true)` should error, as `subset(df, [] => Returns(true))` errors, right?

kimjune01:
Good call. Reworked it to raise a clear ArgumentError instead of broadcasting, so the macro stays consistent with subset(df, [] => Returns(true)); if you want literals as a real feature later, that's the scoped allow_literals / ^(true) path from #259.

src/parsing.jl
```diff
@@ -421,6 +421,24 @@ fun_to_vec(ex::QuoteNode;
...
+# functions return an AbstractVector, and ByRow with empty source columns
+# produces a vector via _empty_selector_helper.
+# This allows syntax like @subset(df, true) or @subset(df, false)
+function fun_to_vec(ex;
+                    no_dest::Bool=false,
+                    gensym_names::Bool=false,
+                    outer_flags::Union{NamedTuple, Nothing}=nothing,
+                    allow_multicol::Bool = false)
+    if no_dest
+        return :([] => $ByRow($(Base.Returns)($ex)))
```
pdeffebach, line 441: Don't you need to have `ByRow` here only if `byrow = true`?

kimjune01: seems like my fix was overly permissive. I was hoping tests passing would be good enough, but I was wrong. Added tests for both modes so to increase coverage.

pdeffebach:
Can we allow literals for non-by-row subsetting as well? 

```
julia> df = DataFrame(x = [1, 2, 3])
3×1 DataFrame
 Row │ x     
     │ Int64 
─────┼───────
   1 │     1
   2 │     2
   3 │     3

julia> b = [true, false, true]
3-element Vector{Bool}:
 1
 0
 1

julia> @subset df b
ERROR: LoadError: ArgumentError: literal value `b` is not a valid column expression; pass a column reference or a vectorised predicate (e.g. `:x .> 0`), or use `@rsubset` for a per-row literal
```

PR #421, opened 2026-06-01
Add Usage and Alternatives sections to README

nathanrboyer:
I think the README could use a little more advertisement up front to demonstrate why it is good.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,8 @@ This syntax simplification is demonstrated with two simple examples below.
 Define a data frame.
 
 ```julia
+# Importing DataFramesMeta also implicitly imports DataFrames
+using DataFramesMeta
 df = DataFrame(a = [1, 2], b = [3, 4]);
 ```
 
--- a/README.md
+++ b/README.md
@@ -22,6 +22,41 @@ or via the `Pkg` REPL mode (enter by typing `]` at the REPL console)
 ] add DataFramesMeta
 ```
 
+# Usage
+
+Instead of using DataFrames.jl [manipulation functions](https://dataframes.juliadata.org/stable/man/basics/#Manipulation-Functions) directly,
+which use operation pairs to define data frame manipulations,
+DataFramesMeta.jl offers macro alternatives to each manipulation function,
+which convert a more readable syntax into the appropriate DataFrames.jl function calls.
+This syntax simplification is demonstrated with two simple examples below.
+
+Define a data frame.
+
+```julia
+df = DataFrame(a = [1, 2], b = [3, 4]);
+```
+
+Add columns `a` and `b` together and store the result in a new column `c`.
+
+```julia
+# With DataFrames
+transform(df, [:a, :b] => ((x, y) -> x + y) => :c)
+
+# With DataFramesMeta
+@transform(df, :c = :a + :b)
+```
+
+Show only the data frame rows where column `a` is equal to `2`.
+(DataFramesMeta.jl offers rowwise 'r' versions of each manipulation macro as well for convenience.)
+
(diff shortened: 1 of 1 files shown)


bkamins (approved):

bkamins:
Looks good. @nalimilan - could you also please have a look?

README.md
```diff
@@ -22,6 +22,43 @@ or via the `Pkg` REPL mode (enter by typing `]` at the REPL console)
...
+
+Show only the data frame rows where column `a` is equal to `2`.
+(DataFramesMeta.jl offers rowwise 'r' versions of each manipulation macro as well for convenience.)
+
+```julia
+# With DataFrames
+subset(df, :a => ByRow(==(2)))
+
+# With DataFramesMeta
+@rsubset(df, :a == 2)
```
pdeffebach, line 59: remove parentheses? So it looks cleaner. 
```suggestion
@rsubset df :a == 2
```

README.md
```diff
@@ -22,6 +22,43 @@ or via the `Pkg` REPL mode (enter by typing `]` at the REPL console)
...
+```
+
+Add columns `a` and `b` together and store the result in a new column `c`.
+
+```julia
+# With DataFrames
+transform(df, [:a, :b] => ((x, y) -> x + y) => :c)
+
+# With DataFramesMeta
+@transform(df, :c = :a + :b)
```
pdeffebach, line 48:
```suggestion
@transform df  :c = :a + :b
```

README.md
```diff
@@ -22,6 +22,43 @@ or via the `Pkg` REPL mode (enter by typing `]` at the REPL console)
...
+
+Define a data frame.
+
+```julia
+# Importing DataFramesMeta also implicitly imports DataFrames
+using DataFramesMeta
+df = DataFrame(a = [1, 2], b = [3, 4])
+```
+
+Add columns `a` and `b` together and store the result in a new column `c`.
```
pdeffebach, line 41:
```suggestion
Add columns `:a` and `:b` together and store the result in a new column `:c`.
```

README.md
```diff
@@ -22,6 +22,43 @@ or via the `Pkg` REPL mode (enter by typing `]` at the REPL console)
 ] add DataFramesMeta
 ```
 
+# Usage
+
+Instead of using DataFrames.jl [manipulation functions](https://dataframes.juliadata.org/stable/man/basics/#Manipulation-Functions) directly,
+which use operation pairs to define data frame manipulations,
+DataFramesMeta.jl offers macro alternatives to each manipulation function,
+which convert a more readable syntax into the appropriate DataFrames.jl function calls.
+This syntax simplification is demonstrated with two simple examples below.
```
pdeffebach, line 31: Add something how its similar to R's dplyr?

README.md
nathanrboyer: I don't personally like the whitespace-dependent syntax, but your call. I am using the same examples from https://juliadata.org/DataFramesMeta.jl/dev/#Introduction which already have parentheses, so I'd change both or neither.

README.md
nathanrboyer: The column's name is `a` regardless of the type used to store the name. I think `"a"` would have similar meaning if you prefer that to ``` `a` ```. But I won't fight you further if you really want to use ``` `:a` ```.

README.md
nathanrboyer: That doesn't mean anything to me, but you can include a statement like that if you wish.

README.md
nathanrboyer: Same comment as above.

